PR #246, opened 2013-07-13
Adding ignore option to ignore a list of elements during the sync

oncletom:
I switched over `exec` to avoid the escaping of the arguments.
Also, it forced to make the `--rsh|-e` flag more explicit.

This is an example or `_config.yml` I have for my own website:
```yml
deploy:
  type: rsync
  host: oncletom
  user: oncletom
  root: ~/
  delete: true
  exclude: [.ssh, .bash*, .htaccess, images, talks]
```

### DIFF
--- a/lib/plugins/deployer/rsync.js
+++ b/lib/plugins/deployer/rsync.js
@@ -1,7 +1,7 @@
 var extend = require('../../extend'),
   term = require('term'),
   util = require('../../util'),
-  spawn = util.spawn,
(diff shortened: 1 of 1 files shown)


leesei:
Can you resubmit PR to [hexojs/hexo-deployer-rsync](https://github.com/hexojs/hexo-deployer-rsync)?

leesei:
Moved

Issue #774, opened 2014-08-05
Is this a bug of  util.highlight ?


wuchong:
When I update hexo to 2.8+ , I find  some code block has problems.

If there are a lot of lines in the code block, it will have a bank on the top and bottom . like this:

![0805103951](https://cloud.githubusercontent.com/assets/5378924/3806499/c5b60e26-1c4d-11e4-97f0-aa9134a81872.jpg)

If we have multiline comment , it will only parse the first line.  The html like this:

```html 
<pre>
<div class="line"><span class="comment">/*</span></div>
<div class="line"> * Integer 类中的 parseInt 函数</div>
<div class="line"> * Copyright (c) 1994, 2010, Oracle and/or its affiliates. All rights reserved.</div><div class="line"> */</div>
...
</pre>
```

I find #687 wrap each line with `<div class="line"> `

Is this a bug  ? Or do I need to modify my theme css? I use  [jacman](https://github.com/wuchong/jacman) for my theme.  @tommy351

leesei:
Fixed in 3.0.

PR #795, opened 2014-08-14
fix blank lines for #792 #774

### LINKED ISSUE #687: Is this a markdown parse error?
markdown as follows:

``` json
{
    "name": "pajjket",
    "time": "2014-6-18 19:10:41",
    "done": true
}
```

and result I got as below:
e9e5d406-2753-4c90-a82a-b2dcd48e821b

is this an issue? @tommy351

0ssifrage:
@tommy351 wrapped each line with `<div class="line">` to fix #687, however, there was a problem with codes with blank lines just like #792 and #774.

Adding css style for `.line` may be a solution,  but it's hard to let all the themes change their css files, I think using `span` instead of `div` and adding `\n` for each line can fix this issue without changing the css files, but I am not sure whether it will cause the same problem in #687. (I don't have the same problem in that issue before wrapping the line with `div`).

### DIFF
--- a/lib/util/highlight.js
+++ b/lib/util/highlight.js
@@ -88,8 +88,8 @@ module.exports = function(str, options){
     firstLine = options.first_line;
 
   lines.forEach(function(item, i){
-    numbers += '<div class="line">' + (i + firstLine) + '</div>';
-    content += '<div class="line">' + item + '</div>';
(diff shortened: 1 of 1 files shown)


leesei:
Fixed in 3.0.

Issue #797, opened 2014-08-14
Add image files to draft asset folder, hexo failed to start server


timnew:
I just added 2 image files to the draft asset folder, and  I found following error:

```
[timnew] hexo server -d                                                      23:33:30  ☁  hexo ☂ ⚡ ✭
Deprecation warning: moment().lang() is deprecated. Use moment().localeData() instead.
[error] HexoError: Process failed: _drafts/Atom-Editor-Background-Image/no_treeview.png
SyntaxError: Unexpected token ;
  at Object.Function (<anonymous>:null:null)
  at [object Object].precompile (/usr/local/lib/node_modules/hexo/node_modules/swig/lib/swig.js:483:11)
  at [object Object].compile (/usr/local/lib/node_modules/hexo/node_modules/swig/lib/swig.js:588:16)
  at Object.render (/usr/local/lib/node_modules/hexo/node_modules/swig/lib/swig.js:507:17)
  at /usr/local/lib/node_modules/hexo/lib/post/render.js:61:29
  at /usr/local/lib/node_modules/hexo/node_modules/async/lib/async.js:610:21
  at /usr/local/lib/node_modules/hexo/node_modules/async/lib/async.js:249:17
  at iterate (/usr/local/lib/node_modules/hexo/node_modules/async/lib/async.js:149:13)
  at async.eachSeries (/usr/local/lib/node_modules/hexo/node_modules/async/lib/async.js:165:9)
  at _asyncMap (/usr/local/lib/node_modules/hexo/node_modules/async/lib/async.js:248:13)
  at Object.mapSeries (/usr/local/lib/node_modules/hexo/node_modules/async/lib/async.js:231:23)
  at Object.async.series (/usr/local/lib/node_modules/hexo/node_modules/async/lib/async.js:608:19)
  at Object.module.exports [as render] (/usr/local/lib/node_modules/hexo/lib/post/render.js:57:9)
  at Array.async.auto.post [as 2] (/usr/local/lib/node_modules/hexo/lib/plugins/processor/post.js:193:17)
  at listener (/usr/local/lib/node_modules/hexo/node_modules/async/lib/async.js:493:46)
  at /usr/local/lib/node_modules/hexo/node_modules/async/lib/async.js:444:17
  at Array.forEach (native)
  at _each (/usr/local/lib/node_modules/hexo/node_modules/async/lib/async.js:46:24)
[...]

PR #885, opened 2014-10-21
Fix issue #797, don't render draft assets

mm--:
Hexo crashes when draft asset directories have images in them (see #797 ).
This happens because hexo attempts to process images as posts.


rscheuermann:
:+1: Would love to see this fixed.

leesei:
Can you resubmit PR to [hexojs/hexo-server](https://github.com/hexojs/hexo-server)?

leesei:
#797 fixed in 3.2beta

PR #909, opened 2014-11-11
use UTC date in permalink to prevent url change

phoenixlzx:
When I generating posts on my new computer, some posts permalink changed.

permalink change

Using UTC date time may fix this issue.


banbanchs:
+1

Xuanwo:
New version in Hexo looks like to fix this problem, isn't it?

leesei:
Affects permalink, closing before someone came up with a good solution.

Is there an issue originally?

Issue #925, opened 2014-11-19
code block parse error (bug)


seanlook:
I got following error when my code block contains `{{.word}}` 

```
Error: Unexpected key "word" on line 26 in file
```

In fact I found `{{word}}` have big problems when `word` is not English(my `word` is any chinese charactor) and any word between {{ }} is not displayed. Need to be fiexed.

Issue #924, opened 2014-11-19
Fail to parse markdown, when it contains code blocks with regular expression


lordfriend:
I have an markdown contain some code blocks like this.
```
```javascript
// prevent {{ }} being parsed by angular
{
  postCompile: function() {
    var match = src.match(/<pre>;[\s\S]+?<\/pre>/mg);
    if(match && match.length > 0) {
      match.forEach(function(str) {
        var subMatch = str.match(/{{[\s\S]+?}}/mg);
        var group, newStr = str;
        if(subMatch && subMatch.length > 0) {
          for( var i = 0; i < subMatch.length; i++) {
            group = subMatch[i].match(/{{([\s\S]+?)}}/m);
            newStr = newStr.replace(subMatch[i], '<span>{{</span>' + group[1] + '<span>}}</span>');
          }
        }
        src = src.replace(str, newStr);
      });
    }
    return src;
  }
}
```
```


 When I run hexo server. It failed with the following message.


[error] HexoError: Process failed: _posts/Problem-with-document-build-using-grunt-markdown.md
Error: Unable to parse "[\s\S]+?" on line 19 in file /Users/bob/Dropbox/blog/source/_posts/Problem-with-document-build-using-grunt-markdown.md.
    at Object.exports.throwError (/usr/local/lib/node_modules/hexo/node_modules/swig/lib/utils.js:183:9)
    at parseVariable (/usr/local/lib/node_modules/hexo/node_modules/swig/lib/parser.js:489:13)
    at /usr/local/lib/node_modules/hexo/node_modules/swig/lib/parser.js:633:15
    at Object.exports.each (/usr/local/lib/node_modules/hexo/node_modules/swig/lib/utils.js:45:11)
    at Object.exports.parse (/usr/local/lib/node_modules/hexo/node_modules/swig/lib/parser.js:622:9)
    at parse (/usr/local/lib/node_modules/hexo/node_modules/swig/lib/swig.js:343:19)
    at precompile (/usr/local/lib/node_modules/hexo/node_modules/swig/lib/swig.js:473:23)
    at compile (/usr/local/lib/node_modules/hexo/node_modules/swig/lib/swig.js:588:16)
    at Object.render (/usr/local/lib/node_modules/hexo/node_modules/swig/lib/swig.js:507:17)
    at /usr/local/lib/node_modules/hexo/lib/post/render.js:61:29

Issue #921, opened 2014-11-17
Code block generates error 


nateswart:
I've run into an issue with code blocks that are URLs and include certain characters. The following code block:

    ``` [html] URL-Scheme
    drafts4://x-callback-url/runAction?text={{%7B%22dataType%22%3A%22consumption-waste%22%2C%22dataSource%22%3A%22manual%22%2C%22urine%22%3Atrue%2C%22feces%22%3Afalse%7D}}&action={{Log QuantifiedSelf Entry}}&x-success={{launch:}}
    ```

Generates this error:

    [error] HexoError: Process failed: _posts/2014-10-19-Implementing-the-QS-Data-Spec-on-iOS.md
    Error: Unexpected colon on line 43 in file /Users/nate/Sites/nateofnine/source/_posts/2014-10-19-Implementing-the-QS-Data-Spec-on-iOS.md.
        at Object.exports.throwError (/usr/local/lib/node_modules/hexo/node_modules/swig/lib/utils.js:183:9)
        at Object.TokenParser.parseToken (/usr/local/lib/node_modules/hexo/node_modules/swig/lib/parser.js:319:15)
        at /usr/local/lib/node_modules/hexo/node_modules/swig/lib/parser.js:86:12
        at Object.exports.each (/usr/local/lib/node_modules/hexo/node_modules/swig/lib/utils.js:45:11)
        at Object.TokenParser.parse (/usr/local/lib/node_modules/hexo/node_modules/swig/lib/parser.js:76:11)
        at parseVariable (/usr/local/lib/node_modules/hexo/node_modules/swig/lib/parser.js:486:18)
        at /usr/local/lib/node_modules/hexo/node_modules/swig/lib/parser.js:633:15
        at Object.exports.each (/usr/local/lib/node_modules/hexo/node_modules/swig/lib/utils.js:45:11)
        at Object.exports.parse (/usr/local/lib/node_modules/hexo/node_modules/swig/lib/parser.js:622:9)
        at parse (/usr/local/lib/node_modules/hexo/node_modules/swig/lib/swig.js:343:19)

I'm basically trying to convert my site over to Hexo, and am stuck with this page:
http://nateofnine.com/blog/2014/10/19/implementing-the-qs-data-spec-on-ios

Any suggestions?

Issue #916, opened 2014-11-14
提交一个bug


lcepy:
代码块解析，遇到{{}} 会直接报错。```  ```  也不行，很多python的模版是这种风格，书写遇到它，就直接报错。
hexo 2.8.3

错误信息：

SyntaxError: Unexpected token ;
    at Object.Function (<anonymous>)
    at precompile (/Users/xiangwenwen/.nvm/v0.10.33/lib/node_modules/hexo/node_modules/swig/lib/swig.js:483:11)
    at compile (/Users/xiangwenwen/.nvm/v0.10.33/lib/node_modules/hexo/node_modules/swig/lib/swig.js:588:16)
    at Object.render (/Users/xiangwenwen/.nvm/v0.10.33/lib/node_modules/hexo/node_modules/swig/lib/swig.js:507:17)
    at /Users/xiangwenwen/.nvm/v0.10.33/lib/node_modules/hexo/lib/post/render.js:61:29
    at /Users/xiangwenwen/.nvm/v0.10.33/lib/node_modules/hexo/node_modules/async/lib/async.js:610:21
    at /Users/xiangwenwen/.nvm/v0.10.33/lib/node_modules/hexo/node_modules/async/lib/async.js:249:17
    at iterate (/Users/xiangwenwen/.nvm/v0.10.33/lib/node_modules/hexo/node_modules/async/lib/async.js:149:13)
    at async.eachSeries (/Users/xiangwenwen/.nvm/v0.10.33/lib/node_modules/hexo/node_modules/async/lib/async.js:165:9)
    at _asyncMap (/Users/xiangwenwen/.nvm/v0.10.33/lib/node_modules/hexo/node_modules/async/lib/async.js:248:13)

PR #926, opened 2014-11-19
Rendering Problems of Code Blocks Contain {{ }}

xing5:
Fix the bug of rendering post content with Swig, which is a template engine that reads all '{{}}' as a variable.

A solution to  https://github.com/hexojs/hexo/issues/925 https://github.com/hexojs/hexo/issues/924 https://github.com/hexojs/hexo/issues/921 https://github.com/hexojs/hexo/issues/916 https://github.com/hexojs/hexo/issues/910

I didn't test very carefully, please check if it works.

### DIFF
--- a/lib/post/render.js
+++ b/lib/post/render.js
@@ -59,6 +59,7 @@ module.exports = function(source, data, callback){
     function(next){
       try {
         data.content = swig.render(data.content, {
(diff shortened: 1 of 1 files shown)


iissnan:
How about code blocks including `{% ... %}`?

PR #930, opened 2014-11-22
Fix typo in this.options for chokidar.

bwinton:
It’s a pretty simple fix, but `hexo generate --watch` wasn’t working for me without this change.


tommy351:
Fixed in commit https://github.com/hexojs/hexo/commit/e604c68ec3f393562b1c9bd947f16d76acdda9fd

PR #936, opened 2014-11-26
Add a configuration option reverse_order for pagination

RussTheAerialist:
Being able to control the order of the pagination is extremely helpful for me as I'm using hexo to create a site for a novel that I'm writing and would like to enable people to be able to read the book online in chapter order, but the default would be in reverse chapter order, so a small change to the pagination plugin would allow you to specify that you would like to reverse the order of the posts.

### DIFF
--- a/lib/plugins/generator/paginator.js
+++ b/lib/plugins/generator/paginator.js
@@ -12,7 +12,14 @@ var Paginator = function(base, posts, num, total){
   this.total = total;
   this.current = num;
   this.current_url = format(base, num);
+
+  this.reverse_order = config.reverse_order;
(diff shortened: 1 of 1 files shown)


Xuanwo:
@RussTheAerialist looks good, but you need to rebase the code so that I can merge it.

leesei:
I think adding a `sort` config is more flexible.
Is it easy to use different config for index (first page), archive, tag can category?
We can pick up here if @RussTheAerialist don't want to do it.

RussTheAerialist:
I've moved away from using hexo in favor of Hugo, so please don't wait for
me to make the change.

PR #945, opened 2014-12-06
Add locale support to date function

ioxo:
Hi all,

Thanks for very good site generator. Works fine.
I've made very tiny change to date.js file. 
Hope this help other peoples too and is accepted to master. When you change _config.yml langauge: parameter you will get right language date string on your post(s).


Xuanwo:
@ioxo This feature may cause posts' permalink change

leesei:
Affects permalink, closing before someone came up with a good solution.

Issue #942, opened 2014-12-01
og:url incorrect when site is put under subdirectory


leesei:
`_config.yml`

```
# URL
## If your site is put in a subdirectory, set url as 'http://yoursite.com/child' and root as '/child/'
url: http://yoursite.com/child
root: /child/
```

In HTML's meta

``` html
<meta property="og:url" content="http://yoursite.com/child/child/">
```

apalskoi:
I am having the same issue. I have fixed it for now by adding a replacement for the 'url' option of open_graph in the layout/_partial/head.ejs:

url: url.replace('/blog/blog', '/blog') - my blog is in the '/blog' subdirectory.

``` <%- open_graph({twitter_id: theme.twitter, google_plus: theme.google_plus, fb_admins: theme.fb_admins, fb_app_id: theme.fb_app_id, url: url.replace('/blog/blog', '/blog')}) %>```

**Comment by kflu** [comment]
Same problem here. Problem is facebook won't display images when sharing links that have the invalid og:image tags. I worked it around by commenting those out. 

But it seems the true fix really should be in `node_modules\hexo\lib\plugins\helper\open_graph.js`:

```javascript

  images = images.map(function(path) {
    if (!urlFn.parse(path).host) {
      if (path[0] !== '/') path = '/' + path;
      return config.url + path;
    }

    return path;
  });
```

kflu:
OK. I've fixed this with:

```

  images = images.map(function(path) {
    if (!urlFn.parse(path).host) {
      if (path[0] !== '/') {
        return urlFn.resolve(url, path);
      }
    }

    return path;
  });
```

PR #947, opened 2014-12-06
Open graph fully absolute urls support

guiohm:
- Fixes the issue https://github.com/hexojs/hexo/issues/942
- Allows using fully absolute urls for og image tags (which is mandatory for Facebook).
- Fixes [poor external link detection](https://github.com/guiohm/hexo/commit/b424e11bfd3fd13f5978439eca545f45dbf547af#diff-62e9cf90d8889a8a3819d3cd1341272dL25)

I don't quite understand why the optional subdirectory had to be added on both `config.url`and `config.root`.
I believe this is more relevant now.

### DIFF
--- a/assets/_config.yml
+++ b/assets/_config.yml
@@ -11,7 +11,7 @@ email:
 language:
 
 # URL
-## If your site is put in a subdirectory, set url as 'http://yoursite.com/child' and root as '/child/'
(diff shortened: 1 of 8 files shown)


Xuanwo:
The build was gone for so long that I can't rebuild it, make a new pr or commit may restart it.

stevenjoezhang:
Superseded by https://github.com/hexojs/hexo/pull/4616
`root` option is now optional

PR #1044, opened 2015-02-10
Default to using ISO-8601 date format

Graham42:
We should encourage people use use the ISO standard date by making it the default.

[If we really need a why...](http://xkcd.com/1179/)


leesei:
can't help but :+1:

tommy351:
xkcd convinced me.

Issue #1132, opened 2015-03-23
Duplicated child categories


naokiy:
I found a bug in post setCategories.

If there were two posts categorized like below,

```yaml
category:
- foo
- bar
```

then hexo create two categories named 'bar'.

  * foo ( 2 posts )
    * bar ( 1 post )
    * bar ( 1 post )

joysboy:
I have the same problem

ppoffice:
Same problem, too.

hanxiao:
Same problem for both english and chinese, seems language-independent.

naokiy:
This problem is language-independent.

For detail, please see #1134.

PR #1134, opened 2015-03-23
Fix post.setCategories() to solve #1132

naokiy:
Use CUID to find parent category.

This fix solve #1132.


hanxiao:
Thanks to @naokiy This patch does fix the problem!

PR #1156, opened 2015-03-29
fix child categories duplication issue

ppoffice:
I note that cats[i - 1] refers to the name of parent category, not the CUID, but arr[i - 1] is the CUID of parent category, and Category schema needs parent's CUID to detect category duplication. So I changed cats[i - 1] to arr[i - 1].

Front-matter:
```yml
categories: 
- Front-End
- JavaScript
```

Child categories before fix:
before

Child categories after fix:
after


tommy351:
Duplicated #1134

PR #1171, opened 2015-04-03
Fixes #1088, update test case

### LINKED ISSUE #1088: Conditional tags helper is broken in hexo 3?
It seems `is_category`, `is_tag`, `is_archive` helpers always return `false`. -> #1020 

Test template(always return None in `category`, `tag`, `archive` page):
``` ejs
<% if (is_category()) { %>
<h1>Category: <%= page.category %></h1>
<% } else if (is_tag()) { %>
<h1>Tag: <%= page.tag %></h1>
<% } else if (is_archive()) { %>
<h1>Blog Archive</h1>
<% } else { %>
<h1>None</h1>
<% } %>
```

`permalink` setting
```
permalink: :year/:month/:day/:title/
```

Version information:
```
hexo: 3.0.0
os: Linux 3.19.0-1-uksm-ck linux x64
http_parser: 2.3
node: 0.12.0
v8: 3.28.73
uv: 1.0.2
zlib: 1.2.8
modules: 14
openssl: 1.0.2
```

Also can test with [landscape](https://github.com/hexojs/hexo-theme-landscape)

https://github.com/hexojs/hexo-theme-landscape/blob/master/layout/_partial/head.ejs#L22

leesei:
Changes to be committed:
	modified:   lib/plugins/helper/is.js
	modified:   test/scripts/helpers/is.js

### DIFF
--- a/lib/plugins/helper/is.js
+++ b/lib/plugins/helper/is.js
@@ -59,43 +59,45 @@ function isPostHelper(){
 
 function isArchiveHelper(){
   /* jshint validthis: true */
-  return Boolean(this.archive);
+  return Boolean(this.page.archive);
 }
 
 function isYearHelper(year){
   /* jshint validthis: true */
-  if (!this.archive) return false;
+  var page = this.page;
+  if (!page.archive) return false;
 
   if (year){
-    return this.year === year;
+    return page.year === year;
   } else {
-    return Boolean(this.year);
+    return Boolean(page.year);
   }
 }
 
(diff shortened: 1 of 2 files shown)


leesei:
I think it's better to define a `page` according to the [docs](http://hexo.io/docs/variables.html#Page_Variables) in fixture so that we can use and extend in test cases.

If the schema of the objects in fixture changes, the test case should be able to pick up.

PR #1186, opened 2015-04-08
Fixes #1185 - search now uses 'sitesearch' param instead of two 'q' params.

### LINKED ISSUE #1185: search_form.js plugin does not search for input query
The search form plugin tries to submit a search with two "q" parameters, the first with the keywords one of which is used for "site:http://mysite.com". Google ignores the first "q" so the search is just for everything on the site.

See also this (misplaced) issue on the landscape theme: https://github.com/hexojs/hexo-theme-landscape/issues/28

jamiehumphries:


jamiehumphries:
Fixes #1185

jamiehumphries:
Oops, will fix broken tests this evening...

PR #1204, opened 2015-04-15
More strict error assertion

naokiy:
I found some of test were not working.

For example...

```javascript
  it('name - required', function(){
    return Tag.insert({}).catch(function(err){
      err.should.have.property('message', '`name` is required!');
    });
  });
```

If `Tag.insert({})` doesn't throw any error, `err.should.have.property` will be skipped.
So this test didn't check whether hexo throws error or not.

**modified**

added sinon.spy and calledOnce to check it.

```javascript
  it('name - required', function(){
    var errorCallback = sinon.spy(function(err) {
      err.should.have.property('message', '`name` is required!');
    });

    return Tag.insert({}).catch(errorCallback).finally(function() {
      errorCallback.calledOnce.should.be.true;
    });
  });
```

----

And I found two more problems.

One is problem of test. Please check https://github.com/hexojs/hexo/commit/1483589185118ff3ba832b02a4dadb417bec5009.
The another is https://github.com/hexojs/hexo/commit/f655f594554e8ce574aad8b89448bc695cacc56a

I fixed these to let test pass.

### DIFF
--- a/lib/models/category.js
+++ b/lib/models/category.js
@@ -67,7 +67,7 @@ module.exports = function(ctx){
       parent: parent || {$exists: false}
     });
 
-    if (cat && cat._id === data._id){
+    if (cat){
       throw new Error('Category `' + name + '` has already existed!');
     }
   });
--- a/lib/models/tag.js
+++ b/lib/models/tag.js
@@ -53,7 +53,7 @@ module.exports = function(ctx){
     var Tag = ctx.model('Tag');
     var tag = Tag.findOne({name: name});
 
-    if (tag && tag._id === data._id){
+    if (tag){
       throw new Error('Tag `' + name + '` has already existed!');
     }
   });
--- a/test/scripts/box/box.js
+++ b/test/scripts/box/box.js
@@ -1,6 +1,7 @@
 'use strict';
 
 var should = require('chai').should();
+var assert = require('chai').assert;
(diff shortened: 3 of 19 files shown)


tommy351:
Is this really needed? If there's a validation error in warehouse, it should be thrown in promise.

naokiy:
@tommy351 

For example, there was a test to detect category duplication ( https://github.com/hexojs/hexo/blob/master/test/scripts/models/category.js#L211 ).

But as you know, the duplication was happen ( https://github.com/hexojs/hexo/issues/1132 ). That test did not work.

I think that test not working is a problem.

----

With new assertion, test became failed by two reasons. One reason is wrong test ( https://github.com/hexojs/hexo/commit/1483589185118ff3ba832b02a4dadb417bec5009 ).

Another reason is that duplication detection in model are not working ( https://github.com/hexojs/hexo/commit/f655f594554e8ce574aad8b89448bc695cacc56a ). I think this problem can be solved as a matter of warehouse. But once I fixed in hexo to let the tests pass. 

We can remove or skip that tests. Should we do so?

tommy351:
You should use sinon or callback to make sure the assertion is really working. I don't think it's a great idea to throw an error.

naokiy:
@tommy351 

I used sinon.spy instead of assert.fail() .

PR #1207, opened 2015-04-17
Add test for improve coverage

naokiy:
Add test for coverage.

  * post.create() with data.content
  * post.create() with callback
  * post.publish() with callback


naokiy:
Test was failed at AppVeyor.
I think the reason is https://github.com/tkellen/js-v8flags/issues/17 .

Issue #1236, opened 2015-04-27
Helper is_post() not validating for permalink with language


ahaasler:
Hi,

I was testing a theme I'm making with multilingual support and I realized that the helper `is_post()` is not working properly when the post's permalink contains `:lang`.

I've seen this behavior with `:lang/:category/:title/`.

I'm already fixing it in a fork and I'll send a pull request when it's properly tested. This issue has been opened to keep things tidy and well tracked.

Thanks!

Xuanwo:
@ahaasler  It looks like to be fixed by https://github.com/hexojs/hexo/commit/62f521b32b7e750d88a8e3207278c0f5b1c14fdf

ahaasler:
That's correct, the helper has resolved this with the help from the post generator.

PR #1237, opened 2015-04-28
Fix is_post() helper not validating permalinks with language tags

ahaasler:
I decided to use the regex validator that the W3C defines in
http://www.w3.org/TR/xmlschema11-2/#language for BCP 47 language identifiers.

This fixes hexojs/hexo#1236.

### DIFF
--- a/lib/plugins/helper/is.js
+++ b/lib/plugins/helper/is.js
@@ -44,6 +44,7 @@ function isPostHelper(){
 
   if (!r){
     var rUrl = escapeRegExp(permalink)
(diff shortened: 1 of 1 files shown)


Xuanwo:
We meet a conflict here, I will fix it manully.

Xuanwo:
I find it was fixed, can you have a test?

ahaasler:
Yes, this pull request no longer applies. Since 62f521b32b7e750d88a8e3207278c0f5b1c14fdf the helper uses a different way to check whether a page is a post.

PR #1248, opened 2015-05-01
Add option to indicate current category with class name `current`

WolfgangKluge:
Currently, this is handled by client-script, if needed.


WolfgangKluge:
Coverage decreased due to an edge case that I'm not able to create a test for...

is there a way to create a test that simulates a category overview page? They don't hold a property with the current categories (only the current category name)...

PR #1250, opened 2015-05-01
Remove the extension from internal links.

WolfgangKluge:
Add a new option `permalink_remove_ext_and_dir` to allow links to internal pages and posts without `.html` and without `/index.html` at the end.

For example, instead of a link to `/2015/05/01/test.html` or alternatively to `/2015/05/01/test/`, I want a link to `/2015/05/01/test` without changing the option `permalink` (I want to keep `.html` as a file extension, but not in the URLs).

This needs a change on hexo-server to work with this option, too!

### DIFF
--- a/lib/plugins/helper/url_for.js
+++ b/lib/plugins/helper/url_for.js
@@ -20,6 +20,10 @@ function urlForHelper(path, options){
     return path;
   }
 
+  if (config.permalink_remove_ext_and_dir) {
+    path = path.replace(/(\/index)?\.html$/, '');
+  }
+
   // Resolve relative url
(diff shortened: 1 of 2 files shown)


tommy351:
The setting name `permalink_remove_ext_and_dir` is too long. It should be shrinked.

WolfgangKluge:
any suggestions?

tommy351:
Jekyll uses `permalink: pretty`. I think maybe we can use `pretty_url` or `pretty_permalink`. Any ideas?
https://github.com/jekyll/jekyll/issues/156

PR #1252, opened 2015-05-01
Add whitespace after tagcloud links

WolfgangKluge:
The links need a whitespace so they can wrap


WolfgangKluge:
Can AppVeyor be persuaded to retest? Seems like an error independent to this change..

tommy351:
I think add a `separator` option would be more flexible.
https://codex.wordpress.org/Function_Reference/wp_tag_cloud

Issue #881, opened 2014-10-14
permalinks :title is file path


m80126colin:
I found that `:title`, a variable of permalinks, is NOT only a filename but also a file path under `source/` if you categories your markdown files into several directories.

Xuanwo:
What do you mean?

nt3rp:
I *might* be having a similar problem as @m80126colin (though I am not entirely sure).

One might have settings as follows:

```yml
new_post_name: :year-:month-:day-:title.md # File name of new post
permalink: :year/:month/:day/:title/
```

In this case, new files are created as expected (e.g. `hexo new "Hello World"` creates `2015-01-01-hello-world.md`)

However, what is not expected (or rather, isn't clear how to change the configuration to support this) is that the permalink will not be correct.

Expected:
```
<root>/2015/01/01/hello-world
```

Actual:
```
<root>/2015/01/01/2015-01-01-hello-world
```

I had a blog previously running on Hexo 2.x that worked with this setup, but it seems something has change in Hexo 3.

leesei:
@nt3rp @tommy351 I think we need to update the doc for these config and variables in `Post`

As a side question, how `permalink` is resolved from all these?
Several [`permalink`  related](https://github.com/hexojs/hexo/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+permalink) issues could be resolved once we get these cleared.

PR #1258, opened 2015-05-06
Add new permalink param: `post_title`

nt3rp:
To avoid breaking any existing setups, create a new param which represents the title of a post.

Also, addresses #881 (as far as the problem I described there)

### DIFF
--- a/lib/plugins/filter/post_permalink.js
+++ b/lib/plugins/filter/post_permalink.js
@@ -13,6 +13,7 @@ function postPermalinkFilter(data){
     id: data.id || data._id,
     title: data.slug,
     name: pathFn.basename(data.slug),
+    post_title: util.slugize(data.title, {transform: 1}),
     year: data.date.format('YYYY'),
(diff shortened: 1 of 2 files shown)


lib/plugins/filter/post_permalink.js
```diff
@@ -13,6 +13,7 @@ function postPermalinkFilter(data){
     id: data.id || data._id,
     title: data.slug,
```
nt3rp: I think that the proper fix would be to change this line, to this:

    title: util.slugize(data.title, {transform: 1}),

... but that could break existing setups :disappointed:

tommy351:
What's different between slug and post_title?

nt3rp:
From my understanding, `slug` in the permalink is equivalent to `new_post_name` minus the file extension. `post_title` is the slugized version of the actual post title.

The problem mostly arises in cases like #881.

tommy351:
I see.

PR #1320, opened 2015-06-07
add https

Xuanwo:
you know, in most place of china, we can't reach google with out https.


tommy351:
It's not necessary to add `https`. `//` means use `https` if available.

Xuanwo:
but it always return google.com without https in win7?
my friend told me about this = =.

PR #1350, opened 2015-06-23
New date helper using momentjs 'fromNow' method

borisschapira:
Moment.js provides a quite useful function for relative time, it would be a shame not to use it ;)

### DIFF
--- a/lib/plugins/helper/date.js
+++ b/lib/plugins/helper/date.js
@@ -5,7 +5,7 @@ var isMoment = moment.isMoment;
 var isDate = require('util').isDate;
 
 function output(date, format, lang, timezone){
-  if (date === null) date = moment();
+  if (typeof myVar === 'undefined' || date === null) date = moment();
   if (!isMoment(date)) date = moment(isDate(date) ? date : new Date(date));
 
   if (lang) date = date.locale(lang);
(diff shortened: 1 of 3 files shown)


Xuanwo:
Looks wonderful~
Make it a question for @tommy351

lib/plugins/helper/date.js
```diff
@@ -5,17 +5,21 @@ var isMoment = moment.isMoment;
 var isDate = require('util').isDate;
 
 function output(date, format, lang, timezone){
-  if (date == null) date = moment();
+  if (typeof myVar === 'undefined' || date === null) date = moment();
```
tommy351: `== null` can detect both `undefined` and `null`.

tommy351:
It's not a good idea to write `fromNow` logic in the `output` function. You should separate it.

borisschapira:
@tommy351 Is it better like this, only checking for `format` nullity in output (I dunno how to do it without changing something to `output`) ?

tommy351:
I think you can rename the `output` function to `getDate` or `transformDate` and only let this function check types and apply locale and timezone and format the date in each helpers.

borisschapira:
Better ?

Xuanwo:
@tommy351

PR #1389, opened 2015-07-15
[helper] add log()

leesei:
Good for logging objects in theme and plugin

 Changes to be committed:
	modified:   lib/plugins/helper/debug.js
	modified:   lib/plugins/helper/index.js

### DIFF
--- a/lib/plugins/helper/debug.js
+++ b/lib/plugins/helper/debug.js
@@ -8,9 +8,9 @@ function inspectObject(object, options){
 }
 
 // wrapper to log to console
-function serverLog(){
-  return console.log.call(null, arguments);
+function log(){
+  return console.log.apply(null, arguments);
 }
 
 exports.inspectObject = inspectObject;
-exports.serverLog = serverLog;
+exports.log = log;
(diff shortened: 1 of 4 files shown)


Xuanwo:
Add more coverage test?

leesei:
K, but I think I can only asset true for the `serverLog()` test.

tommy351:
The name `serverLog` does not make sense. This helper just called `console.log`.

leesei:
That's to distinguish between printing `script` and calling browser's
`console.log()`
I used to have the latter function but not finding it very useful.

What's your take on the name?

BTW, can I detect `serve` and `generate` command in helper? I thinking
about automatically disabling this function in commands other than `server`.

tommy351:
If you only want to enable this helper in the server. You should move this helper to https://github.com/hexojs/hexo-server

tommy351:
And I think `log` or `print` would be a better name for this helper.

leesei:
I don't think we can get the server object or call its function in templates, right?

I'll close this PR and reopen a new one with test.

PR #1428, opened 2015-08-03
Allow specifying classes of code blocks to be excluded from processing.

cazgp:
Pandocfilters, for example, uses code block classes to allow extra types of content. This PR allows a user to specify which code blocks (if any) are to be excluded from processing, so that another program can process them instead.


leesei:
Isn't this can be achieved by `lang:plaintext`?

Issue #1440, opened 2015-08-08
Image link not correct at index page when site is under subdirectory


Tomyail:
```
hexo: 3.1.1
os: Darwin 14.4.0 darwin x64
http_parser: 2.3
node: 0.12.7
v8: 3.28.71.19
uv: 1.6.1
zlib: 1.2.8
modules: 14
openssl: 1.0.1p
```
I put a image under the `source` folder and use it like this in my post
```
![image](../image.jpg "title")
```
This works fine when I set the `root ` to  `/` 

But when I set the `root` to `/blog/`,on the `Index` page I got the image link like 

`http://mydomain.com/image.jpg`. so the link not exist.

And by the way,the image link is correct in the post's detail page

PR #1444, opened 2015-08-12
Support root path configure for img tag

liuhongjiang:
Current img tag doesn't prepend root path if it's not a slash. With
this fix, the root path will be prepend to the img src if it's not a
slash.
This fix only works for the hexo tags, does not work for the markdown
syntax of links and images. The issues of markdown links and images should
be fixed in the hexo-renderer-marked module.

Resolves: #1440

### DIFF
--- a/lib/models/post.js
+++ b/lib/models/post.js
@@ -47,7 +47,8 @@ module.exports = function(ctx) {
   });
 
   Post.virtual('permalink').get(function() {
-    return ctx.config.url + '/' + this.path;
+    var url_for = ctx.extend.helper.get('url_for');
+    return ctx.config.url + url_for.call(ctx, this.path);
   });
 
   Post.virtual('full_source').get(function() {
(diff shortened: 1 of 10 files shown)


liuhongjiang:
This pull request only fix the image tag.
For the image and link of markdown syntax, please refer to the pull request:
https://github.com/hexojs/hexo-renderer-marked/pull/7

liuhongjiang:
I test run the test with iojs using the master branch, it gets the same error, I think this may be related to the recent upgrade of iojs.

liuhongjiang:
Add one more commit to fix the problem, when the root path is set to a subdirectory, there is no this root path in the permalink of post.

liuhongjiang:
I rebased the code on the latest master, And it has the same logic as this pull request: https://github.com/hexojs/hexo-renderer-marked/pull/7

@Xuanwo please check it.

Xuanwo:
:+1:

PR #1526, opened 2015-10-01
Fix category path when placed on root

giuseppelt:
This patch fixes the double slashes (i.e. http://site.com//cat1) in the category path, when you configure `category_dir` with '/' or ''

### DIFF
--- a/lib/models/category.js
+++ b/lib/models/category.js
@@ -30,7 +30,8 @@ module.exports = function(ctx) {
 
   Category.virtual('path').get(function() {
     var catDir = ctx.config.category_dir;
-    if (catDir[catDir.length - 1] !== '/') catDir += '/';
+    if (catDir === '/') catDir = '';
+    if (catDir.length > 0 && catDir[catDir.length - 1] !== '/') catDir += '/';
 
     return catDir + this.slug + '/';
   });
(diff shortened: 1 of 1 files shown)


lib/models/category.js
```diff
@@ -30,7 +30,8 @@ module.exports = function(ctx) {
 
   Category.virtual('path').get(function() {
     var catDir = ctx.config.category_dir;
-    if (catDir[catDir.length - 1] !== '/') catDir += '/';
+    if (catDir === '/') catDir = '';
+    if (catDir.length > 0 && catDir[catDir.length - 1] !== '/') catDir += '/';
```
tommy351: `catDir.length > 0` can be `catDir.length`.

giuseppelt: what you mean? I know the evaluation is the same, I used explicit comparison for better reading.
Have I remove `> 0`? If Y, no prob.

Xuanwo: Look like to keep code style?

kbaley:
Wondering when this will be merged. We're looking to go live with a site we've converted from Jekyll and we need this to match our existing URLs.

leesei:
Sorry we are a bit slow in merging the PR.
Please use `npm install` your own repo for your release.

leesei:
Merged via https://github.com/hexojs/hexo/commit/9aca27b2bddaca35d22880d66aa03d77c6114a4e.
Thanks.

PR #1539, opened 2015-10-08
Correct tag trimming

aausch:
fixing a bug with tag trimming which leads to incorrect indentation for some code blocks.
---
currently, when this is put inside of a codeblock:

```
    return result;
  }
  if (tired && night){
    sleep();
  }
```

it will incorrectly be rendered as (approximately, can't reproduce the exact error with the github editor):

```
return result;
  }
  if (tired && night){
    sleep();
  }
```


Xuanwo:
wonderful work!

PR #1557, opened 2015-10-19
Adds swig-lodash support

petergoes:
This adds the support for the `swig-lodash` package. 
All of lodash v2.4.1 functionality as a swig filter.

I added it side-by-side to the, already implemented, `swig-extra` package. The only overlap was the `pluck` filter. So i removed the `swig-extra` version in favour of the `lodash` version.

Once the author of the `swig-lodash` package upgrades to the latest `lodash` version (3.10.1), more overlapping filters can be removed. I will post a pull request to that package as well.

### DIFF
--- a/lib/plugins/renderer/swig.js
+++ b/lib/plugins/renderer/swig.js
@@ -13,7 +13,6 @@ extras.useFilter(swig, 'batch');
 extras.useFilter(swig, 'groupby');
 extras.useFilter(swig, 'markdown');
 extras.useFilter(swig, 'nl2br');
-extras.useFilter(swig, 'pluck');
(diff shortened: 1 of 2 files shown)


leesei:
I think we could wait a bit for `swig-lodash` can replace all `swig-extra` features if the former is superior.

BTW, aren't we using Nunjunks to replace swig a while ago? Where are these swig filters used?

dww510:
Swig has been dead for about two years so Nunjunks is the way to go. Anything swig at this point should be declined. My opinion of course, but that's my vote.

jrbenito:
While installing  npm warns:

> npm WARN deprecated swig@1.4.2: This package is no longer maintained

PR #1580, opened 2015-11-06
Fix issue cannot correctly analyze arguments in tag with space 

winterTTr:
I found that when use the tag with space in parameter, even i wrapped it within a string literal, it do not work. For example:

> {% qnimg "image path with space" %}

when this (custom) tag is process, the arguments will become an array of

> ["image", "path", "with", "space"]

instead of 

>["image path with space"]

This is a big issue for any kind of tag ( no matter hexo native tag or custom tag plugin).

I found that the issue is, hexo simply connect all the lexical node to a string and then split it with space, this is not correct and result in the issue i mention.

So i fix it.

### DIFF
--- a/lib/extend/tag.js
+++ b/lib/extend/tag.js
@@ -77,11 +77,27 @@ NunjucksTag.prototype._parseArgs = function(parser, nodes, lexer) {
   var args = [];
   var token;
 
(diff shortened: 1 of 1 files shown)


winterTTr:
Well, seems you use the code style checker ? OK, let me resubmit again.

PR #1581, opened 2015-11-06
Fix issue cannot correctly analyze arguments with space in tag 

winterTTr:
I found that when use the tag with space in parameter, even i wrapped it within a string literal, it do not work. For example:

> {% qnimg "image path with space" %}

when this (custom) tag is process, the arguments will become an array of

> ["image", "path", "with", "space"]

instead of 

>["image path with space"]

This is a big issue for any kind of tag ( no matter hexo native tag or custom tag plugin).

I found that the issue is, hexo simply connect all the lexical node to a string and then split it with space, this is not correct and result in the issue i mention.

So i fix it.

### DIFF
--- a/lib/extend/tag.js
+++ b/lib/extend/tag.js
@@ -77,11 +77,27 @@ NunjucksTag.prototype._parseArgs = function(parser, nodes, lexer) {
   var args = [];
   var token;
 
-  while ((token = parser.nextToken(true)) && token.type !== lexer.TOKEN_BLOCK_END) {
-    args += token.value;
+  var argarray = new nodes.Array(tag.lineno, tag.colno);
+
+  var argitem = "";
+  while( (token = parser.nextToken(true)) ){
+    if( token.type == lexer.TOKEN_WHITESPACE || token.type == lexer.TOKEN_BLOCK_END ){
+      if( argitem != "" ){
+        var argnode = new nodes.Literal(tag.lineno, tag.colno, argitem.trim());
+        argarray.addChild(argnode);
+        argitem = "";
+      }
+
+      if( token.type == lexer.TOKEN_BLOCK_END ){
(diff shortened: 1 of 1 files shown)


winterTTr:
The failed build error seems not come from my change. So I just leave it without any more change.

winterTTr:
AssertionError: expected Fri, 06 Nov 2015 14:26:59 GMT to deeply equal Fri, 06 Nov 2015 14:26:59 GMT
    + expected - actual
    
    -[Date: 2015-11-06T14:26:59.243Z]
    +[Date: 2015-11-06T14:26:59.000Z]
    
    at C:\projects\hexo\test\scripts\processors\post.js:505:33
    at tryCatcher (C:\projects\hexo\node_modules\hexo-fs\node_modules\bluebird\js\main\util.js:26:23)
    at Promise._settlePromiseFromHandler (C:\projects\hexo\node_modules\hexo-fs\node_modules\bluebird\js\main\promise.js:505:31)
    at Promise._settlePromiseAt (C:\projects\hexo\node_modules\hexo-fs\node_modules\bluebird\js\main\promise.js:581:18)
    at Promise._settlePromises (C:\projects\hexo\node_modules\hexo-fs\node_modules\bluebird\js\main\promise.js:697:14)


Seems this error is not coming from my change? Let me know if this is real my fault.

winterTTr:
I explain more detail of my fix here
http://winterttr.me/2015/11/07/hexo-cannot-use-space-within-tag-arguments/

Xuanwo:
@winterTTr 
I rechack the fault, and it looks like an error on system time.
And sorry for my rebuild operation, no one will see what happened indeed beacuse the changes of `.travis.yml`.
Now I will merge it into master, thank you for your work.

PR #1599, opened 2015-11-20
tag: fix special tag names

magicdawn:
I have a post got tags like this https://gitcafe.com/magicdawn/magicdawn/raw/master/source/_posts/js-constructor-return-value.md

```
tags: 
- js
- constructor
- return
```

Cause `constructor` is a special key. `map['constructor'] === Object`, And will error `str must be string`.


Xuanwo:
@magicdawn failed in 0.12, but passed 4 & 5, maybe it's a problem of travis, I will rebuild it manually.

Issue #1654, opened 2015-12-15
server won't run or connect


tomByrer:
Seems when I run locally, I can't connect via several browsers.  Related to failed dep?
``` CLI

E:\client\Tim\> nodist
  (x64)
> nodev4.2.3  (global)

E:\client\Tim\> npm -v
2.14.10

E:\client\Tim> npm install hexo-cli -g
npm WARN optional dep failed, continuing fsevents@1.0.6
C:\p\n\nodist\bin\hexo -> C:\p\n\nodist\bin\node_modules\hexo-cli\bin\hexo
hexo-cli@0.1.9 C:\p\n\nodist\bin\node_modules\hexo-cli
├── abbrev@1.0.7
├── minimist@1.2.0
├── bluebird@3.0.6
├── tildify@1.1.2 (os-homedir@1.0.1)
├── chalk@1.1.1 (ansi-styles@2.1.0, supports-color@2.0.0, escape-string-regexp@1.0.3, has-ansi@2.0.0, strip-ansi@3.0.0)
└── hexo-fs@0.1.5 (escape-string-regexp@1.0.3, graceful-fs@4.1.2, chokidar@1.4.1)

E:\client\Tim> hexo init blog
INFO  Copying data to E:\client\Tim\blog
INFO  You are almost done! Don't forget to run 'npm install' before you start blogging with Hexo!

E:\client\Tim> cd blog

E:\client\Tim\blog> npm install
npm WARN optional dep failed, continuing fsevents@1.0.6
npm WARN optional dep failed, continuing fsevents@0.3.8
-
> dtrace-provider@0.6.0 install E:\client\Tim\blog\node_modules\hexo\node_modules\bunyan\node_modules\dtrace-provider
> node scripts/install.js
...

hexo server
INFO  Hexo is running at http://0.0.0.0:4000/. Press Ctrl+C to stop.
```

leesei:
Try this URL: http://localhost:4000

tomByrer:
There it is @leesei, cheers!
Leaving open until documented somewhere.

leesei:
This is tracked by https://github.com/hexojs/hexo/issues/1657, which is a more thorough solution.

tomByrer:
Great, thank you.

PR #1656, opened 2015-12-15
add localhost help v2

tomByrer:
Was a minor [pain point](https://github.com/hexojs/hexo/issues/1654) when trying to quickly demo.
Open to other ideas (CLI output).


leesei:
see https://github.com/hexojs/hexo-server/pull/17

Issue #1695, opened 2016-01-04
Helpers should output snippets in HTML5


Imyssed:
Some helpers are still using the old HTML standards,not in HTML5.
For instance:
```html
<%- css('style.css') %>
<!--Output-->
<link rel="stylesheet" href="/style.css" type="text/css">

<%- js('script.js') %>
<!--Output-->
<script type="text/javascript" src="/script.js"></script>
```
Specifying `type` attributes in these tags isn't necessary as HTML5 implies `text/css` and `text/javascript` as defaults. This can be safely done even for older browsers.

Xuanwo:
Closed for pr merged.

PR #1706, opened 2016-01-09
Remove unneccessary refs to type attr in helper output snippets

brycepj:
See Issue #1695


brycepj:
Sounds good -- thank you!

PR #1726, opened 2016-01-25
fixed eslint and jscs linting issues

greg-js:
just some superficial changes to make eslint and jscs happy


lib/plugins/helper/feed_tag.js
```diff
@@ -4,7 +4,7 @@ function feedTagHelper(path, options) {
   options = options || {};
 
   var title = options.title || this.config.title;
-  var type = options.type || 'atom';
+  var type = options.type || 'atom'; // eslint-disable-line no-unused-vars
```
tommy351: If the variable is not used, it should be removed.

Issue #1756, opened 2016-02-17
Proposal to add mark (line highlight) support to code block tag plugin


lotabout:
When writing codes, sometimes we'll need to highlight some lines like:

![selection_035](https://cloud.githubusercontent.com/assets/1527040/13102145/b829b9ee-d586-11e5-84c9-5dd13d290853.png)

As for how to specify the lines, we could follow [octpress' rules](http://octopress.org/docs/plugins/backtick-codeblock/), i.e.

> mark:#,#-# - Mark one or more lines of code with the class name "marked". Accepts one number, numbers separated by commas, and number ranges. Example mark:1,5-8 will mark lines 1,5,6,7,8. Note: If you've changed the beginning line number be sure these match rendered line numbers 

So that we could have something like:

```
{% codeblock mark:1,5-8 %}
code snippet
{% endcodeblock %}
```

As hexo depends on hexo-util for highlighting, I've submit an issue to hexo-util for adding such support. However we still need support from hexo's tag plugin.

greg-js:
+1

greg-js:
I'm just a guy who read this and really wanted this feature for my blog so I made a pretty basic version of it in order to highlight lines in my own code snippets, just figured I'd share

leesei:
Thanks.

leesei:
@greg-js I see the `div` has `class="line marked"`
But I think we need a CSS to actually highlight the line. Did you changed the style in theme?

greg-js:
I didn't, I figured it was up to the theme designers to decide how to style
it. On my own blog I made it dark yellow but I'm using a custom highlight
theme.

leesei:
I see.

leesei:
One more question, the gutter don't have the `mark` class, is it intended (your snapshot seems to also mark the gutter)?

greg-js:
Oh I didn't make or recreate that screenshot as that would have involved rewriting the whole code block. In my implementation, just the code lines are styled because that could be done just by changing a span and line break to a div. 
I'm on holiday in Spain now until early next week but could do some work on completely rewriting the code block when I get back if you'd like.

leesei:
Current:
http://leesei.github.io/hexo-vanilla-site/2016/02/26/1756/

Naive mockup by adding '.marked' to '.gutter.line':
[...]

PR #1766, opened 2016-02-20
add mark option to code blocks

greg-js:
#1756 

Not sure if this is a good enough implementation but I just wanted it for my own blog and figured I'd share, feel free to close the PR if you don't like the style or think it's an unnecessary feature.

### DIFF
--- a/lib/plugins/tag/code.js
+++ b/lib/plugins/tag/code.js
@@ -13,12 +13,13 @@ var rLang = /\s*lang:(\w+)/i;
 var rLineNumber = /\s*line_number:(\w+)/i;
 var rHighlight = /\s*highlight:(\w+)/i;
 var rFirstLine = /\s*first_line:(\d+)/i;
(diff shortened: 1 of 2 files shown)


lotabout:
@greg-js I believe that we still need to patch the `highlight` function in hexo-util to make it work, is that right?

About your implementation,  I think the resulting value of `mark` should contain the ranges such as `[2,4,[6,8]]` instead of expand them into `[2,4,6,7,8]`. The concern is that expanding them may result into a very long list.

greg-js:
That's right and I submitted a PR on hexo-util as well actually.

As for your concern, I hadn't considered that. For my own blog this isn't
an issue, but if there is more interest in this, I will take a look at
doing it in a different way. I'm soon to leave on a week long trip to Spain
soon though so it'd probably be after I come back unless I find time
tomorrow

leesei:
Merged. Thanks.

leesei:
@tommy351 I've updated landscape, please update `hexo-starter`

Issue #1791, opened 2016-03-01
Precompiled renderer will not trigger `after_render` filter


akfish:
Precompilation bypasses `render.render` method:
https://github.com/hexojs/hexo/blob/master/lib/theme/view.js#L106
Which executes `after_render` filter:
https://github.com/hexojs/hexo/blob/master/lib/hexo/render.js#L69

This will cause inconsistent results for plugins that consume `after_render` filter.

leesei:
Since the file has not changed, there's no need to render.
Isn't the modifications committed to the rendered HTML?

akfish:
A plugin could modify rendered results from `after_render` filter. With current precompilation workflow, `after_render` will never be called. This means the *expected modifications* will never happen.

PR #1794, opened 2016-03-01
Execute 'after_render' filter when rendering with precompiled renderer

akfish:
As reported in #1791, if a renderer supports precompilation, the `'after_render'` filter will not be executed after the actual rendering is done. This is different from previous version's behavior and will potentially break plugins that consume this filter.

In previous version, rendering layout `layouts/foo.ejs` will trigger `after_render:html` filter. An important use case for this behavior is that a plugin can inject `<script>` tags once per generated page dynamically based on its content. 

For example, [hexo-math](https://github.com/akfish/hexo-math) scans generated html for MathJax expressions and inject MathJax on-demand from `after_render:html`. With this behavior gone, the alternative would be modifying theme layouts directly, which is too messy and error-prone.

### DIFF
--- a/lib/theme/view.js
+++ b/lib/theme/view.js
@@ -104,15 +104,28 @@ View.prototype._precompile = function() {
   };
 
   if (typeof renderer.compile === 'function') {
-    var compiled = this._compiledSync = renderer.compile(data);
+    var compiled = renderer.compile(data);
+
+    function buildFilterArguments(result) {
+      var output = render.getOutput(ext) || ext;
+      return [
+        'after_render:' + output,
+        result,
+        {
+          context: ctx,
+          args: [data]
+        }
+      ];
+    }
+
+    this._compiledSync = function(locals) {
+      var result = compiled(locals);
+      return ctx.execFilterSync.apply(ctx, buildFilterArguments(result));
+    };
     this._compiled = (function(locals) {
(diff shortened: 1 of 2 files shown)


lib/theme/view.js
```diff
@@ -103,7 +105,14 @@ View.prototype._precompile = function() {
 
   if (typeof renderer.compile === 'function') {
     var compiled = this._compiledSync = renderer.compile(data);
-    this._compiled = Promise.method(compiled);
+    this._compiled = Promise.method(function(locals) {
+      var result = compiled(locals);
```
tommy351: The template engine may be asynchronized. So it should be:

``` js
this._compiled = Promise.method(compiled).then(function(result) {
  var output = render.getOutput(ext) || ext;
  return ctx.execFilter('after_render:'  + output, result, {
    context: ctx,
    args: [data]
  });
});
```

akfish: @tommy351 I see.

tommy351:
It would be better to add support for `compiledSync` function.

akfish:
Added. Along with 2 test specs.
My Travis CI got the wired `npm run test-cov` error at the first time too. 
Restart the build will fix it though.

tommy351:
All tests passed :+1:

PR #1809, opened 2016-03-05
Add a new option `max_depth` to the helper `toc`

dorayx:
An toc option to use headings whose depth is at most maxdepth.


leesei:
Thx

PR #1811, opened 2016-03-06
codeblock tag : escape HTML code otherwise it will be interpreted by the browser

LouisBarranqueiro:
Hey, when highlight.js is disabled, `<` must be escaped otherwise HTML code is interpreted by the browser.

### DIFF
--- a/lib/plugins/tag/code.js
+++ b/lib/plugins/tag/code.js
@@ -3,6 +3,7 @@
 // Based on: https://raw.github.com/imathis/octopress/master/plugins/code_block.rb
 
 var util = require('hexo-util');
+var escapeHTML = util.escapeHTML;
 var highlight = util.highlight;
 var stripIndent = require('strip-indent');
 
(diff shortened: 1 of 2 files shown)


tommy351:
I think not only `<` but other HTML entities should be escaped too. You can try `escapeHTML` method in [hexo-util](https://github.com/hexojs/hexo-util#escapehtmlstr).

LouisBarranqueiro:
`<` is enough to escape an entire HTML code and more efficient than replace all HTML entities.

leesei:
@LouisBarranqueiro it think it's better to conform to standard and escape all the non-safe HTML entities with `escapeHTML()`.

LouisBarranqueiro:
Got it. I'm on it.

leesei:
Our test case is asserting against non-escaped content:
https://github.com/hexojs/hexo/blob/master/test/scripts/tags/code.js#L64
https://github.com/hexojs/hexo/blob/master/test/scripts/tags/code.js#L94

Adding other tag parameters disabled highlight:
https://github.com/hexojs/hexo/blob/master/test/scripts/tags/code.js#L103
https://github.com/hexojs/hexo/blob/master/test/scripts/tags/code.js#L114

LouisBarranqueiro:
Thanks dude

leesei:
Thank *you* :wink:

PR #1827, opened 2016-03-11
keywords meta seo: Added lean option to tag model

raphaklaus:
The keywords meta tag wasn't generating accordingly. Opened issue #1828

In open_graph.js:

```javascript
  var keywords = page.tags;
  ...
  if (keywords && Array.isArray(keywords))
```
keywords wasn't an array, was an queried object (lack of lean option). This way the keywords will never be added to result meta tag.

### DIFF
--- a/lib/models/post.js
+++ b/lib/models/post.js
@@ -76,7 +76,7 @@ module.exports = function(ctx) {
       return item.tag_id;
     });
 
-    return Tag.find({_id: {$in: ids}});
+    return Tag.find({_id: {$in: ids}}, {lean:true});
   });
 
   Post.method('setTags', function(tags) {


gboudreau:
This change creates issues:

```
Unhandled rejection TypeError: .../themes/maupassant/layout/_partial/tag.jade:3
    1| if page.tags
    2|   .tags
  > 3|     for tag in page.tags.toArray()
    4|       a(href=url_for(tag.path))= tag.name
    5| 

page.tags.toArray is not a function
```

How about this change instead, in `libs/plugins/helper/open_graph.js`:

```diff
-var keywords = page.tags;
+var keywords = page.tags.toArray();
```

raphaklaus:
Hi @gboudreau,

thank you for your feedback, here is my considerations:

* This issue occurs specific for your theme. Just drop ``toArray()`` from your ``tag.jade`` and it should work with **this** fix. You can also use the ``list_tags`` helper which do what you need out of the box.  
* About your suggestion, I thought something familiar when I was fixing. But it's a bad approach of design (decapsulated casting), and it solves the issue only locally. Futhermore it's no good with performance.

Since all tests passed (only node v4 on travis-ci got error, but it's probably a config error I think) I am confident with this fix.

leesei:
It seems that many objects in `Post` are not lean.
I do not favor a case by case conversion.

Since this affects themes and some themes *may* be using the non-lean Warehouse object.
This change requires discussion and deserves a major release.

leesei:
@raphaklaus `lib/models/` is part of exposed API to theme. It cannot be changed without theme update.
Closing this PR.

PR #1896, opened 2016-04-17
Added gitlab page support

huangcheng:
Since gitlab page service needs .gitlab-ci.yml file to work, but hexo ignores files likes this during page generating process, so I added some codes to let hexo copy those files into content folder on generating pages.


leesei:
We have the yet to document combo: `include`, `exclude`, `skip_render` to control files in `source/`
https://github.com/hexojs/hexo/blob/master/lib/plugins/processor/asset.js#L110-L120

Put list of globs in `_config.yml` to use them, see
https://github.com/hexojs/hexo/blob/master/test/scripts/processors/asset.js#L58-L71

Issue #1855, opened 2016-03-26
Ignore dot file/folder in "layout/*"


leesei:
It seems Hexo 3.2 breaks [ignoring files](https://hexo.io/docs/themes.html), causing
`TypeError: Cannot read property 'compile' of undefined`.
See #1807, #1793 and many others.

> **source**
> Source folder. Place your assets (e.g. CSS and JavaScript files) here. Hexo ignores hidden files and files or folders prefixed with _ (underscore).
1. ignoring files are processed and causes error  
   
   ``` sh
   touch themes/landscape/layout/.DS_Store  
   touch themes/landscape/layout/.gitkeep
   ```
2. ignoring files in source with `layout/` in path get processed #1793
   since it matched the pattern in `hexo/lib/theme/processors/view.js`
   (I tried to fix but failed)
   
   ``` sh
   touch themes/landscape/source/_css/layout/_sidebar.scss
   ```

---

**Workaround 1**: don't name any folder `layout` in theme (need code change)
https://github.com/LouisBarranqueiro/hexo-theme-tranquilpeak/issues/224#issuecomment-219365465

**Workaround 2**: downgrade to 3.1.1 :cry: 
https://github.com/hexojs/hexo/issues/1793#issuecomment-213907822

buildAll:
I deleted all the .DS_Store files in my hexo project folder but I still got the same 'compile'  issue.

The "hexo generate" command works after I switch to the version 3.1.1 of Hexo.

w0rd-driven:
To reproduce the problem, you need a .DS_Store file in the theme's layout folder (layout/.DS_Store), or at least that's how I was able to reproduce it in the theme I use.

In line 98 of `hexo/lib/theme/view.js`, `var renderer = render.getRenderer(pathFn.extname(this.path));` tries to get the renderer for .DS_Store, as DS_Store is the file extension, and `renderer` returns as undefined.

One way to stop it would be to check `typeof renderer !== 'undefined'` but another would be to stop dot files or other ignores from even being processed upstream.

leesei:
#1851 `.viminfo` also cause error during deploy

leesei:
#1850 `.gitkeep` also cause compile error

AlexKasaku:
This also appears to occur if there is a folder named "layout" within the theme, besides the standard layout folder. I have a theme with a layout folder within assets, which was causing `*.scss` files to get picked up and then create this error.

leesei:
[...]

PR #1910, opened 2016-04-23
#1855: Addresses "TypeError: Cannot read property 'compile' of undefined" 

w0rd-driven:
This checks to make sure an object is returned from `render.getRenderer(ext);`.

The following snippet changes https://github.com/hexojs/hexo/blob/master/lib/theme/processors/view.js#L18 to be something like https://github.com/hexojs/hexo/blob/master/lib/theme/processors/source.js#L28-L35, but there may be cases it doesn't properly account for.
```
exports.pattern = new Pattern(function(path) {
  if (!_.startsWith(path, 'layout/')) return false;

  path = path.substring(7);
  var regex = /(^|\/)[\.]/;
  if (regex.test(path) || common.isTmpFile(path) || ~path.indexOf('node_modules')) return false;

  return {path: path};
});
```

### DIFF
--- a/lib/theme/view.js
+++ b/lib/theme/view.js
@@ -115,6 +115,7 @@ View.prototype._precompile = function() {
     ];
   }
 
+  if (!renderer) return;
   if (typeof renderer.compile === 'function') {
     var compiled = renderer.compile(data);
 


chrisjlee:
Seems simple enough.

Issue #1812, opened 2016-03-06
share link wrong with sub path


Jintin:
Thanks for your great work, I'm Taiwanese too.

Please check out this sample.
https://jintin.github.io/blog/2016/02/28/monthly-github/

The share link and comment link is wrong (one more /blog/).

https://jintin.github.io/blog/blog/2016/02/28/monthly-github/#disqus_thread
https://Jintin.github.io/blog/blog/2016/02/28/monthly-github/

My config is

```
url: https://Jintin.github.io/blog/
root: /blog/
```

tommy351:
It seems `post.permalink` return a wrong URL. I'll fix it later.

leesei:
#1608 may be related

ged1959:
Is this the same problem?

[http://ged1959.github.io/blog-hexo/](http://ged1959.github.io/blog-hexo/)

leesei:
@ged1959 no, you have to set these two correctly

```yaml
## If your site is put in a subdirectory, set url as 'http://yoursite.com/child' and root as '/child/'
url: http://ged1959.github.io/blog-hexo/
root: /blog-hexo/
```

ged1959:
@leesei **Thanks a lot!!!** It works.

sabrinaluo:
same issue!
hexo s, console.log(page.permalink) output is correct
hexo g, console.log(page.permalink) output is wrong

ppoffice:
A quick workaround for you guys:

![Code Screenshot](https://cloud.githubusercontent.com/assets/8849362/14759489/19cf533e-0958-11e6-989a-d499dec7cacf.png)

![Log Screenshot](https://cloud.githubusercontent.com/assets/8849362/14759490/1bbfb332-0958-11e6-814a-1d63ff336aa5.png)

Edit this line:

> https://github.com/hexojs/hexo/blob/master/lib/models/post.js#L59

to:

```js
return require('url').resolve(ctx.config.url, url_for.call(ctx, this.path));
```

PS: I didn't test this when `config.root` is `/`.

meghprkh:
#2048 may be related

wanteracc:
when I shared to my Qzone,the URL is wrong.That is to say,can't open my website through the shared link.

hollsteinm:
Popping in here to say that this issue still exists after I setup a default blog with disqus.

rickygao:
Havs this bug been fixed? The same situation encountered.

stevenjoezhang:
[...]

PR #1911, opened 2016-04-23
Fix the unit test on permalink_root_prefix - virtual

199911:
When the site is put in a subdirectory we have the
`url = 'http://yoursite.com/root';` and
`root = '/root/';`
But the test only set `hexo.config.root` but not `hexo.config.url`.
So the test cannot catch the bug in issue #1812 .

I cannot pass the unit test when I fix the bug.


199911:
The Travis CI build failed because I fix the test. 
There is a bug in current version.

PR #1912, opened 2016-04-23
Fix the wrong share link when the site is in subdirectory

199911:
Fix the bug mentioned in #1812 

When the site is in subdirectory, we are suggest to have set url
as 'http://yoursite.com/child' and root as '/child/'.
But the share link in article footer will become
'http://yoursite.com/child/child/slug'.

Fix the bug by replace '/child/' by '/' in virtual link.

### DIFF
--- a/lib/models/post.js
+++ b/lib/models/post.js
@@ -55,9 +55,10 @@ module.exports = function(ctx) {
   });
 
   Post.virtual('permalink').get(function() {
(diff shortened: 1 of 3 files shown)


Xuanwo:
Please update the test and rebase you code.

199911:
@Xuanwo 
Done

oncletom:
Ace! Any chance to have a patch release of hexo to benefit from this change anytime soon?

PR #1969, opened 2016-05-22
Access data files from source folder

gemedet:
This allows pages in the `source` folder to access properties in the `source/_data` files.

For example, in `source/index.html`:
```
<p>Hello {{site.data.foo.name}}</p>
```


Xuanwo:
My be we can support two different data, one from site and the other from theme.
What do you think about that?

gemedet:
Sure, that'd give us more flexibility. I think it would be a separate PR, since you'd want to access the data from the theme's templates as well.

PR #1976, opened 2016-05-24
Don't add /(?:og:|twitter:)?description/ meta tags if there is no description (invalid HTML)

matthew-andrews:
I believe `content` is a required attribute, so without it and without a description you get invalid markup.
http://htmldoctor.info/reference/html/meta.html

A suggested fix and tests ☺️

### DIFF
--- a/lib/plugins/helper/open_graph.js
+++ b/lib/plugins/helper/open_graph.js
@@ -63,7 +63,10 @@ function openGraphHelper(options) {
     });
   }
 
-  result += meta('description', description);
+  if (description) {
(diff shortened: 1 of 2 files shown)


matthew-andrews:
Hm, if I switch my node version to 0.12.14 and run `npm run test-cov` locally it passes…… :-/

matthew-andrews:
Hm…

matthew-andrews:
Hi I think this pull request is ready for review, assuming of course the feature is welcome? ☺️

PR #2022, opened 2016-06-28
Add support for npm scoped packages as plugins

aredridel:


Xuanwo:
What does this patch for?

tommy351:
Please add tests.

PR #2146, opened 2016-09-06
Return null from hexo.generate

tuck182:
- When using hexo.generate from a gulp task, the following warning is generated:
  `Warning: a promise was created in a handler but was not returned from it` (see https://goo.gl/rRqMUw for details). Explicitly returning null here silences this warning.

### DIFF
--- a/lib/hexo/index.js
+++ b/lib/hexo/index.js
@@ -390,6 +390,7 @@ Hexo.prototype._generate = function(options) {
 
         log.warn('No layout: %s', chalk.magenta(path));
       });
+      return null;
     });
   }).then(function() {
     // Remove old routes


AWulkan:
Just a small question, why did you break apart these?
```
}
);
```

Not that it affects the end result, but it doesn't really follow the code structure.

tuck182:
I couldn't really say. Probably habit, or trying to understand where to put the return. Easy enough to fix though.

tuck182:
There, fixed.

NoahDragon:
Could you please fix the jscs errors?

yoshinorin:
@stevenjoezhang 
The current master branch has many diff with this PR. IMHO, we no need to this merge PR. I think we should close this. How do you think?

SukkaW:
> The current master branch has many diff with this PR. IMHO, we no need to this merge PR. I think we should close this. How do you think?

I believe the original issue is no longer there since Bluebird has updated its implementation to comply with Promise A+ spec (which doesn't require a return value from Promise). We can safely close the PR now.

Issue #2201, opened 2016-10-06
Inserting file contents into post


pipirt:
Is there markdown to insert the contents of a file into a post? Ideally, it would look something like
```
{% inject path/to/file.html %}
```
Similar to how you can insert into a template, how can I inject the contents of `file` into the current markdown file. I understand that markdown isn't for templating, but I feel like there should be a way to do this without building a whole new layout for a single post.

I'd like to do this for easy maintenance and to keep the post.md clutter-free.

NoahDragon:
I don't know if there is a way for markdown, but for the theme template, we could use the `<%- partial('path/to/file') %> ` to insert a whole file.

pipirt:
Yes, but to include a bit of external code in a post, one would have to essentially copy the existing layout into a new layout with the partial inserted where it belongs. What I'm thinking of is the desire to have the post look identical to all the other posts, but simply include some HMTL from an external file. Essentially like `<% include_code filename.bar %>` that injects the actual contents of `filename.bar` into the post rather than inserting the code into a codeblock. It would be useful if you had some nifty widget or a pre-written html document you wanted in the body of your post.

I'm just curious if something like this exists or if people think it maybe should..

unnamed42:
Maybe you need `include_code`, one of the default tag plugins.
See https://hexo.io/docs/tag-plugins.html#Include-Code
But this will result in a &lt;pre&gt;&lt;code&gt; block, cannot be used as a html inserter.

pipirt:
I created a plugin with the functionality to include arbitrary code and text files directly into the content of the page/post.

https://github.com/pipirt/hexo-include

PR #2202, opened 2016-10-07
added inject tag

pipirt:
After looking around a bit and creating the question/feature request https://github.com/hexojs/hexo/issues/2201, I created a tag plugin that will insert the raw contents of a file into a post directly from its markdown file. In this way, you can inject pre-written HTML or JS or **the raw content of _any_ file** into the body of the post without having to fumble around with new layouts to use the layout equivalent `<%- partial('path/to/file') %>` or rewrite exiting files in hexo markdown. To inject any content into a post you simply add this to your markdown:
```
<% inject path/to/file.bar %>
```
All paths are relative to `/source/`. I chose `inject` so as not to get confused with `insert_code`.

I'll write a bit in the docs for it if it's happily merged.


NoahDragon:
Could you please make this functionality as a plugin? The Hexo 3.x tries to split core functions and plugins to make it easier to maintain.

pipirt:
Sure thing! I'll see what I can do.

pipirt:
I've created a simple plugin with the functionality. It is [here](https://github.com/pipirt/hexo-include).

I changed the name to `include` because npm already has a `hexo-inject`.

Issue #2210, opened 2016-10-15
Multiple configuration files.


Lindsor:
Can you make it possible to load multiple configuration files.

For example in jekyll we would run

```shell
jekyll --config=_config.yml,_config2.yml serve
```

Which would pull all the configurations from the first file and then the second (so the seond file would take precedence).

PirtleShell:
I agree that this is a good idea.

We could change the config path logic from [this line](https://github.com/hexojs/hexo/blob/master/lib/hexo/index.js#L56). When `args.config === true` we could send it to a module that returns this same filepath when only one config is present, but writes a composite config file if multiple options are found.

I'd like to help with this if we think it's a good idea?

NoahDragon:
Thanks to @PirtleShell, the feature has been implemented. #2211

PR #2211, opened 2016-10-17
support for multiple config files (#2210)

PirtleShell:
Give the `--config` flag support for multiple config files when generating or serving your site.

Inspired by Jekyll as suggested by @Lindsor in #2210.
## Usage

List the desired config files (JSON or YAML) in a comma-separated list (no spaces) after the config flag:

```
hexo server --config config1.yml,config2.json
```

This will combine all the config files and save the merged settings to `_multiconfig.yml`. The later values take precedence. It works with any number of JSON and YAML files with arbitrarily deep objects.

For instance, in the above example if `foo: bar` is in `config1.yml`, but `"foo": "dinosaur"` is in `config2.json`, `_multiconfig.yml` will contain `foo: dinosaur`.

### DIFF
--- a/lib/hexo/index.js
+++ b/lib/hexo/index.js
@@ -22,7 +22,6 @@ var Router = require('./router');
 var Theme = require('../theme');
 var Locals = require('./locals');
 var defaultConfig = require('./default_config');
-var multiConfigPath = require('./multi_config_path');
 var loadDatabase = require('./load_database');
 
 var libDir = pathFn.dirname(__dirname);
@@ -54,8 +53,9 @@ function Hexo(base, args) {
     init: false
   };
 
+  var multiConfigPath = require('./multi_config_path')(this);
   this.config_path = args.config ? multiConfigPath(base, args.config)
-                                 : pathFn.join(base, '_config.yml');
+                                   : pathFn.join(base, '_config.yml');
 
   this.extend = {
     console: new extend.Console(),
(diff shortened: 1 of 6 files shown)


PirtleShell:
I'm unable to reproduce the appveyor errors on my machine with the same `node` and `npm` versions. This may be a Windows error? Can anyone confirm?

```bash
$ npm install
......
$ node -v
v0.12.16
$ npm -v
3.10.8
$ npm test
......
  812 passing (7s)
  7 pending
```

NoahDragon:
I will take a look at it sometime later to see if I can reproduce the error..

NoahDragon:
I think this issue is caused by the AppVoyer's user permission. It works fine on my computer.

image

We have to fix it before merging, otherwise, this error will repeat every time. 

Could you please try to fix it?

PirtleShell:
I'll see what I can do tonight. In the tests, I check for appropriate warnings by reading the debug log, but it looks like Appveyor doesn't give itself permission to read that file... Thanks for looking into it!

PirtleShell:
Well that's aggravating. The before all hook for the Console tests timed out so all its tests failed.

the failing tests

This PR does not affect the Console submodule. [These same tests passed in the previous commit](https://travis-ci.org/hexojs/hexo/jobs/169843022) and the only things that changed were some semicolons and new lines added to a test file that happens _after_ the console tests are run.

the previous commit before style edits

NoahDragon:
Great work, LGTM.

NoahDragon:
Tested and works well. Good job!

Usage suggestion:
1. Run `hexo g` with config file first;
2. Then run `hexo s` with config file accordingly.

E.g.
```
hexo g --config ./config1.yml ./config2.yml
hexo s --config ./config1.yml ./config2.yml 
```

PR #2262, opened 2016-11-13
Post sort fix

KernelPanicAUS:
I've noticed that the 'next' and 'previous' page attributes were back to front.
This was due to how the posts were sorted in the post generator.
Tests have been updated accordingly.


NoahDragon:
Could you please remove yarn.lock or add it to the gitignore?

KernelPanicAUS:
Done :)

NoahDragon:
After rethinking about this issue, I think this may be varied based on the perspective of view on previous and next.

Let's say we have several posts from newest to oldest:

New<--------------------------------->Old
                     Next<--|-->Prev                         `posts.sort('date')`
                     Prev<--|-->Next                         `posts.sort('-date')`

The `Prev` and `Next` do not have any time directions, either way is okay.

However,  `posts.sort('-date')` has used since Hexo 2.0. I'm going to revert this change as it is starting to break existing themes.

PR #2274, opened 2016-11-18
fix css and js helper to prevent adding .css/.js when the path alread…

aulphar:
fix css and js helper to prevent adding .css/.js when the path already has parameter


NoahDragon:
LGTM. But I don't understand why we need this change?

aulphar:
css and js is often used like `<script src="jquery.min.js?v=1.9.1"></script>` the origin code will append `.js` after the `1.9.1` part

psykzz:
The check should be greater than -1.

There is I doubt there is never going to be a time where ? is the first part of url, buts its still good practice.

(unless of course i'm missing something else)

Issue #2318, opened 2016-12-18
Can't use backtick code block in "blockquote" tag plugin


seaoak:
## Environment Info

```
$ hexo version
hexo: 3.2.2
hexo-cli: 1.0.2
os: Linux 4.4.0-45-generic linux x64
http_parser: 2.7.0
node: 6.9.1
v8: 5.1.281.84
uv: 1.9.1
zlib: 1.2.8
ares: 1.10.1-DEV
icu: 57.1
modules: 48
openssl: 1.0.2j
$
$ npm ls --depth 0
hexo-site@0.0.0 /home/seaoak/blog/test
tqq hexo@3.2.2
tqq hexo-generator-archive@0.1.4
tqq hexo-generator-category@0.1.3
tqq hexo-generator-index@0.2.0
tqq hexo-generator-tag@0.2.0
tqq hexo-include@1.1.0
tqq hexo-renderer-ejs@0.2.0
tqq hexo-renderer-marked@0.2.11
tqq hexo-renderer-stylus@0.3.1
mqq hexo-server@0.2.0
$
$ (cd node_modules/hexo && git log HEAD~1...)
commit 1da7ad5c920fb16e182342c4acda7d8a94af31f9
Author: Junliang Huang <JLHwung@users.noreply.github.com>
Date:   Tue Dec 6 22:48:55 2016 +0800

    appveyor: add node.js 7 testing environment (#2302)
$
$ perl -ne 'print if /^# Writing/.../^# Category/' _config.yml
# Writing
new_post_name: :title.md # File name of new posts
default_layout: post
titlecase: false # Transform title into titlecase
external_link: true # Open external links in new tab
filename_case: 0
render_drafts: false
post_asset_folder: true
relative_link: false
future: true
highlight:
  enable: true
  line_number: true
  auto_detect: false
  tab_replace:
fancybox: false
marked:
  breaks: false

# Category & Tag
$
```

## For BUG

When a backtick code block exists as contents of a "blockquote" tag plugin,
it is translated to a string "undefined" in HTML.

For example, the following markdown text in a post:

    {% blockquote %}
    ```
    This is a test.
    ```
    {% endblockquote %}

is translated to HTML as follows:

    <blockquote>undefined
    </blockquote>

By the way, other inline markdown syntax is translated well.

For example, the following markdown text in a post:

    {% blockquote %}
    _aaa_
    bbb
    
    ```
    This is a test.
    ```
    
    ccc
    {% endblockquote %}

[...]

PR #2319, opened 2016-12-18
Allow backtick code block in "blockquote" tag plugin (#2318)

seaoak:
When backtick code block(s) exist as contents of a "blockquote" tag plugin,
each code block is translated to a string "undefined" in HTML (Issue #2318).

In analyzing markdown source text, while the replacement of these elements
with placeholders are nesting, recoveries from placeholders are executed
only once.  So I modify to repeat the recovery process until all
placeholders are recovered.

### DIFF
--- a/lib/hexo/post.js
+++ b/lib/hexo/post.js
@@ -244,9 +244,25 @@ Post.prototype.render = function(source, data, callback) {
 
   function tagFilter(content) {
     // Replace cache data with real contents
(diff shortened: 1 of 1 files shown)


NoahDragon:
Just curious, why close this PR?

seaoak:
@NoahDragon This PR was just replaced with #2321.  I'm sorry that I made a new PR instead of updating this PR.  I was inexperienced in GitHub at that time.

PR #2321, opened 2016-12-19
Allow backtick code block in "blockquote" tag plugin (#2318)

seaoak:
When backtick code block(s) exist as contents of a "blockquote" tag plugin,
each code block is translated to a string "undefined" in HTML (Issue #2318).

In analyzing markdown source text, while the replacement of these elements
with placeholders are nesting, recoveries from placeholders are executed
only once.  So I modify to repeat the recovery process until all
placeholders are recovered.

NOTE: From Pull Request #2319, I correct the condition of exception (accept empty string as escape content).

### DIFF
--- a/test/scripts/hexo/post.js
+++ b/test/scripts/hexo/post.js
@@ -687,7 +687,7 @@ describe('Post', () => {
     });
   });
 
-  // test for PR [#3573](https://github.com/hexojs/hexo/pull/3573)
+  // test for PR #3573
   it('render() - (disableNunjucks === true)', () => {
     const renderer = hexo.render.renderer.get('markdown');
     renderer.disableNunjucks = true;
@@ -702,7 +702,7 @@ describe('Post', () => {
     });
   });
 
-  // test for PR [#3573](https://github.com/hexojs/hexo/pull/3573)
+  // test for PR #3573
   it('render() - (disableNunjucks === false)', () => {
     const renderer = hexo.render.renderer.get('markdown');
     renderer.disableNunjucks = false;
@@ -717,8 +717,8 @@ describe('Post', () => {
     });
   });
 
-  // test for Issue [#2318](https://github.com/hexojs/hexo/issues/2318)
-  it('render() - allow backtick code block in "blockquote" tag plugn (Issue#2318)', () => {
+  // test for PR #2321
+  it('render() - allow backtick code block in "blockquote" tag plugin', () => {
     const code = 'alert("Hello world")';
     const highlighted = util.highlight(code);
 
--- a/test/scripts/hexo/post.js
+++ b/test/scripts/hexo/post.js
@@ -717,4 +717,26 @@ describe('Post', () => {
     });
   });
 
+  // test for Issue [#2318](https://github.com/hexojs/hexo/issues/2318)
+  it('render() - allow backtick code block in "blockquote" tag plugn (Issue#2318)', () => {
+    const code = 'alert("Hello world")';
+    const highlighted = util.highlight(code);
+
+    const content = [
+      '{% blockquote %}',
+      '```',
+      code,
+      '```',
+      '{% endblockquote %}'
+    ].join('\n');
+
+    return post.render(null, {
+      content
+    }).then(data => {
+      data.content.trim().should.eql([
+        '<blockquote>' + highlighted,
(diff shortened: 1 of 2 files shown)


NoahDragon:
Could you please also add the test case for it?

NoahDragon:
#2318

weyusi:
@seaoak `post.js` has been significantly updated, please rebase.

YoshinoriN:
@weyusi @NoahDragon @hexojs/core 
I tried to rebase this PR to master.
But, current post.js source code changed drastic compare with this PR.

I propose we re-create other PR instead of this PR if @seaoak will not react our mention.

tcrowe:
Good idea @YoshinoriN

weyusi:
@YoshinoriN 
if someone can port

https://github.com/hexojs/hexo/blob/bb9aaa5246d14f9668973f6d5aa07a00932aad36/lib/hexo/post.js#L249-L256

to the current `post.js` that would be great. Before I ping @seaoak, I also tried to port that function, but too bad it's beyond me.

I think that function (or similar) could fix https://github.com/hexojs/hexo/issues/3318 https://github.com/hexojs/hexo/issues/2821

seaoak:
Sorry, I'm too late to response you.
I thank you for picking up this PR!

I was busy over the past year, but quite recently I have a little time to enjoy programming.
If allowed, I'll try to rebase this PR.
Of course, I should catch up with current codes of Hexo.
So it may need about a month or more...

If anymone takes this PR, I'm pleased to leave this PR. 😄

YoshinoriN:
Welcome back seaoka !!
It’s okay. Please don't worry, we patiently waiting for your PR 😄

JLHwung:
@seaoak Please resolve the conflict.

stevenjoezhang:
Any updates on this thread?

seaoak:
I'm sorry I restarted to catch up latest Hexo last month.
I had left from Hexo development for a long time,
it will take a while.

I'll be glad if you give me a chance to update this patch.

seaoak:
I rewrite this patch for latest Hexo.
And also add a test case for this patch.

@hexojs/core Could you please review?

By the way #3318 is not fixed by this patch.
It seems like #2969.

NOTE:
  I did "forced push" to this branch because the code base of Hexo changed drastically.
  If it violates the common manner, I will make new PR (and close this PR).

seaoak:
Travis CI erros only on Node.js 6.x.
The Build options of this PR might be too old.

May I adjust build options of this PR? (Although I'm not familiar with Travis CI...)

SukkaW:
@sseaoak We have already drop node 6 for Travis CI in #3598 

I will trigger a rebuild for you.

curbengh:
> NOTE:
> I did "forced push" to this branch because the code base of Hexo changed drastically.
> If it violates the common manner, I will make new PR (and close this PR).

It's fine to rebase to the latest master (in fact I often do it), which requires forced push.

test/scripts/hexo/post.js
```diff
@@ -717,4 +717,27 @@ describe('Post', () => {
     });
   });
 
+  // test for Issue [#2318](https://github.com/hexojs/hexo/issues/2318)
```
curbengh: I think `// test for Issue #2318` is sufficient without the full link.

curbengh:
> By the way #3318 is not fixed by this patch.
> It seems like #2969.

I created a fix https://github.com/hexojs/hexo-util/pull/110 for those issues. Tested with this PR.

But the fix is quite ugly; oh well, I would suggest user to use `{% blockquote %}` then.

test/scripts/hexo/post.js
```diff
@@ -717,4 +717,27 @@ describe('Post', () => {
...
+  it('render() - allow backtick code block in "blockquote" tag plugn (Issue#2318)', () => {
```
curbengh: ` (Issue#2318)` is not necessary

curbengh:
@seaoak 

I've rebased this PR, so all test pass now. My apology if this come off as a bit rude, I just want to include this PR into v4.

seaoak:
@curbengh Thank you for your review and rebase! 😄 
I agree all your comments and changes.
And I'm glad this patch will be included into v4.

Issue #2198, opened 2016-10-05
title text generated by toc helper is not escaped


unnamed42:
For example, I have:

``` markdown
## Get modulus of (1<<s)-1
```

HTML in main document is correct:

``` html
<h2 id="get-modulus-of-1-s-1">Get modulus of (1&lt;&lt;s)-1</h2>
```

But HTML generated by `toc()` helper is not escaped:

``` html
<li class="toc-item toc-level-2">
    <a class="toc-link" href="#get-modulus-of-1-s-1">
        <span class="toc-text">Get modulus of (1<<s)-1</span>
    </a>
</li>
```

By the way, I'm using `hexo-renderer-markdown-it` as my markdown renderer.

NoahDragon:
We could move the discuss to the PR #2437 .

PR #2437, opened 2017-03-08
One possible fix for #2198

unnamed42:


NoahDragon:
Sorry for the late reply. The code LGTM. Could you please add a test case for it?

PR #2461, opened 2017-03-21
Fix error when tag.render <code>

popomore:


popomore:
Same problem as https://github.com/hexojs/hexo/commit/2ba4e148b8bc159c7d014ae80d7f8f0e99f36dd8

popomore:
It's broken in node7, can you retry it?

popomore:
cc @NoahDragon

NoahDragon:
Great, thank you for the PR. However, I don't clearly understand which scenario will cause the issue?

popomore:
If I write nunjucks template in markdown like

```
`{{help.json()}}`
```

Hexo will render it as a nunjucks helper, so this pr will ignore this case

NoahDragon:
Thanks. Make sense to me.

NoahDragon:
@popomore Could you please have a look at this issue https://github.com/hexojs/hexo/issues/2563 ? Do you have any ideas to fix it?

popomore:
It seems have been fixed?

NoahDragon:
Actually, it's not. It's still under investigating, I'm not sure if it related to this PR.

popomore:
#2568 has fixed it, the bug is just the greedy regexp, it should only match the content between the tag.

seaoak:
@popomore Before your Pull Request #2461 was merged, `lib/extend/tag.js` escaped a `<code>` element only if it was the only child of a `<pre>` element.  Do you know the reason why we can not all `<code>` elements unconditionally?

popomore:
@seaoak Hi, there two style is different in markdown. 

screenshot 2017-05-22 11 59 07
screenshot 2017-05-22 11 59 25

Sorry for the bug I made.

NoahDragon:
@popomore No worries. I think the issue hides there for a long while, it is a good time to actually fix the code.

lsdsjy:
https://github.com/hexojs/hexo/pull/2578#issuecomment-303488627

PR #2478, opened 2017-03-29
Give option to disable Nunjucks when processing posts.

be5invis:
This PR adds...

- An option, `disable_nunjucks`.
- And the corresponded code to skip Nunjucks when this options is set.

### DIFF
--- a/lib/hexo/post.js
+++ b/lib/hexo/post.js
@@ -234,6 +234,7 @@ Post.prototype.render = function(source, data, callback) {
 
   var ctx = this.context;
   var config = ctx.config;
+  var disableNunjucks = config.disable_nunjucks;
   var cache = [];
   var tag = ctx.extend.tag;
   var isSwig = data.engine === 'swig' || (source && pathFn.extname(source) === '.swig');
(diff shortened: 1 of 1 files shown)


NoahDragon:
Thanks for the PR. Since Swig deprecated and Nunjucks is major renderer for tags in Hexo 3.0, this setting could affect all tag plugins ([Hexo tags](https://hexo.io/docs/tag-plugins.html) or third-parties). The impact could be huge. 

One solution is we give up the Nunjuncks as it really causes troubles with math and LeTex symbols. However, adding a system setting doesn't do it once for all.

be5invis:
@NoahDragon This is intensional: if we do NOT need any tag plugin (for me, my Youki can handle them with its macro system), then you can turn the entire thing off.

NoahDragon:
I totally understand your situation. However, this fix is a shortcut to solve the conflict between Nunjucks and other marked languages. We are also working on to resolve the confliction. I would recommend modifying the Hexo version in your package.json to: 
```
"hexo": ",https://github.com/be5invis/hexo.git"
```
So that you could keep this change for your site, we also could move on our changes without worrying about the new setting.

PR #2520, opened 2017-04-18
Fix multiple config issue #2518

NoahDragon:
Thank you for creating a pull request to contribute to Hexo code! Before you open the request please review the following guidelines and tips to help it be more easily integrated:

- [x] Add test cases for the changes.
- [x] Passed the CI test.

PR #2521, opened 2017-04-18
fixes Hexo multi-config

### LINKED ISSUE #2499: multi-config fails during Hexo initialization if it tries to log
## Environment Info

Node version(`node -v`): 6.9.1

Hexo and Plugin version(`npm ls --depth 0`): 3.3.0

## BUG

When creating a Hexo object, if anything in multi_config_path causes a log attempt, it fails with:
```
[21:12:02] TypeError: Cannot read property 'w' of undefined
    at multiConfigPath (.../node_modules/hexo/lib/hexo/multi_config_path.js:27:12)
    at new Hexo (.../node_modules/hexo/lib/hexo/index.js:57:36)
```

This is because multi_config_path initialization happens in the Hexo constructor before the log object is created and assigned to `this` (which is what's passed to multi-config as `ctx`).

devpaul:
`multi_config_path` depends on the passed context to have a `log()` method available to it; however, Hexo doesn't initialize `log` until later. This change moves the initialization of `log` above `multi_config_path`.

Resolves #2499 and adds tests to ensure proper integration of Hexo w/ multi_config_path

### DIFF
--- a/lib/hexo/index.js
+++ b/lib/hexo/index.js
@@ -53,6 +53,8 @@ function Hexo(base, args) {
     init: false
   };
 
+  this.log = logger(this.env);
+
   var multiConfigPath = require('./multi_config_path')(this);
   this.config_path = args.config ? multiConfigPath(base, args.config)
                                  : pathFn.join(base, '_config.yml');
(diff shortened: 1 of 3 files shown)


NoahDragon:
Thanks for the PR. I have created a similar PR #2520. I will add your test cases, and merge them together.

PR #2535, opened 2017-04-25
Allow open_graph generates meta tag "keywords" by page.keywords, page.tags or config.keywords

restran:
The open_graph generates HTML meta tag "keywords" only according to page.tags. Allow HTML meta tag "keywords" generated in order of page.keywords, page.tag and config.keywords could be more flexible. 

```
// mypost.md
// custom keywords for html meta tag "keywords"
keywords: [web, site, hexo]
tags: [web, site]
```

If no keywords or tags specified in page, keywords in _config.js will be used. 
```
// _config.js
keywords: blog, hexo, site
```

```js
var keywords = page.keywords || (page.tags && page.tags.length ? page.tags : undefined) || config.keywords;
```

I also check the type of keywords to support string or array.

```js
  if (keywords) {
    if (typeof keywords === 'string') {
      result += meta('keywords', keywords);
    } else if (keywords.length) {
      result += meta('keywords', keywords.map(function(tag) {
        return tag.name ? tag.name : tag;
      }).filter(function(keyword) {
        return !!keyword;
      }).join());
    }
  }
```

I also add escapeHTML to the content of meta tag in consider of avoiding XSS issue. These characters ' "
 <> & / in content of meta tag should be escaped.

```js
var escapeHTML = util.escapeHTML;

function meta(name, content, escape) {
  if (escape !== false) {
    content = escapeHTML(content);
  }

  return htmlTag('meta', {
    name: name,
    content: content
  }) + '\n';
}

function og(name, content, escape) {
  if (escape !== false) {
    content = escapeHTML(content);
  }

  return htmlTag('meta', {
    property: name,
    content: content
  }) + '\n';
}
```

- [x] Passed the CI test.


restran:
@NoahDragon OK, i will do it later.

NoahDragon:
@restran Thank you. 👍

restran:
@NoahDragon Please review.

NoahDragon:
LGTM. Thanks for the contribution.

Issue #2563, opened 2017-05-13
Nunjucks enclosed by two backtick-delimited strings fails.


jo4x962k7JL:
e.g.

```
`backtick test1`
{% blockquote %}
Blockquoe test1
{% endblockquote %}

`backtick test2`
{% blockquote %}
Blockquoe test2
{% endblockquote %}
```

will get the result like:
![2017-05-13 11 10 05](https://cloud.githubusercontent.com/assets/26728779/26026819/3b49620e-3835-11e7-8c11-8a8a09b0fa3c.png)

Any idea to solve this?

NoahDragon:
This is definitely something need to be fixed.

lijxug:
I encountered the same bug while using {% asset_img img.jpg %}

seaoak:
I found that this issue is caused by a latent bug.

The Pull Request #2461 (commit 5134cba97bed6ea83f1695bf054f08dc3f5f72ac) made the bug easier to be realized.

To fix the bug, change a regexp to be non-greedy.

https://github.com/hexojs/hexo/blob/master/lib/extend/tag.js#L65

```
diff --git a/lib/extend/tag.js b/lib/extend/tag.js
index 1222b01..372bd9c 100644
--- a/lib/extend/tag.js
+++ b/lib/extend/tag.js
@@ -62,7 +62,7 @@ Tag.prototype.render = function(str, options, callback) {
   var env = this.env;

   return new Promise(function(resolve, reject) {
-    str = str.replace(/(?:<pre>)?<code.*>[\s\S]*?<\/code>(?:<\/pre>)?/gm, escapeContent);
+    str = str.replace(/(?:<pre>)?<code.*?>[\s\S]*?<\/code>(?:<\/pre>)?/gm, escapeContent);
     env.renderString(str, options, function(err, result) {
       if (err) return reject(err);
       resolve(result.replace(rPlaceholder, function() {
```

I will add test cases, and make a pull request.

Issue #2575, opened 2017-05-20
Code Block parse error


SawyerSun:
## Environment Info

Node version(`node -v`): V7.10.0

Your site `_config.yml` (Optional):

Your theme `_config.yml` (Optional): 

Hexo and Plugin version(`npm ls --depth 0`): 

exo-site@0.0.0 /Users/sp/oneDrive/blog
├── hexo@3.3.5
├── hexo-deployer-git@0.3.0
├── hexo-generator-archive@0.1.4
├── hexo-generator-baidu-sitemap@0.1.2
├── hexo-generator-category@0.1.3
├── hexo-generator-feed@1.2.0
├── hexo-generator-index@0.2.1
├── hexo-generator-search@2.1.1
├── hexo-generator-sitemap@1.1.2
├── hexo-generator-tag@0.2.0
├── hexo-renderer-ejs@0.3.0
├── hexo-renderer-marked@0.3.0
├── hexo-renderer-stylus@0.3.3
├── hexo-server@0.2.0
└── hexo-wordcount@2.0.1

## For BUG
I have some code in markdown like this:
<img width="417" alt="1" src="https://cloud.githubusercontent.com/assets/13533065/26276140/eba72e74-3da3-11e7-866e-8f7d9862db31.png">
However, the code will be parsed to :
<img width="650" alt="2" src="https://cloud.githubusercontent.com/assets/13533065/26276142/f76acc20-3da3-11e7-8dd9-27f93f479a69.png">

So, is there any idea to solve it?

NoahDragon:
Have you tried the 3.3.6 version? If it still not fix your issue, please lock the hexo version to 3.2.2. It probably related to a know issue: #2573

popomore:
I don't think it's relative to this issue

@SawyerSun can you give some example to reproduce the bug?

NoahDragon:
similar to https://github.com/hexojs/hexo/issues/2574 issue.

PR #2578, opened 2017-05-23
Revert "Fix parsing error about <code> tag (Issue #2563)"

NoahDragon:
Reverts hexojs/hexo#2568

### DIFF
--- a/lib/extend/tag.js
+++ b/lib/extend/tag.js
@@ -62,7 +62,7 @@ Tag.prototype.render = function(str, options, callback) {
   var env = this.env;
 
   return new Promise(function(resolve, reject) {
-    str = str.replace(/(?:<pre>)?<code.*>[\s\S]*?<\/code>(?:<\/pre>)?/gm, escapeContent);
+    str = str.replace(/<pre><code.*>[\s\S]*?<\/code><\/pre>/gm, escapeContent);
     env.renderString(str, options, function(err, result) {
       if (err) return reject(err);
       resolve(result.replace(rPlaceholder, function() {
--- a/lib/extend/tag.js
+++ b/lib/extend/tag.js
@@ -1,6 +1,5 @@
 'use strict';
 
-var assert = require('assert');
 var stripIndent = require('strip-indent');
 var nunjucks = require('nunjucks');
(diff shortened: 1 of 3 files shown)


NoahDragon:
@seaoak I'm reverting the changes in #2461 and #2568 for a hotfix. Could you please help me review and approve the changes?

popomore:
Why do you revert these commit? Is there any bug exist?

popomore:
I try to look at the vue issue tomorrow, it seems still broken.

NoahDragon:
@popomore Yes, although reverting the #2568 could fix the generating error, the parsing error https://github.com/hexojs/hexo/issues/2575 would remain.

popomore:
Vue is using `raw` tag in markdown, I don't know wheter it is a suitable usage.

https://github.com/vuejs/vuejs.org/blame/master/src/v2/guide/render-function.md#L556

And I found `<pre v-if="!result.staticRenderFns.length"><code>{{ result.staticRenderFns }}</code></pre>` which is not matched by our regexp, so the [assertion](https://github.com/hexojs/hexo/blob/master/lib/extend/tag.js#L62) throws.

popomore:
If I change it to `(<pre.*?>)?<code.*?>([\s\S]*?)<\/code>(<\/pre>)?/`, and it works.

seaoak:
I agree this simple reverting.  These two patches **break** existing user's build.  I think that first priority shuld be given to wind down user's confusion.

By user's responses, it is revealed that we did't know the precise impact of these patches.  So I think that we should not add any modification to this function until we understand this function more precisely (and add enough test cases for compatibility).

seaoak:
Question:

In general, when we make a new release of a product, it is important whether a user's build depending on a latent bug should be kept or be subject to modification.

I basically think that a *minor* release should not break such a build.  At least both changelogs and appropriate error messages are necessary.  This is because users may think it as a regression and become distrustful of the product.  Of course, there are exceptions like a security fix.

Are there policies for such a decision in Hexo?

NoahDragon:
@seaoak As far as I know, we don't have a strict policy on version releases. But I agree with you, we should test our program more carefully before release it to public.

PR #2580, opened 2017-05-23
Fix match attribute in pre tag when render hexo tag

popomore:

- [ ] Passed the CI test.

Relate https://github.com/hexojs/hexo/pull/2578#issuecomment-303488253

### DIFF
--- a/lib/extend/tag.js
+++ b/lib/extend/tag.js
@@ -68,7 +68,7 @@ Tag.prototype.render = function(str, options, callback) {
   var env = this.env;
 
   return new Promise(function(resolve, reject) {
-    str = str.replace(/(<pre>)?<code.*?>([\s\S]*?)<\/code>(<\/pre>)?/gm, escapeContent);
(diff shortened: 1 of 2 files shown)


seaoak:
This new regexp still marks *any* `<code>` element as a target which will be escaped, regardless whether it is wrapped by a `<pre>` element or not.  So, why can we not use simpler regexp such as follows?

```
/<code.*?>([\s\S]*?)<\/code>/
```

stevenjoezhang:
Any updates on this thread?

SukkaW:
@stevenjoezhang IMHO, the fix here is unnecessary.

Hexo only highlight code with `code()` tag and `backtick_code_filter`. Markdown renderer only wrapped code block inside `<pre><code>` (which means no attribute will be added to `<pre>` tag during post rendering process).

However, since `code()` is a tag helper while `backtick_code_filter` already escaped swig tag, there is no need for this fix then.

stevenjoezhang:
I researched these related issues, and one reason for the problem is the use of `{% raw %}` tags. This PR can be considered to be superseded by https://github.com/hexojs/hexo/pull/3459.

PR #2587, opened 2017-05-28
Lazy load image support for Next theme.

ivan-nginx:

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/tag/img.js
+++ b/lib/plugins/tag/img.js
@@ -40,6 +40,7 @@ module.exports = function(ctx) {
     var height;
     var title;
     var alt;
+    var lazysrc = '';
     var src;
     var item = '';
     var i = 0;
@@ -85,7 +86,14 @@ module.exports = function(ctx) {
       alt = match[2];
     }
 
+    // Find lazy load classname
+	  if (classes[0] === 'lazy') {
+  		lazysrc = src;
+  		src = '/images/loading.gif';
+  	}
+
     var attrs = {
+      'data-original': lazysrc,
       src: src,
       class: classes.join(' '),
       width: width,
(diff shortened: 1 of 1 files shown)


ivan-nginx:
@NoahDragon > Please fix the errors in Travis-ci and add test cases.
I not sure how do it. Can u show?

> Moreover, could this change be more generic, like based on config file?
What are u mean? Need to write something like docs or what?

ivan-nginx:
@NoahDragon all possible errors was fixed.
About second question: this is lazy image load not instantly. So, original lazy source need `data-original` attribute to take saved image and load it later at scroll.
If u may give any suggestion about lazy loading, i listen. But for now i don't see any gooder variants to do this.

NoahDragon:
@ivan-nginx Sorry for the late reply. I missed this PR in the pile of notifications. 
The test cases could be added into https://github.com/hexojs/hexo/blob/master/test/scripts/tags/img.js 

IMO, the string `lazy`, `/images/loading.gif`, and etc. are hard coding, which forces others to use the same string, and may break other themes using the same syntax. Probably write those string into config file, and using the `config.lazy_load_class` `config.lazy_load_path` to access them.

ivan-nginx:
@NoahDragon lazy load path — it's good idea, ok. But lazy load class? This is to big and how about speed generation? For this little option add many other cases?

Can u show me how add test cases? I don't add any cases before.

NoahDragon:
@ivan-nginx Could you please follow the samples in the test file https://github.com/hexojs/hexo/blob/master/test/scripts/tags/img.js . It's using the Mocha as the test framework, you could find more info on their website https://mochajs.org/.

PR #2593, opened 2017-06-02
disable Nunjucks when the renderer spcify that

be5invis:
This PR adds Nunjucks disabling if the renderer claim that it can handle “tags” on its own.
Currently it is not active (i.e., Nunjucks would work) for all existing renderers.

### DIFF
--- a/lib/hexo/post.js
+++ b/lib/hexo/post.js
@@ -33,7 +33,7 @@ function Post(context) {
   this.context = context;
 }
 
-Post.prototype.create = function (data, replace, callback) {
+Post.prototype.create = function(data, replace, callback) {
   if (!callback && typeof replace === 'function') {
     callback = replace;
     replace = false;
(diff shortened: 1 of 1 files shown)


be5invis:
@NoahDragon Hold on.

be5invis:
@NoahDragon Finish.

NoahDragon:
IMO, give this decision to renderers is better than to the users. https://github.com/hexojs/hexo/pull/2478

NoahDragon:
I'm thinking extend this function to post front matters. Instead of site-wide configuration variable, each post can disable Nunjucks when specified. Do you think this will be feasible and compatible with your change?

https://github.com/hexojs/hexo/issues/2622

9662:
Sorry, could you please provide an example of how would a renderer use this feature?

be5invis:
@9662 
I have a custom renderer which is completely not MarkDown and Nunjucks notation may conflict with some of its inputs. That's (originally) why I created this PR.

PR #2602, opened 2017-06-09
enable async tags to have more nested async tags

brian-mann:
nunchucks supports nested async tags.

this PR simple uses a callback function to properly chain these tags together


NoahDragon:
Could you please fix the travis-ci errors?

PR #2630, opened 2017-06-30
include og:locale

tilsammans:
This OpenGraph property is optional, but generally recommended. Since Hexo has i18n it's easy to support and will provide extra detail.

The `og:locale` property content will be set according to: (in order)

1. the `language` option to the open graph helper;
2. the page lang/language;
3. the site config.

When no language is set anywhere, the og:locale attribute is not set. In that case, the OpenGraph specification uses `en_US` as default.

``` html
<meta property="og:locale" content="zh-tw">
```

- [x] Added test cases for the changes.
- [x] Passed the CI test.


tilsammans:
🙇

PR #2641, opened 2017-07-08
chore(package-lock): add package lock

JLHwung:
Added package lock generated from npm 5.

We can discuss on whether to add package lock file. If we agree on the lockfile, I will add package-log to every hexo-* repository.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,3256 @@
+{
+  "name": "hexo",
+  "version": "3.3.7",
+  "lockfileVersion": 1,
+  "dependencies": {
+    "abab": {
+      "version": "1.0.3",
+      "resolved": "https://registry.npmjs.org/abab/-/abab-1.0.3.tgz",
+      "integrity": "sha1-uB3l9ydOxOdW15fNg08wNkJyTl0=",
+      "optional": true
+    },
+    "abbrev": {
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.0.tgz",
+      "integrity": "sha1-0FVMIlZjbi9W58LlrRg/hZQo2B8="
+    },
+    "acorn": {
+      "version": "2.7.0",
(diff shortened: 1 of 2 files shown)


NoahDragon:
IMHO, the package lock should maintain by the user. We could keep all dependencies are compatible with Hexo, and assume they are following the semver standard.

JLHwung:
Hi NoahDragon, honestly I have different opinions on package lock. I would explain in the following.

CI guards our builds is working with the dependencies on every building datetime T, but our package user will definitely use our package in a datetime t > T. During this time gap a new semver version of our dependencies might be published but untested. If our release cycle is long, say monthly, this will leave user vulnerable to unexpected dependencies bump.

The dependencies of our package's dependencies is uncontrolled unless `shrink-wrap` is used, that is, even `1.2.3` is used in our dependencies specification, the dependency of this `1.2.3` package will be slightly different over time.

The core concept of package-lock is to make `npm` a pure function dependent of time. Once the CI share exactly same dependencies tree with our user, then we can ship our tested packages with firm confidence.

NoahDragon:
Because Hexo depends on many other packages that developed by us like hexo-fs and hexo-util, each time when releasing a dependency package the Hexo main repo has to update due to the lock file (also has to publish a new version on npm, otherwise installing via npm won't get the latest lock file). This just increases the maintenance difficulty.

Actually, you almost convinced me to merge this PR, but when I publish the new version of hexo-fs to npm, I realize that each time we fix a dependency, we have to publish two npm package (one for the dependency, one for updated lock file in Hexo main repo). This is good and meets the industry standard, however, I don't think it will fit in Hexo for now, as lack of maintainers. 

This is just my own opinion, please feel free to discuss.

JLHwung:
OKay. If we are lack of maintainers and would not consider package lock, I suggest add package-lock.json to gitignore so that devs using npm5 would not be bothered by an untracked package-lock.

Also I would raise a PR to remove package-lock on hexo-fs.

Issue #2426, opened 2017-03-01
"TypeError: this.relative_url is not a function" occurs when "relative_link: true"


seaoak:
Using the latest `master` branch of [hexojs/hexo](https://github.com/hexojs/hexo) repository:    
hexojs/hexo@5a7ea3e988dd6a0165792aef6ea3febad2897aff

When just only `relative_link: true` is modified from the default initial settings (`_config.yml`), `hexo generate` command shows an error message and *normally finishes*.

The procedure for reproducing this issue is as follows:

```bash
$ hexo init blog
$ cd blog
$ npm install
$ npm install deep-assign
$ npm ls --depth 0
hexo-site@0.0.0 /home/foobar/blog
tqq deep-assign@2.0.0 extraneous
tqq hexo@3.2.2
tqq hexo-generator-archive@0.1.4
tqq hexo-generator-category@0.1.3
tqq hexo-generator-index@0.2.0
tqq hexo-generator-tag@0.2.0
tqq hexo-renderer-ejs@0.2.0
tqq hexo-renderer-marked@0.2.11
tqq hexo-renderer-stylus@0.3.1
mqq hexo-server@0.2.0

npm ERR! extraneous: deep-assign@2.0.0 /home/foobar/blog/node_modules/deep-assign
$ cd node_modules
$ rm -rf hexo
$ git clone ssh://git@github.com:22/hexojs/hexo.git
$ cd hexo
$ git status
On branch master
Your branch is up-to-date with 'origin/master'.
nothing to commit, working directory clean
$ git log -1 --oneline | cat
5a7ea3e Replace new lines by spaces in description (#2354)
$ cd ../..
$ hexo version
hexo: 3.2.2
hexo-cli: 1.0.2
os: Linux 4.4.0-64-generic linux x64
http_parser: 2.7.0
node: 7.6.0
v8: 5.5.372.40
uv: 1.11.0
zlib: 1.2.11
ares: 1.10.1-DEV
modules: 51
openssl: 1.0.2k
icu: 58.2
unicode: 9.0
cldr: 30.0.3
tz: 2016j
$ grep '"version":' node_modules/hexo*/package.json
node_modules/hexo-front-matter/package.json:  "version": "0.2.3"
node_modules/hexo-fs/package.json:  "version": "0.1.6"
node_modules/hexo-generator-archive/package.json:  "version": "0.1.4"
node_modules/hexo-generator-category/package.json:  "version": "0.1.3"
[...]

PR #2656, opened 2017-07-18
fix relative_url reference

dranovsk:
fix for "TypeError: this.relative_url is not a function" error, issue #2426

- [ ] Passed the CI test.

### DIFF
--- a/lib/plugins/helper/url_for.js
+++ b/lib/plugins/helper/url_for.js
@@ -2,6 +2,7 @@
 
 var url = require('url');
 var _ = require('lodash');
+var relative_url = require('./relative_url');
 
 function urlForHelper(path, options) {
   path = path || '/';
@@ -25,7 +26,7 @@ function urlForHelper(path, options) {
 
   // Resolve relative url
   if (options.relative) {
-    return this.relative_url(this.path, path);
+    return relative_url(this.path, path);
   }
 
   // Prepend root path


seaoak:
Thank you for the PR!

I'm sorry, but I think that, if this patch is necessary, other helper functions should be modified in the same manner.

`this.relative_url()` is a common way to call other helper function in a helper function:

- https://github.com/hexojs/hexo/blob/master/lib/plugins/helper/css.js#L16
- https://github.com/hexojs/hexo/blob/master/lib/plugins/helper/date.js#L47
- https://github.com/hexojs/hexo/blob/master/lib/plugins/helper/favicon_tag.js#L4

So I think that the issue #2426 is not caused by the `url_for()` itself.  Rather, it would be caused by the way to call the `url_for()`.  We might need to correct `lib/models/post.js`.

NoahDragon:
@seaoak Thanks for pointing this out. 

@dranovsk Have you reproduced the  #2426 issue, and this fix resolved it?

seaoak:
I found a solution for the issue #2426, but it's a bit *hacky*.

```
$ git checkout HEAD~
$
$ git log -1 --oneline | cat
e54f5e1 Update package.json
$
$ vi lib/models/post.js
$
$ git diff | cat
diff --git a/lib/models/post.js b/lib/models/post.js
index c14afb0..cf83a33 100644
--- a/lib/models/post.js
+++ b/lib/models/post.js
@@ -55,9 +55,9 @@ module.exports = function(ctx) {
   });

   Post.virtual('permalink').get(function() {
-    var url_for = ctx.extend.helper.get('url_for');
+    var self = _.assign({}, ctx.extend.helper.list(), ctx);
     var config = ctx.config;
-    var partial_url = url_for.call(ctx, this.path);
+    var partial_url = self.url_for(this.path);
     return config.url + _.replace(partial_url, config.root, '/');
   });
$
```

I found that only `lib/models/post.js` calls `ctx.extend.helper.get()` directly in the `hexo` package.  So I just have to correct `lib/models/post.js`.

I referred to `lib/theme/view.js` which enables helper functions in the context of its subroutines (such as `render()`).

Do you have better idea?

NoahDragon:
@seaoak Excellent! Thanks for digging into this issue. Let me have a look first. I will come back to you.

NoahDragon:
@seaoak I think you are right. I will revert this PR, and create new PR regarding your suggestion.

PR #2670, opened 2017-07-27
When promisifying, store does not preserve disableNunjucks property

tcr:
This was causing difficulties in a less-maintained renderer I was using, but I think this may also count as a bug upstream.


JLHwung:
Hmm, I don't think `Promise.promisify` will discard property of a given `nodeFunction`. You can check on the Chrome console of http://bluebirdjs.com/docs/getting-started.html.
```
f = function() {}
f.foo = "bar"
// returns "bar"
Promise.promisify(f).foo 
```

tcr:
You're correct! It's only Promise.method that doesn't preserve the property.

NoahDragon:
@tcr Is that possible to pass in all the configuration to the sync process? If later on, new configuration needed/added, our goal is to not modify the same piece of code again.

tcr:
@NoahDragon Would that be a proposal that instead of attaching disableNunjucks to the rendering function, it's passed in as an extra argument (or something) to the register method?

One way I can think of is to change `sync` to check if it's a boolean (former behavior), and if it's an object e.g. `{sync: true, disableNunjucks: true}` treat it as an options object. Or, just add another argument to the render method to accept an object hash.

The change would impact #2593 in either case—let me know if you think one of these is a good idea. cc @be5invis

NoahDragon:
@tcr Sorry for the late reply. I would prefer the old approach, which `check sync if it's boolean`. We could review later how much impacts to the #2593 change.

be5invis:
@NoahDragon 
Whether enabling Nunjucks could be a "OR" of multiple sources:
 - Whether user specifies to disable it in the entire site
 - Whether user specifies to disable it in for a specific post
 - Whether the renderer said that it can handle everything on its own

JLHwung:
I raised an issue (https://github.com/petkaantonov/bluebird/issues/1484) to request retaining custom properties on `Promise.method`. Before the decision is made, hexo is not looking forward to implement it is own `copyDescriptors`.

stevenjoezhang (dismissed):
👍

SukkaW (approved):
Since Bluebird is no longer actively maintained, I guess we can accept this workaround for now.

We could, however, replace `Bluebird.method` with `Bluebird.promisfy` in the future.

stevenjoezhang:
Found another patch: https://github.com/hexojs/hexo/pull/4498
We could revert #4498 since the solution in this PR is simpler
See also https://github.com/hexojs/hexo/issues/3860

stevenjoezhang (approved):

PR #2678, opened 2017-07-31
Actually allow plugins to be stored as devDependencies

jd-carroll:
Plugins should be sourced from all available options, meaning both dependencies and devDependencies.  In the current implementation (#2558), plugins are only loaded from devDependencies *if there are no dependencies*.

### DIFF
--- a/lib/hexo/load_plugins.js
+++ b/lib/hexo/load_plugins.js
@@ -31,9 +31,10 @@ function loadModuleList(ctx) {
     // Read package.json and find dependencies
     return fs.readFile(packagePath).then(function(content) {
       var json = JSON.parse(content);
-      var deps = json.dependencies || json.devDependencies || {};
+      var deps = Object.keys(json.dependencies || {});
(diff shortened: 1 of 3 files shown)


JLHwung:
I see several Cut the xxx release commit. If this is for test purpose only, could you drop these commits? Thank you.

NoahDragon:
I agree with @JLHwung . The change hasn't merged yet, any additional changes need re-review. Moreover, the version number should maintain by Hexo core team. Thanks for the contribution.

jd-carroll:
Appologies, that's my fault was testing something complete unrelated on a different branch didn't realize it was also updating this.  I will revert these changes.

jd-carroll:
Closing & re-opening

NoahDragon:
@jd-carroll Got it, no worries.

stevenjoezhang:
See https://github.com/hexojs/hexo/pull/2680

PR #2680, opened 2017-08-01
Actually allow plugins to be stored as devDependencies

jd-carroll:
Plugins should be sourced from all available options, meaning both dependencies and devDependencies.  In the current implementation (#2558), plugins are only loaded from devDependencies *if there are no dependencies*.


jd-carroll:
Any chance this could be merged?  I view this more as a bug than an enhancement...

JLHwung:

Issue #848, opened 2014-09-19
Support multiple non-hierarchical categories


diaocow:

Alex1990:
支持多级目录，但是不支持一篇文章多个同级目录，比如：

```md
- HTML
- CSS
```

就会生成 `HTML`>`CSS` 结构的目录，HTML变成了CSS的父级目录，而我原先使用WordPress写的文章很多文章使用了两个目录，结果我从WordPress导入的文章目录列表结果很乱。不知道怎么才能修正好，迟迟没有迁移呢。

Xuanwo:
@Alex1990 这种需求，还是推荐使用Tags。

@diaocow 正如楼上所说，多级目录可以通过
```
- HTML
- CSS
```
实现。

leesei:
This is a feature request for multiple category.

Xuanwo:
@leesei the difference between `categories` and `tags` ?
I don't support the idea to add this featur, one way do one thing. 
How about your idea?

leesei:
@Xuanwo Categories can be both hierarchical and non-hierarchical. Tags are not hierarchical.

Xuanwo:
@leesei How about rename it as `feature request` ?

Xuanwo:
@leesei  I take some time to think this feature request.
Now I support to add `multilevel categories` whitout `same level categories`. And I think this way can give `categories` and `tags` different meanings  and work together well.

scorchio:
For people coming from WordPress this can cause a bit of a headache, since the WP migrator tries mapping back the multi-category selection present in WP to something, creating a completely new and unintentional category structure. An example:

In WP, there are three posts: one in the Personal category, one in the Travel category and the other one in both the Personal and Travel categories. (Those are same level categories.) As the result of the migration, you would end up with the following category structure:

- Personal
- Personal > Travel
- Travel

Note the difference between Personal > Travel and Travel; the post originally with both categories is not going to be available under Travel.

I'm not saying that any way (WP vs. Hexo) is better than the other, but this can be definitely confusing. If someone just wants to convert his existing site to Hexo like I do, one might need to rework all blog posts to clean up the category structure.

(I'm also aware that this might be the WordPress migrator's responsibility to handle.)

yannduran:
I just got caught by this myself. It took me a little while to figure out that I hadn't done anything wrong.

**How I expected it to work would be:**

- Category 1
- Category 2
- Category 3
  - Subcategory 3.1

**produces:**:

[...]

Issue #2492, opened 2017-04-06
Hexo 4.0 Roadmap


NoahDragon:
Here is a list for Hexo 4.0 roadmap:

- [ ] Convert old js code to ES6 standard.
- [ ] Speed is the key. 
- [ ] Expand the cli tool.
- [ ] Reduce the number of issues to below 100, which is a decent amount to manage by few people.
- [ ] Raise fund to cover existing fees (like domain name fee) and buy extra services (like NPM organization).
- [ ] Clean up left over tasks in old milestone

Please feel free to leave your thought in comments.

kirkcameron:
make sure, that hexo s and hexo g are in sync, before introducing any new features, please.

https://github.com/hexojs/hexo/issues/2503

mnlbox:
Please work on performance and multi core support on `hexo server` and `hexo generate`. I have major problem with big site generation in Hexo like this: #2579

NoahDragon:
@mnlbox Yes, it will be the major update in 4.0 build.

kirkcameron:
any news, regarding the release date, especially for https://github.com/hexojs/hexo/issues/2503 ?

NoahDragon:
@kirkcameron Not yet. The server calls the `render` function is as same as the generator, don't have clue why causes the differences.

huiwang:
@NoahDragon could you please share more details on the performance improvement?

NoahDragon:
@huiwang Thanks for asking. 

The generating performance issue is really hard to reproduce. It may cause by plugin, renderer, or Hexo self. We occasionally receive issue regarding performance, but only few shared their source to investigate. As far as I know, all of them are plugin and theme issue. 

I also tested performance on Hexo vanilla install, 1000 posts generated costs 3 mins. Still, have space to improve, but seems not a major issue.

tcrowe:
I noticed there's some confusion about how to do unit testing. This `mocha.opts` will help:
https://gist.github.com/tcrowe/b22a63daea19eaee69fe2bc37adee854

tcrowe:
[Prettier](https://prettier.io) can be included to force a uniform style across all hexo projects and it will integrate with [eslint-config-hexo](https://github.com/hexojs/eslint-config-hexo).

`npm install prettier eslint-plugin-prettier

`./.pretterrc`

```json
{
  "proseWrap": "never"
}
```

`./.eslintrc`

```json
"plugins": ["node", "prettier"],
```

[...]

PR #2734, opened 2017-08-30
848 hierarchical categories

danieljsummers:
This implements the ability to have multiple category hierarchies attached to a single post (#848).

By default, everything works the way it currently does; a list of categories will be assembled for a single category hierarchy. The change implemented here is that, if a category is a list, `setCategories()` will switch into multiple category mode, and each top-level item in the list will be a category hierarchy itself.

```yaml
categories:
- Example 1
- Example 2
- [Example 3, Subexample 3.1]
```
This will result in four different categories assigned to the post, with three different hierarchies (`example-1`, `example-2`, and `example-3/subexample-3.1`). The category generator already created index pages for intermediate categories, so `[url]/category/example-3/` is a good URL, and would contain this post.

I have just forked/cloned the `site` repo, and will get the documentation for this change there.  Additionally, I plan to look at the WordPress export format; if it exports the categories in their hierarchy, I'll see if I can PR a change to that migrator to bring the hierarchies over intact.

- [x] Passed the CI test.

### DIFF
--- a/lib/models/post.js
+++ b/lib/models/post.js
@@ -134,26 +134,26 @@ module.exports = function(ctx) {
 
   Post.method('setCategories', function(cats) {
     // Remove empty categories, preserving hierarchies
-    cats = cats.filter(cat =>
-      Array.isArray(cat) ? true : cat != null && cat !== ''
-    ).map(cat =>
-      Array.isArray(cat) ? removeEmptyTag(cat) : cat + ''
-    );
+    cats = cats.filter(function(cat) {
+      return Array.isArray(cat) ? true : cat != null && cat !== '';
+    }).map(function(cat) {
+      return Array.isArray(cat) ? removeEmptyTag(cat) : cat + '';
+    });
 
     var PostCategory = ctx.model('PostCategory');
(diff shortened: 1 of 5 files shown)


danieljsummers:
No fat arrow functions, huh?  :)  I'll have that pushed up momentarily.

NoahDragon:
Add reference https://github.com/hexojs/hexo/issues/848

danieljsummers:
I realized that I didn't check for duplicate parents not being added.  I wrote a test to check for that, and it passed with no code changes.  I'll commit the test to this branch.

danieljsummers:
FWIW, the code here enabled me to [move my tech blog](https://github.com/danieljsummers/techblog.djs-consulting.com), which is [live here](https://techblog.djs-consulting.com)!

NoahDragon:
@danieljsummers The live sample is really cool. This is also a major feature will in 4.0. https://github.com/hexojs/hexo/issues/2492

Thanks for the contribution!

NoahDragon:
Add reference: need to update hexo site when the feature released. https://github.com/hexojs/site/pull/550

PR #2736, opened 2017-09-02
Change plugins loading order

bolaum:
Intuition says local changes should have precedence over external ones, so this enforces the following load order:

external modules -> theme scripts -> local scripts

This way, any changes made locally will surely be applied.

The motivation was a long time trying to debug something I've overwritten in a local script that wasn't running.


NoahDragon:
Agree. LGTM.

PR #2760, opened 2017-09-19
fix a bug that post.permalink error when config.relative_link is true

tcatche:
When set `relative_link` to `true` in `config.yml`, the `post.permalink` will incrrect.

For example:
When `config.url` is `http://yoursite.com/` and the post's path is `2017/09/a-post` ,the result will `http://yoursite.com2017/09/a-post` . It misses a separator `/` after the site url.

This pull request aims to repair this problem.

- [ ]  Passed the CI test.


NoahDragon:
I will keep the PR unmerged, hope someone uses `relative_url` can give it a try.

JLHwung:

PR #2788, opened 2017-10-07
feat(config): use yml.load to support custom js types

JLHwung:
The `safeLoad` method is recommended for untrusted data. However, hexo uses `js-yaml` to parse config files handwritten by SAs. These config files should be treated as trusted data and thus we can use `load` method for flexibility. The `load` method uses `DEFAULT_FULL_SCHEMA` to support !!js/regexp or !!js/function types.

This commit does not introduce breaking change as `DEFAULT_FULL_SCHEMA` is a superset of `DEFAULT_SAFE_SCHEMA`, which is defaults of `safeLoad`.

- [x] Passed the CI test.

### DIFF
--- a/lib/hexo/multi_config_path.js
+++ b/lib/hexo/multi_config_path.js
@@ -47,7 +47,7 @@ module.exports = function(ctx) {
       var ext = pathFn.extname(paths[i]).toLowerCase();
 
       if (ext === '.yml') {
-        deepAssign(combinedConfig, yml.safeLoad(file));
+        deepAssign(combinedConfig, yml.load(file));
         count++;
       } else if (ext === '.json') {
         deepAssign(combinedConfig, yml.safeLoad(file, {json: true}));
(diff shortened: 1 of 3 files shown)


NoahDragon:
Just not confident about this change. I saw many issues related to the wrong format of config file. What happened if the yaml file is corrupt? The `load` will continue parsing the correct parts?

JLHwung:
@NoahDragon I don't think the corruption behavior of `safeLoad` differs to the one of `load`.

As from the [documentation](https://github.com/nodeca/js-yaml#load-string---options-) and the [code](https://github.com/nodeca/js-yaml/blob/c76b837cacc69de6b86a0781db31a9bb7a193875/lib/js-yaml/loader.js#L1579), the only difference between `safeLoad` and `load` is the schema support: The `load` has support of `!!js/regexp`, `!!js/undefined` and `!!js/function` type schema other than `safeLoad`. The `safeLoad` calls `load` under the hood but pass in restricted schema.

We can add corruption behavior test on `js-yaml` though.

NoahDragon:
Thank you. Then I think it is good for us.

PR #2790, opened 2017-10-09
Reuse img in asset img

zbicin:
This change reuses `img` plugin within the `asset_img`. Thanks to this, we can use more flexible syntax:

`{% asset_img [class names] slug [width] [height] [title text [alt text]]%}`

If Google Translate is right, this PR is related to #2275.


NoahDragon:
Thanks for the PR. BTW, the google translate is correct. 👍

JLHwung:

PR #2820, opened 2017-10-29
Rel external switched to noopener

ILAsoft:
Rel "external" switched to "noopener" for security best practices.
See https://developers.google.com/web/tools/lighthouse/audits/noopener (and other related documentation) for more info on this threat.

- [V] Add test cases for the changes.
- [V] Passed the CI test.

### DIFF
--- a/lib/plugins/filter/after_post_render/external_link.js
+++ b/lib/plugins/filter/after_post_render/external_link.js
@@ -30,7 +30,7 @@ function externalLinkFilter(data) {
 
     $(this)
       .attr('target', '_blank')
-      .attr('rel', 'external');
+      .attr('rel', 'noopener');
   });
 
   data.content = $.html();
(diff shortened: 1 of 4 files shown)


ILAsoft:
Good point, thank you, @JLHwung! 
The standard approach appears to be just noopener vs external noopener though -- should I still switch this PR to follow #2728's implementation or perhaps, vice versa, update and switch link_to to just be noopener as well?

ILAsoft:
Glitch with branches -- resubmitting, sorry for the extra "spam" due to the side-effects.

JLHwung:
@ILAsoft I think it make sense to replace `external` by `noopener`. Considering #2728, could you revert this PR and replace `external` by `noopener`.

@NoahDragon The `rel=external` is only used in HTML 4.01 STRICT DOCTYPE mode as an alternative of `target=_blank`, see http://www.css88.com/archives/112. I don't think any hexo theme will comply to HTML 4.01 only, so I think it's safe to remove `rel=external`.

ILAsoft:
@JLHwung Wasn't able to revert #2728 directly, so just did the next best thing by including the switch from external noopener to just noopener as part of this PR. Hope this works!

JLHwung:
@ILAsoft Great job. Now that we only have to wait for travis working again.

Issue #2123, opened 2016-08-30
after set post_asset_folder: true and root: /xxxx/，the image url was wrong


allenn:
## Environment Info

Node version(`node -v`)

```
v6.4.0
```

site `_config.yml`

```
# Hexo Configuration
···
# Hexo Configuration
## Docs: https://hexo.io/docs/configuration.html
## Source: https://github.com/hexojs/hexo/

# Site
title: Allen's Blog
subtitle:
description:
author: ""
email: ""
language: zh-Hans
timezone:

# URL
## If your site is put in a subdirectory, set url as 'http://yoursite.com/child' and root as '/child/'
url: http://pages.xxx.com/xxx/blogs
root: /xxx/blogs/
permalink: articles/:year-:month/:title/
permalink_defaults:

# Directory
source_dir: source
public_dir: public
tag_dir: tags
archive_dir: archives
category_dir: categories
code_dir: downloads/code
i18n_dir: :lang
skip_render:

# Writing
new_post_name: :title.md # File name of new posts
default_layout: post
titlecase: false # Transform title into titlecase
external_link: true # Open external links in new tab
filename_case: 0
render_drafts: false
post_asset_folder: true
relative_link: false
future: false
highlight:
  enable: true
  line_number: true
  auto_detect: false
  tab_replace: true

# Category & Tag
default_category: uncategorized
category_map:
tag_map:

# Date / Time format
## Hexo uses Moment.js to parse and display date
## You can customize the date format as defined in
## http://momentjs.com/docs/#/displaying/format/
date_format: YYYY-MM-DD
time_format: HH:mm:ss

# Pagination
## Set per_page to 0 to disable pagination
per_page: 10
pagination_dir: page

# Extensions
## Plugins: https://hexo.io/plugins/
## Themes: https://hexo.io/themes/
theme: landscape
#theme: yilia
# Deployment
## Docs: https://hexo.io/docs/deployment.html
```

Your theme `_config.yml`

```
# Header
menu:
  Home: /
  Archives: /archives
rss: /atom.xml

# Content
excerpt_link: Read More
fancybox: true

# Sidebar
sidebar: right
widgets:
- category
- tag
- tagcloud
- archive
- recent_posts

# display widgets at the bottom of index pages (pagination == 2)
index_widgets:
# - category
# - tagcloud
# - archive

# widget behavior
archive_type: 'monthly'
show_count: false

# Miscellaneous
google_analytics:
favicon: /favicon.png
twitter:
[...]

PR #2824, opened 2017-11-01
fix bug: remove dupliated root path when using asset_img tag

darkrasid:
solving #2123

- plugin img tag resolve root path twice 

on `lib/plugin/tag/asset_img.js`

```
  args[i] = url.resolve(ctx.config.root, asset.path);  << first 
```
and on `lib/plugin/tag/img.js`

```
  function makeUrl(path) {
    if (path[0] === '#' || path.substring(0, 2) === '//') {
      return path;
    }

    var data = url.parse(path);

    if (data && data.protocol) {
      return path;
    }

    path = config.root + path;

    return path.replace(/\/{2,}/g, '/'); << second
  }
```

I think there is no reason to prefix root config twice.

### DIFF
--- a/lib/plugins/tag/asset_img.js
+++ b/lib/plugins/tag/asset_img.js
@@ -27,7 +27,7 @@ module.exports = function(ctx) {
 
     if (!asset) return;
 
-    args[i] = url.resolve(ctx.config.root, asset.path);
+    args[i] = url.resolve('/', asset.path);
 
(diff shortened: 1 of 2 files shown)


darkrasid:
@JLHwung No problem. I add test case.

darkrasid:
There are lint issues, I'll check.

JLHwung:

PR #2837, opened 2017-11-09
Use eslint ci

JLHwung:
Use eslint-ci to improve CI build time

- [x] Passed the CI test.


JLHwung:
@NoahDragon The eslint takes 6 seconds in our build time, we run every eslint on node.js v6, v8.x and v9.x. `eslint-ci` will have it running only on one of v6.x, v8.x and v9.x, the total build time might or might not be shrinked since we still have to run eslint.

PR #2858, opened 2017-11-19
Fix a bug then pages display more reasonable.

Femonzor:
When I use `paginator` function like this:
`paginator({
    space: "",
    end_size: 0,
    mid_size: 2,
    prev_text: " ",
    next_text: " "
})`, then in `/page/2`, the page nums in `nav` tag whose class name is `page-nav` will be 2, 3, 4.
I think the corrent display should be 1, 2, 3, 4.

- [ ] Passed the CI test.


JLHwung:
After careful testing I can confirm the fix is working. Really great job!

Issue #2851, opened 2017-11-17
Hexo API incorrect report in use“Cannot read property 'f' of undefined”


YuYanDev:
I want to call the HEXO API in the Express.js framework as a hook,But it is wrong.

Perhaps due to the documentation is too simple and lead developers to understand the error caused the error.

I can not get any answer from Google. so,How should I call api?

This is my fail code and console output error:
``` javascript
var express = require('express');
var router = express.Router();
var Hexo = require('hexo');

router.get('/', function(req, res, next) {
  var hexo = new Hexo(process.cwd(), {});
  hexo.init().then(function(){
    hexo.watch().then(function(){
      hexo.call('generate').then(function(){
        return hexo.exit();
      }).catch(function(err){
        return hexo.exit(err);
      });
    });
  });
  res.send('hexo g ture');
});

module.exports = router;
```

and console error
``` 
FATAL Something's wrong.  Maybe you can find the solution here: http://hexo.io/docs/troubleshooting.html
TypeError: Cannot read property 'f' of undefined
    at Hexo.generateConsole (/Users/Hiroya/Project/express-blog-hexo/hexo/node_modules/hexo/lib/plugins/console/generate.js:16:20)
From previous event:
    at /Users/Hiroya/Project/express-blog-hexo/hexo/node_modules/hexo/lib/hexo/index.js:195:9
From previous event:
    at Hexo.call (/Users/Hiroya/Project/express-blog-hexo/hexo/node_modules/hexo/lib/hexo/index.js:191:10)
    at /Users/Hiroya/Project/express-blog-hexo/hexo/hexo.js:12:12
From previous event:
    at /Users/Hiroya/Project/express-blog-hexo/hexo/hexo.js:11:18
    at runCallback (timers.js:800:20)
    at tryOnImmediate (timers.js:762:5)
    at processImmediate [as _immediateCallback] (timers.js:733:5)
From previous event:
    at /Users/Hiroya/Project/express-blog-hexo/hexo/hexo.js:10:15
    at Layer.handle [as handle_request] (/Users/Hiroya/Project/express-blog-hexo/node_modules/express/lib/router/layer.js:95:5)
    at next (/Users/Hiroya/Project/express-blog-hexo/node_modules/express/lib/router/route.js:137:13)
    at Route.dispatch (/Users/Hiroya/Project/express-blog-hexo/node_modules/express/lib/router/route.js:112:3)
[...]

PR #2862, opened 2017-11-20
fix: Add defaults to console command generate

bhaskarmelkani:
Add defaults to console command `generate`.

Related to [#Issue](https://github.com/hexojs/hexo/issues/2851)

### DIFF
--- a/test/scripts/console/generate.js
+++ b/test/scripts/console/generate.js
@@ -20,7 +20,6 @@ describe('generate', () => {
   afterEach(() => fs.rmdir(hexo.base_dir));
 
   function testGenerate(options) {
-    options = options || {};
 
     return Promise.all([
       // Add some source files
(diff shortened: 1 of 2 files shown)


bhaskarmelkani:
Labelled bug as tests were not in sync with code.

test/scripts/console/generate.js
```diff
@@ -20,7 +20,6 @@ describe('generate', () => {
   afterEach(() => fs.rmdir(hexo.base_dir));
 
   function testGenerate(options) {
-    options = options || {};
```
JLHwung: `%s/generate(\{\})/generate()/g`.

bhaskarmelkani: Oh yeah, makes sense. (y)

JLHwung:

Issue #2145, opened 2016-09-06
How to generate highlightjs's style of codeblock


xcatliu:
Using the default config will generate code block like this:

``` html
<figure class="highlight plain"><table><tbody>
  <tr><td class="gutter"><pre><div class="line">1</div></pre></td><td class="code"><pre><div class="line">@<span class="keyword">import</span> <span class="string">'mobi.css/src/mobi'</span>;</div></pre></td></tr>
</tbody></table></figure>
```

Is there a way to generate code block like the official highlightjs's stlye?

``` html
<pre><code class="hljs scss">
  @<span class="hljs-keyword">import</span> <span class="hljs-string">'mobi.css/src/mobi'</span>;
</code></pre>
```

stackOverMind:
gutter can be removed by setting:
```yml
highlight:
  enable: true
  line_number: false
  auto_detect: false
  tab_replace:
```

xcatliu:
@stackOverMind Setting `line-number` to `false` will only disable lines, but the generated code is still `<figure><table>`, how to generate `<pre><code>` like http://highlightjs.org did? Then I can reuse all the styles that http://highlightjs.org provides.

leggetter:
@xcatliu The code for `highlight` is available via https://github.com/hexojs/hexo-util/blob/master/lib/highlight.js

So, you could take that and customise it to suit your needs. I think that's something we may do as we want to make the code highlighting functionality more of our own.

Maybe build your own tag that uses the functionality.

bensmithett:
I had this problem too. It seems like hexo was configured to use the old non-`hljs` style way back in 001b89d9bd7a6a13f0504d54c09278bf11cf594d to fix #434.

I think [this block](https://github.com/hexojs/hexo-util/blob/master/lib/highlight.js#L8-L10) is the problem - seems odd to kill highlight's default, good-for-css-architecture behaviour in hexo to fix a CSS problem in some hexo themes.

Happy to PR up a change reversing this if anyone involved thinks it's a good idea?

NoahDragon:
@bensmithett Agree. The `classPrefix: ''` should be configurable in _config.yml.

mathiasbynens:
+1 to somehow making it easy to customize `highlight` output.

In my case, I want something like:

```
<pre><code class="language-%language%">%code%</code></pre>
```

mathiasbynens:
Being able to define a custom util.highlight would be an elegant solution.

matthidinger:
[...]

PR #2901, opened 2017-12-11
Configs are now expanded for hexo-util/highlight to make it easier to…

alxbl:
This is part of the code necessary to address #2145.

It simply expands `config.highlight` inside of the `options` object that is passed to `hexo-util/highlight`.

I went this route to make sure that the plugin remains backward compatible for themes and people who have already styled based on the current highlight engine.

This is a quick proof of concept, and if everything is fine, I will update the documentation accordingly.

The other part of the PR will be in [`@hexojs/hexo-util`](https://github.com/hexojs/hexo-util)

The matching PR to support the added `highlight.hljs` setting is located at [`hexo-util/#30`](https://github.com/hexojs/hexo-util/pull/30)

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/filter/before_post_render/backtick_code_block.js
+++ b/lib/plugins/filter/before_post_render/backtick_code_block.js
@@ -18,7 +18,7 @@ function backtickCodeBlock(data) {
     var content = arguments[4];
 
     var options = {
-      ...config,
+      hljs: config.hljs,
       autoDetect: config.auto_detect,
       gutter: config.line_number,
       tab: config.tab_replace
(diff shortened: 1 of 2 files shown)


alxbl:
Ah, I just realized why `...config` wasn't being used... I hadn't noticed that node 6 support was required. In that case I can adjust the code to only extract the new setting into `options.`.

Let me know what course of action is best.

alxbl:
FYI: Don't approve the PR yet, there's a bug with the line numbers at the moment due to letting hljs render the lines on its own terms. I'm about to call it a night, but I should have time to fix it up tomorrow.

Issue #2952, opened 2018-01-08
Compatibility problem with hexo-disqus-proxy


ChrisXJH:
## Environment Info

Node version: v8.9.4

Hexo and Plugin version(`npm ls --depth 0`): 
├── hexo@3.4.4
├── hexo-deployer-git@0.3.1
├── hexo-disqus-proxy@1.3.4
├── hexo-generator-archive@0.1.5
├── hexo-generator-category@0.1.3
├── hexo-generator-feed@1.2.2
├── hexo-generator-index@0.2.1
├── hexo-generator-tag@0.2.0
├── hexo-migrator-rss@0.1.4
├── hexo-renderer-ejs@0.3.1
├── hexo-renderer-marked@0.3.0
├── hexo-renderer-stylus@0.3.3
├── hexo-server@0.2.2
└── hexo-wordcount@3.0.2

## For BUG

Bug description:
Wrong description meta tag content after installing hexo-disqus-proxy

Way to reproduce:
1. `npm install hexo-disqus-proxy --save`
2. Set description in _config.yml to some value
3. `hexo server`
4. Go to localhost and open element tag in browser developer tool. In head->meta description tag it should show:
`undefined                window.disqusProxy={           shortname: 'xxx',           username: 'xxx',           server: 'disqus-proxy.ycwalker.com',           port: 443,           adm`

PR #2953, opened 2018-01-08
Fixed hexo-disqus-proxy plugin compatibility problem

ChrisXJH:
Fixed issue #2952
`config.description` should have higher priority than `content`


stevenjoezhang:
You should fix the hexo-disqus-proxy plugin, not hexo...

PR #2968, opened 2018-01-16
Delete duplicate eslint rule not overwritten

segayuu:
Most of the rules done in the previous PR have been set to [eslint-config-hexo@2](https://github.com/hexojs/eslint-config-hexo).
Both of these are based on [jscs-preset-hexo](https://github.com/hexojs/jscs-preset-hexo), so there are many duplicate settings.
Therefore, it is not intentional that settings not overwritten from eslint-config-hexo will remain.
(That is, the rest are adjustments for the purpose of avoiding errors.)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -2,116 +2,16 @@
   "extends": "hexo",
   "root": true,
   "rules": {
-    "indent": [
-      2,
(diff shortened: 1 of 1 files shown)


JLHwung:

PR #2982, opened 2018-01-20
robust for url compuation of 'view raw'

richardo2016:

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/tag/include_code.js
+++ b/lib/plugins/tag/include_code.js
@@ -64,7 +64,7 @@ module.exports = function(ctx) {
       // If the language is not defined, use file extension instead
       lang = lang || pathFn.extname(path).substring(1);
 
-      caption = '<span>' + title + '</span><a href="' + ctx.config.root + codeDir + path + '">view raw</a>';
+      caption = '<span>' + title + '</span><a href="' + pathFn.join(ctx.config.root, codeDir, path) + '">view raw</a>';
 
       return highlight(code, {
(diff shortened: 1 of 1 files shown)


lib/plugins/tag/include_code.js
```diff
@@ -64,7 +64,7 @@ module.exports = function(ctx) {
       // If the language is not defined, use file extension instead
       lang = lang || pathFn.extname(path).substring(1);
 
-      caption = '<span>' + title + '</span><a href="' + ctx.config.root + codeDir + path + '">view raw</a>';
+      caption = '<span>' + title + '</span><a href="' + pathFn.join(ctx.config.root, codeDir, path) + '">view raw</a>';
```
JLHwung: The `join` will generate `\\` on Windows platform, but `\\` is not a valid URI hierarchy separator, see [RFC 1738 Section 2.3](https://www.ietf.org/rfc/rfc1738.txt).

tcrowe:
For URL join I usually use this: https://github.com/jfromaniello/url-join

I'm not recommending that we add it as a dependency but just recognize that to join URLs can be complex.

lib/plugins/tag/include_code.js
stevenjoezhang:
```
pathFn.posix.join
```

stevenjoezhang:
Superseded by https://github.com/hexojs/hexo/pull/4996

PR #2998, opened 2018-01-29
Add eslint-plugin-node

segayuu:
[eslint-plugin-node](https://www.npmjs.com/package/eslint-plugin-node) adds a rule specific to node.js.
This plugin detects whether inappropriate syntax / API is used for the version of package node.js.
Actually, the current Hexo does not work in Node.js: v4, and the use of deprecated API is confirmed in Node.js: v6 or v8!
In order to be able to check by automatic test, I will send this PR.
The aim of this PR is to explicitly set the supported Node.js version!

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,6 +1,7 @@
 {
   "extends": "hexo",
   "root": true,
+  "plugins": [ "node" ],
   "rules": {
     "one-var": 0,
     "operator-linebreak": [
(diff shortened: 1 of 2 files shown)


JLHwung:
I recommend to migrate the work to `eslint-config-hexo` as under most circumstances we will use node to run hexo.

segayuu:
moved to hexojs/eslint-config-hexo#3

PR #3007, opened 2018-02-03
Explicitly set the supported version of nodejs. 

segayuu:
Its value is the lowest version logically tested on travis ci.

- [ ] Passed the CI test.


segayuu:
All errors that occurred in travis ci are due to too many monitored objects. It is not an error caused by changing the code itself.

PR #3028, opened 2018-02-17
feat: Improved date formatting.

xeres:
The monthly format 'MMMM YYYY' is not work well in some languages (ex. Japanese).
It was possible as to customizable by config.

- [ ] Passed the CI test.

### DIFF
--- a/lib/plugins/helper/list_archives.js
+++ b/lib/plugins/helper/list_archives.js
@@ -19,7 +19,9 @@ function listArchivesHelper(options) {
   var self = this;
 
   if (!format) {
-    format = type === 'monthly' ? 'MMMM YYYY' : 'YYYY';
+    format = (type === 'monthly')
+      ? (config.month_format ? config.month_format : 'MMMM YYYY')
+      : (config.year_format ? config.year_format : 'YYYY');
   }
 
   var posts = this.site.posts.sort('date', order);


tcrowe:
Thank you for the idea @xeres. Where does `monthly` or `config.type` come from?

xeres:
What you are saying is `options.type`?
It's an archive page type, I think that it is not the essence.

NoahDragon:
We alread have `date_formate`, do we still need new configurations? https://hexo.io/docs/configuration.html#Date-Time-format

xeres:
I think that it is necessary, because we have **only** date and time formats.
We don't have year and monthly formats.

My commit improves anchor texts that linked monthly archive pages.
It's hard-coded as 'MMMM YYYY', but it does not work well in some languages (ex. Japanese).

Does that make sense?

NoahDragon:
@xeres I don't know Japanese, so I couldn't imagine why it is not supported. Could you please provide a test case (in Japanese) which the current date_format option can't work?

xeres:
ex.) 2018-03-16, 'MMMM YYYY'

**In English:**
Expected/Actually: 'March 2018'

**In Japanese:**
Correct Japanse notation: '2018年3月' (2018-Nen, 3-Gatsu; Nen, Gatsu is suffix)
Actualy: '3月 2018'

1) MMMM is displayed in locale, but YYYY is not. And there is no one instead.
2) The 1st word is MMMM, 2nd word is YYYY. This order does not work in some languages.

I want to configure monthly formats, but it is hard-coded in `hexojs`.
And, `date_format` is not monthly format, rights?

So, I would like to propose `monthly_format`.

Make sense?

NoahDragon:
@xeres Thanks for the elaboration. I see the problem. Let me try if there is a way to work around. :smile:

JLHwung:
I suggest we put default value of `date_format`, `month_format` to i18n and align with `moment` [locale settings](https://github.com/moment/moment/blob/develop/locale/ja.js#L21) accordingly. We should not suppose our user to be pretty familiar with `moment` and are likely live with improper month format unless we set default value accordingly.

And we should have further unit test for coverage. It should be pretty straightforward, though.

stevenjoezhang:
Any updates on this thread?

Issue #1125, opened 2015-03-20
Default language


dcardosods:
The theme I'm using has 3 files in the languages folder , `default`, `fr-FR` and `zh-Hans`.
I need English (default) as my site language. Leaving the `language` option in `_config.yml` blank doesn't fallback to default (as I suppose is the expected behavior). iissnan/hexo-theme-next#57 has more details.
In the line [lib/theme/index.js#L26](https://github.com/hexojs/hexo/blob/master/lib/theme/index.js#L26) it isn't checked if `languages` is `null` before adding it to the array. Maybe the bug is here.

kfranqueiro:
This happened to me as well, both with the default `landscape` theme as well as the `light` theme.  It used the alphabetically-last language in the theme's `languages` folder (zh-TW in this case).  I had to set `language` to `'default'` explicitly to actually make it use "default".

It should be simple to reproduce.  Run `hexo init` somewhere, run `hexo server`, open your browser to localhost:4000, and boom, you have Chinese text in the sidebar.

leesei:
Agreed, I read in some comment that `language`'s default is `en`.
This means the theme are incorrect in using `default` as language name.
Maybe someone can help verify that?

mnn:
In my case using default `landscape` theme with `en` lead to **random** language being picked up. Changing language to `default` fixed all issues.

mnn:
It turns out that it is (and probably was) still behaving "randomly", after adding a post I'm getting index page in second language even though I have in confg:

```
language:
- en
- cs
```
I'm getting `cs` translation on main page (in theme I have copied a `default.yml` file to `en.yml` just ot be sure). This bug is quite annoying :disappointed:.

jbuddha:
same issue.... random language everytime... so frustrating.

weqeo:
same...i need chinese,but it always de.......

NoahDragon:
@weqeo set language in hexo _config.yml to zh-cn may solve your issue.

weqeo:
@NoahDragon thanks,i found that this bug only appearanced when i  test in localhost,it's just show everything fine when i deploy it to my domain.

SonicDMG:
This issue still seems to exist.  I am also using a variation of the landscape template and it is totally random on my end either for localhost or when I deploy.  I have tried setting "language:" in my _config.yml file to "en", "default", added it to my posts directly, etc... it doesn't seem to care what I do, goes random.

frumpel:
[...]

PR #3069, opened 2018-03-11
Use default language when no language set

bobd91:
Use default language when no language set, or there is no translation for specified language

If no language is set in _config, or the theme has no language file for the one set, then the translation data is chosen in indeterminate manner as it is taken from a list of all available translations returned by hexo.theme.i18n#list().

The changes to lib/plugins/filter/template_locals/i18n.js use i18n.languages in preference to config.language and i18n.list() as this provides an array of configured languages with 'default' on the end.  This ensures that if no language is set or the language cannot be found then then default translation will be used.

The changes to test/scripts/filters/i18n_locals.js set i18n.languages rather than config.language as the original test code did not set the 'default' language correctly.

Added two tests
+ use config by default - with no languages, default language should be used
+ use config by default - with unknown language, default language should be used

Both of the new tests fail on the current release but pass with the changes in the pull request.

This is a fix for #1125

### DIFF
--- a/lib/plugins/filter/template_locals/i18n.js
+++ b/lib/plugins/filter/template_locals/i18n.js
@@ -21,7 +21,8 @@ function i18nLocalsFilter(locals) {
       lang = data.lang;
       page.canonical_path = data.path;
     } else {
-      lang = getFirstLanguage(i18nConfigLanguages);
+      // i18n.languages is always an array with at least one argument ('default')
+      lang = i18nConfigLanguages[0];
     }
 
     page.lang = lang;
@@ -36,11 +37,3 @@ function i18nLocalsFilter(locals) {
 }
 
 module.exports = i18nLocalsFilter;
-
-function getFirstLanguage(lang) {
-  if (Array.isArray(lang)) {
-    return lang[0];
-  }
-
-  return lang;
-}
--- a/lib/plugins/filter/template_locals/i18n.js
+++ b/lib/plugins/filter/template_locals/i18n.js
@@ -11,6 +11,7 @@ function i18nLocalsFilter(locals) {
   const page = locals.page;
   let lang = page.lang || page.language;
   const i18nLanguages = i18n.list();
+  const i18nConfigLanguages = i18n.languages;
 
   if (!lang) {
     const pattern = new Pattern(`${i18nDir}/*path`);
@@ -20,7 +21,7 @@ function i18nLocalsFilter(locals) {
       lang = data.lang;
       page.canonical_path = data.path;
     } else {
-      lang = getFirstLanguage(config.language);
+      lang = getFirstLanguage(i18nConfigLanguages);
     }
 
     page.lang = lang;
@@ -28,7 +29,7 @@ function i18nLocalsFilter(locals) {
 
   page.canonical_path = page.canonical_path || locals.path;
 
-  const languages = _([].concat(lang, i18nLanguages)).compact().uniq().value();
+  const languages = _([].concat(lang, i18nConfigLanguages)).compact().uniq().value();
 
   locals.__ = i18n.__(languages);
   locals._p = i18n._p(languages);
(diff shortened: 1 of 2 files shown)


bobd91:
LOL, if you remove code then the coverage % goes down even though all of the code you have changed is covered!
I only removed the code in the first place as my original changes caused a particular branch to never be executed (code run if variable is not an array on something that is now always an array).
Feeling frustrated with coveralls.

JLHwung:
Good catch and thanks!

ivan-nginx:
> If no language is set in _config, or the theme has no language file for the one set, then the translation data is chosen in **indeterminate manner** as it is taken from a list of all available translations returned by hexo.theme.i18n#list().

Not indeterminate manner, but [take first catched language](https://github.com/hexojs/hexo/pull/3069/files#diff-22a1deaf9f8d06dd89a1cf0dfd7ecad4R13) in `themes/languages` directory. So, if first language will `de.yml`, it's will set DE language if this 2 cases above will reproduce.

For now this pull create [1 additional bug in NexT theme](https://github.com/theme-next/hexo-theme-next/issues/209): if 2 cases above will reproduce and `default.yml` not exists in `themes/languages` directory, no one language will be selected. So, for now it's mean need to create 1 additional symlink or rename `en.yml` on `default.yml`, but before this pull this bug was solved just by renaming `en.yml` on `_en.yml`.

## Breaking Changes for most Hexo themes

This pull with 2 cases above will brake many themes in which `default.yml` in `themes/languages` not exists. There is the list:

* [NexT](https://github.com/theme-next/hexo-theme-next/tree/master/languages) (~~[old NexT](https://github.com/iissnan/hexo-theme-next/tree/master/languages)~~ exists duplicate `default.yml` file of `en.yml`)
* [Tranquilpeak](https://github.com/LouisBarranqueiro/hexo-theme-tranquilpeak/tree/master/languages)
* ~~[Yilia](https://github.com/litten/hexo-theme-yilia/tree/master/languages)~~ (`en.yml` renamed on `default.yml` by default)
* [Indigo](https://github.com/yscoder/hexo-theme-indigo/tree/card/languages)
* [Material](https://github.com/viosey/hexo-theme-material/tree/master/languages)
* [Icarus](https://github.com/ppoffice/hexo-theme-icarus/tree/master/languages)
* [Apollo](https://github.com/pinggod/hexo-theme-apollo/tree/master/languages)
* [Hueman](https://github.com/ppoffice/hexo-theme-hueman/tree/master/languages)
  **And many other themes which authors don't know this bug cases, looks like this:**
  

So, this bugfix now **produce ≈ 80% breaking changes for all Hexo themes**.
@JLHwung, @NoahDragon, please, test pulls like that before merge for all possible breaking changes, Thank's!

### Possible fixes for this changes

> This ensures that **if no language is set** or the **language cannot be found** then **default translation** will be used.

[Use first language in list](https://github.com/hexojs/hexo/pull/3069/files#diff-22a1deaf9f8d06dd89a1cf0dfd7ecad4L31) (instead of [this](https://github.com/hexojs/hexo/pull/3069/files#diff-22a1deaf9f8d06dd89a1cf0dfd7ecad4R33)) if `default.yml` file not exists.

bobd91:
> Not indeterminate manner, but take first catched language in themes/languages directory. 

On your system maybe, but the order of the language files is dependent upon the order that Node's readdir function provides.  It is [well documented](https://github.com/nodejs/node/issues/3232) that Node's readdir on Windows does not provide a sorted list and even on Linux the underlying readdir C function [makes no promises on ordering](https://stackoverflow.com/questions/8977441/does-readdir-guarantee-an-order).

ivan-nginx:
> Node's readdir on Windows does not provide a sorted list and even on Linux the underlying readdir C function makes no promises on ordering.

It's alphabetical or vice-versa. Will be load first or last language. But it's can't load random language, i think. So, it can't be called *indeterminate manner*.

Issue #2436, opened 2017-03-08
unable to exclude specific folders or files when run 'hexo clean'


ZhaoC:
> before you submit your issue, please delete all the example code in template

## For feature request

the `public_dir` folder will be delete and re-generated after run 'hexo clean' currently. Is there a way to exclude specific files or folders when do 'hexo clean' to flush caches?  In my case, I specific the `public_dir` to a git dir, and  the `.git` folder will be delete if I run 'hexo clean'. Hope there is a way to handle it properly.  

## Environment Info

Node version(`node -v`)
```
v6.10.0
```

Plugin version(`npm ls --depth 0`)
```
├── hexo@3.2.2
├── hexo-algoliasearch@0.2.1
├── hexo-generator-archive@0.1.4
├── hexo-generator-category@0.1.3
├── hexo-generator-feed@1.2.0
├── hexo-generator-index@0.2.0
├── hexo-generator-sitemap@1.1.2
├── hexo-generator-tag@0.2.0
├── hexo-renderer-ejs@0.2.0
├── hexo-renderer-marked@0.2.11
├── hexo-renderer-stylus@0.3.1
└── hexo-server@0.2.0
```

NoahDragon:
Currently, in my opinion, there is no way to do so.

ZhaoC:
@NoahDragon  thanks for the response. I think it would be a `good to have` feature for hexo.  I found a walk around which can avoid deleting all the content of workspace folder once the db.json file is mess up : 

Before run 'hexo clean', change the 'public_dir: my_working_dir' under '_config.yml' to 'public_dir: temp_dir' , then execute 'hexo clean'. After that, change the dir back to 'my_working_dir'. If you run 'hexo generate' now, the data should be fine. Hope that will be useful to someone who encounters the same issue as me.

ertrzyiks:
I would love to have this possibility too. Maybe it could be a new filter? `after_clean` for example, so you could create a simple script to clean up other things with `hexo clean` command. I will try to prepare a PR for that.

stevenjoezhang:
The `.git` directory in `public_dir` is undefined behavior for Hexo, please use `hexo-deployer-git`.

Closing due to #3091

PR #3091, opened 2018-03-24
Add after_clean filter

ertrzyiks:
Related to #2436 

It allows injecting code to the `hexo clean` command to
let plugin developers perform additional clean up.

- [x] Passed the CI test.


JLHwung:
Thank you so much for your contribution. Hexo can't be rock without contributors like you.

Issue #3108, opened 2018-04-04
Breaking Changes for most Hexo themes in languages


ivan-nginx:
## For BUG

https://github.com/theme-next/hexo-theme-next/issues/209

## For feature request

#3069 (merged)

## Explanation & Suggestion

https://github.com/hexojs/hexo/pull/3069#issuecomment-378536168

bobd91:
This is a difficult problem as the documented behaviour (use 'default' language file if no language specified) did not work so you had to come up with some scheme that did work.

Relying on the behaviour of a bug is always risky, and the longer the bug takes to get fixed the more people rely on the buggy behaviour and the more pain fixing the bug causes.

However, this should not be classified as a breaking change.  The behaviour that you and others relied upon was not a documented feature but the side-effect of a defect that was reported and scheduled to be fixed.

ivan-nginx:
I understand this, but we have for now badly situation then it was before:

1. **Before fix**: new Hexo / Theme user just install Hexo & Theme and with default config settings see first language in list (at most cases it's `en.yml`). Anyway, he see language without big bugs, before he will define needed by him language.
2. **After fix**: with same situation, user don't see any language and just see untranslated values like `post.posted`, `footer.theme`, etc. and will think what this bug of Theme or, at least, of Hexo.

This is big difference from point of view new Hexo users and create some troubles for Theme mainteiners under Hexo (from this pull need at least to create additional symlink or duplicate file to around this bugfix).

ivan-nginx:
I think u see my opinion and i appreciate this. All maintainers of themes under the Hexo will appreciate this, i think. And, yes, i already import fix on this fix, but as for me, always need to have any competidors in similar things, to be better for all of us.

PR #3110, opened 2018-04-04
Use old language file selection when no default file exists

bobd91:
Issue #3108 points out that many existing themes rely on the old but incorrect way of selecting language files prior to the pull request #3069 which fixes the incorrect behaviour.

This PR makes a small change so that when no language is configured and no default language file exists then the fallback is to use the old method of selecting a language file rather than not selecting any file.

A single test has been added to confirm that the first language file added will be used as the fallback.


ivan-nginx:
Thank's. Looks better. :+1:

JLHwung:
Will land in next patch release. Thanks.

PR #3118, opened 2018-04-15
updated config loader to accept absolute paths to config files

mgoldsborough:
Update to support absolute paths to config files. 

Commit checklist:
- [x] Passed the CI test.

### DIFF
--- a/lib/hexo/multi_config_path.js
+++ b/lib/hexo/multi_config_path.js
@@ -22,23 +22,24 @@ module.exports = ctx => function multiConfigPath(base, configPaths, outputDir) {
 
   } else {
     // only one config
+    let configPath = defaultPath;
 
     // check for absolute config file
     if (pathFn.isAbsolute(configPaths)) {
       if (!fs.existsSync(configPaths)) {
         log.w(`Config file ${configPaths} not found, using default.`);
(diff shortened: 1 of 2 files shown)


lib/hexo/multi_config_path.js
```diff
@@ -22,12 +22,24 @@ module.exports = ctx => function multiConfigPath(base, configPaths) {
 
   } else {
     // only one config
+    let configPath = defaultPath;
+
+    // check for absolute config file
+    if (pathFn.isAbsolute(configPaths)) {
```
JLHwung: It can be simplified to
```
configPath = pathFn.isAbsolute(configPaths) ? configPaths ? pathFn.resolve(base, configPaths);
```
then we can check `fs.existsSync` in a single branch.

mgoldsborough: Nice add. Looks good to me.

mgoldsborough: @JLHwung do I need to implement these changes on my branch? Or can you do it as a collaborator directly via GitHub?

mgoldsborough:
@JLHwung I implemented your requested changes. Let me know if there is anything else.

PR #3122, opened 2018-04-16
added command line option `output` to write db.json and _multiconfig.…

mgoldsborough:
I'm working with a read-only file system and I need a way to direct the output to a different path.

I added command line option `output` to write db.json and _multiconfig.yml to a path provided

- [ ] Passed the CI test.

### DIFF
--- a/test/scripts/hexo/multi_config_path.js
+++ b/test/scripts/hexo/multi_config_path.js
@@ -14,6 +14,22 @@ describe('config flag handling', () => {
 
   function ConsoleReader() {
     this.reader = [];
+    this.d = function(...args) {
+      var type = 'debug';
+      var message = '';
+      for (var i = 0; i < args.length;) {
+        message += args[i];
+        if (++i < args.length) {
+          message += ' ';
+        }
+      }
+
+      this.reader.push({
+        type,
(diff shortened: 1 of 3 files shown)


test/scripts/hexo/multi_config_path.js
```diff
@@ -14,6 +15,22 @@ describe('config flag handling', () => {
 
   function ConsoleReader() {
     this.reader = [];
+    this.d = function(...args) {
+      var type = 'debug';
```
JLHwung: If the differences of `this.d` and `this.i` are only `type`, could you refactor into a factory so that
```
this.d = readerStub("debug")
this.i = readerStub("info")
```

Thanks.

mgoldsborough: I was just copying the style of the existing code. I didn't want to touch to many lines. It's also a test case, not production code. If you want this refactored, change impact `this.w` and `this.e` which are below these changes in the same file.

If you'd like this refactored, let me know. Otherwise I'll leave it as it.

mgoldsborough: @JLHwung bump.

PR #3124, opened 2018-04-18
Replace String#substring() with String#startsWith() or String#endsWith()

segayuu:
And String#indexOf() with String#includes()

- [ ] Passed the CI test.

### DIFF
--- a/lib/plugins/filter/meta_generator.js
+++ b/lib/plugins/filter/meta_generator.js
@@ -1,0 +1,17 @@
+'use strict';
+
+let cheerio;
+const hexoGeneratorTag = '<meta name="generator" content="Hexo %s" />';
+
+function hexoMetaGeneratorInject(data) {
+  if (!cheerio) cheerio = require('cheerio');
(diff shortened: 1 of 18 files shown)


segayuu:
Since I have not responded for more than a week since I requested a review, I will notify @hexojs/core member once more.
If there is no response after one week from this comment, we will merge.

ivan-nginx:
@segayuu > If there is no response after one week

Why week? Go about 1-2 month? It's normal period for awaiting something important.

Also, i see codeclimate issues, mb u can fix them?

lib/plugins/filter/meta_generator.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
```
ivan-nginx: Also, it seems new filter added here. But in pull readme there is nothing about this.

segayuu: That change is the result of # 3129 being resolved by running git rebase.

PR #3129, opened 2018-04-20
Misc: add meta generator injection filter

h404bi:

- [x] Passed the CI test.

This implements the `meta generator injection` feature. For more information, see Miscellaneous section of [Hexo 4.0 Roadmap](https://github.com/hexojs/hexo/issues/2492), and https://github.com/hexojs/site/pull/649.

### DIFF
--- a/lib/plugins/filter/meta_generator.js
+++ b/lib/plugins/filter/meta_generator.js
@@ -1,14 +1,13 @@
 'use strict';
 
-const rMetaGenerator = /<meta\s+name=['|"]?generator['|"]?/i;
-const hexoGeneratorTag = '<meta name="generator" content="Hexo %s" />';
 let cheerio;
+const hexoGeneratorTag = '<meta name="generator" content="Hexo %s" />';
 
 function hexoMetaGeneratorInject(data) {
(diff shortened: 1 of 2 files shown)


lib/plugins/filter/meta_generator.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+const rMetaGenerator = /<meta\s+name=['|"]?generator['|"]?/i;
```
segayuu: Current code will recognize incorrect expressions as html.

``` html
<meta name="generator'>
```

Also, "|" is described in the character set, this expression will detect `<meta name=|generato |>`. Is this just as intended?

segayuu: I prepared regExp which seems to be able to express the intention of writing regExp.
`/meta\s+name(["']?)generator\1/i`

h404bi: @segayuu Oops, didn't notice that, I took that regExp from [hugo](https://github.com/gohugoio/hugo/blob/master/transform/hugogeneratorinject.go#L24). I'll make a tweak. I'm considering to use cheerio to detect this meta tag.

PR #3235, opened 2018-08-18
Use date instead of file mtime for updated date

tomap:
Because this date is unreliable when cloning a repository from git
matches the last clone instead of lat update
This is particularly important when used in cunjunction with hexo-generator-feed

- [X] Add test cases for the changes.
- [X] Passed the CI test.


tcrowe:
Hi @tomap. It seems like a good idea. Where does the `updated` variable show up in the site?

tomap:
My use case is in feed plugin where the update date of a post makes no sense in my case, where I would want them to not use the file dates, but instead use the date, when no update is set.

However, after giving it some thought, I think it is more an option of feed to introduce or a new post property, otherwise, it will create a change of behavior

tcrowe:
@tomap Should we ask hexo core team their thoughts and see if anyone wants to talk more about this? I cannot yet understand what depends on these dates but it seems like a good idea you had.

tomap:
Yes, we should, especially given the breaking change possibility

NoahDragon:
IMHO, this will change the purpose of the updated date field. I know some plugins use this field to set the updated date for the posts. Why not we modify the hexo-generator-feed to handle the date correctly, instead of changing the hexo itself?

tomap:
Agreed. I'll do a PR there

tomap:
Hi, I thougth a bit more about it and it is important to have this change inside hexo. 
Here is why:
1/ if I do it at the plugin level, I'll have two variables available:
* Date (date of publication)
* Updated, which will be either the "true" updated date (from the front-matter), and a meaningless date (mtime) but with no way for me to kwno which is which

2/ If I do it at hexo level (with an option, for backward compatibility)
* Date (date of publication)
* Updated, which will be either the "true" updated date (from the front-matter), or the date of publication (if no updated date is available)

So I'm reopening this PR, and will amend it soon to include the new option for backward compatility (backward compatible behavior if false or unset)

SukkaW:
Using file mtime as `updated` affect not only `hexo-generator-feed` but also any theme using variable `post/page.updated`.

For example, my `hexo-theme-suka` supports a feature which based on `post/page.updated`:

image

In my case, I believe it is a b better option to use `date` as `updated` when no `updated` is given in front-matter.

curbengh:
I prefer not dropping `mtime`, it's quite useful, although I personally prefer to manually set the `post.updated`. I'm ok with introducing new option (as shown in your latest commit.

Would `updated:` (in similar format as `date:`) in front matter resolve the issue you mentioned? Kinda similar approach as https://github.com/hexojs/hexo/pull/3612.

tomap:
@curbengh indeed, mentionning updated in my posts would solve the situation, but I don't want to have to mention them to fix the issue.
I want the default to work in my case
That is why I'm introducing a new option

curbengh:
These two seem to be conflicting:
> Updated, which will be either the "true" updated date (from the front-matter), or the date of publication (if no updated date is available)
`// Use post date for updated date instead of file modification date when front-matter provides no date`

should it be "...when front-matter provides no updated date"?

---

On the other note, it seems `updated:` can already be user-specified in front-matter. I didn't realize this before.

tomap:
> These two seem to be conflicting:
> 
> > Updated, which will be either the "true" updated date (from the front-matter), or the date of publication (if no updated date is available)
> > `// Use post date for updated date instead of file modification date when front-matter provides no date`
> 
> shouldn't it be "...when front-matter provides no updated date"?

I fixed that. Thank you

curbengh:
Thanks. I'll test it in the meantime.

curbengh:
Somehow lines 49-55 of "lib/plugins/processor/asset.js" only affects page, not post.
I put 
``` js
        console.log('2data.date =  ' + data.date)
        console.log('2data.updated: ' + data.updated)
        console.log('2data.source: ' + data.source)
```

 in those lines, no output if there isn't any page.

curbengh:
Note for theme dev, If a theme currently display updated date, it needs to add another condition:

```
<% if (post.updated && post.updated.toString() !== post.date.toString()) { %>
...
<% } %>
```

tomap:
Or we could introduce an easier property that would leave the updated property empty in case it is not set in front matter?

curbengh:
The condition I mentioned is for compatibility when `use_date_for_updated` is enabled. Now I think about it, the following condition works better,

```
<% if (post.updated && !config.use_date_for_updated) { %>
...
<% } %>
```

curbengh:
> Or we could introduce an easier property that would leave the updated property empty in case it is not set in front matter?

I would prefer that option, so `if (post.updated)` works better.

`mtime_updated` perhaps?

curbengh:
@tomap 

> Or we could introduce an easier property that would leave the updated property empty in case it is not set in front matter?

Do you want to implement that option in this PR or later? Otherwise, this is good to go.

tomap:
I'm not going to change this PR.

PR #3276, opened 2018-10-01
Drop node.js v9 and support v10 on AppVeyor

YoshinoriN:
Node.js v9 is already EOL.
Drop v9 and support v10 on AppVeyor.

- [x] Passed the CI test.


YoshinoriN:
@tomap @segayuu 
Thanks !!

PR #3277, opened 2018-10-01
Update devDependencies

YoshinoriN:
Some package are out of date.
This PR is update devDependencies and upgraded husky's script, because it [has breaking change in v1.0.0](https://github.com/typicode/husky#upgrading-from-014)

- [x] Passed the CI test.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,6 @@
     "eslint": "eslint-ci .",
     "test": "mocha test/index.js",
     "test-cov": "istanbul cover --print both _mocha -- test/index.js",
-    "precommit": "lint-staged",
     "lint-staged": "lint-staged"
   },
   "directories": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -62,20 +61,25 @@
...
-    "mocha": "^5.0.5",
-    "rewire": "^3.0.2",
-    "sinon": "^4.1.2"
+    "hexo-renderer-marked": "^0.3.2",
+    "husky": "^1.0.1",
+    "istanbul": "^0.4.5",
+    "lint-staged": "^7.3.0",
+    "mocha": "^5.2.0",
+    "rewire": "^4.0.1",
+    "sinon": "^6.3.4"
```
segayuu: IMO think the version difference is very large even for the sinon library alone, and it is not appropriate for the update target at once.

YoshinoriN: Thanks!

At first about sinon.As you know This PR update sinon 4.x to 6.x.
sinon breaking changes in 5.x and 6.x is only removed `spt.reset()`. But this repositories doesn't using `spy.reset()`.

Please see sinon migrate guide.

* [4.x to 5.x](https://sinonjs.org/guides/migrating-to-5.0)
* [5.x to 6.x](https://sinonjs.org/guides/migrating-to-6.0)

Second about rewire. rewire 4.0.0 has two breaking changes.
One is drop nodejs v4. This hexo repository already set node-engines higher 6.9.0, so we don't worry this breaking change.
Another one is potentially breaking, sorry I can not be say absolutely no problem, but unit test are passed. I think may be no problem.

* [rewire 4.0.0](https://github.com/jhnns/rewire/releases/tag/v4.0.0)

So, I think we can bump up these dependencies to higher major version. But, if hexojs's committer require more carefully, I change them major version up to minor version up.

segayuu: Thanks @YoshinoriN for your quick reply!
I think there is no problem with the change itself.
Since the version difference is simply very large, I was worried about whether it is appropriate for PR unit.
It was a pleasure for you to leave the content of breaking change as a comment!

Including the reply to the review, I thought that I could merge. However, I do not know whether it is correct as a guideline for the hexojs project.

YoshinoriN:
OK. I wait other hexo core member's review.
I will try to devided this PR into some PRs if this PR not accepte.

NoahDragon:
I agree with @segayuu, but since the upgrades packages are all in dev. IMHO, I feel it is safe to merge as it passes the CI test.

YoshinoriN:
@segayuu @NoahDragon 
Thanks.
OK, I will close this PR and I divide this PR into some PRs.

PR #3289, opened 2018-10-15
Use Reflect.apply()

segayuu:
`Function#call()` and `Function#apply()` can express assertion that it converts `this` value more clearly by converting it to `Reflect.apply()`.

- [ ] Passed the CI test.


YoshinoriN:
It seems good. But, I can't judge clearly which is better.
I want other member's review.

NoahDragon:
LGTM, we eventually need to convert the old js code to es6, this change makes the target closer. :+1:

PR #3291, opened 2018-10-15
Spread function: Hexo built-in plugin blockquote

segayuu:
The blockquote built-in plugin is too large as a single function.
Also, most of the code concerns only footer output.
Therefore, we will create a new `parseFooter()` to reduce complexity.

- [ ] Passed the CI test.


segayuu:
The issue detected by codeclimate is only an improved part of the issue before PR.

PR #3294, opened 2018-10-18
Refactoring let (e.g. reduce scope)

segayuu:
- [x] Passed the CI test.

### DIFF
--- a/lib/extend/filter.js
+++ b/lib/extend/filter.js
@@ -67,12 +67,11 @@ Filter.prototype.execSync = function(type, data, options = {}) {
   const filters = this.list(type);
   const ctx = options.context;
   const args = options.args || [];
-  let result;
 
   args.unshift(data);
 
   for (let i = 0, len = filters.length; i < len; i++) {
-    result = filters[i].apply(ctx, args);
+    const result = filters[i].apply(ctx, args);
     args[0] = result == null ? args[0] : result;
   }
 
--- a/lib/extend/filter.js
+++ b/lib/extend/filter.js
@@ -45,12 +45,9 @@ Filter.prototype.unregister = function(type, fn) {
   const list = this.list(type);
   if (!list || !list.length) return;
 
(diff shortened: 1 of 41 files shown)


segayuu:
The cause of failure only with currentVersion (11.0.0) is unknown.

tomap:
Also confused

tomap:
Quick way to fix/debug, split in two your changes, which seem pretty isolated, and make two PRs, and repeat with whichever fails...

tomap:
Oups, sorry, I just spotted that it has nothing to do with your changes. Maybe a minor update on one of our dependency triggered that...

segayuu:
As a result of examination, it turned out that it was caused by the difference of nodejs version.
The initial value of the `depth` option of `util.inspect()` has changed since node v11.
[node v11: `20`](https://nodejs.org/api/util.html#util_util_inspect_object_options) [node v10: `2`](https://nodejs.org/docs/latest-v10.x/api/util.html#util_util_inspect_object_options)
[Here is the test line where the error appears.](https://github.com/hexojs/hexo/blob/3.8.0/test/scripts/helpers/debug.js#L24)
Given the default value change, it is clear that the test on this line will fail on v11.

lib/extend/filter.js
```diff
@@ -70,12 +67,11 @@ Filter.prototype.execSync = function(type, data, options = {}) {
   const filters = this.list(type);
   const ctx = options.context;
   const args = options.args || [];
-  let result;
 
   args.unshift(data);
 
   for (let i = 0, len = filters.length; i < len; i++) {
-    result = filters[i].apply(ctx, args);
+    const result = filters[i].apply(ctx, args);
```
NoahDragon: I'm not understanding why we need to change this? Due to performance?

segayuu: If that comment is about this line, the purpose is to reduce the scope of `result`.

segayuu:
An error occurred because the conflict resolution failed. It is resolved now.

PR #3332, opened 2018-11-05
findIndex() to indexOf()

segayuu:

- [x] Passed the CI test.


tomap:
What is the purpose of this or?

segayuu:
`array.findIndex()` is more powerful than `array.indexOf()`, but since it uses a function, it is more verbose than `indexOf()`. For a simple comparison like `array.findIndex(item => item === foo)`, we can express it more easily as `array.indexOf(foo)`.

segayuu:
thank @tomap !

PR #3335, opened 2018-11-06
Refactor helper pagenator

segayuu:
The complexity of `pagenatorHelper` is `46`, it is too large as a single function, so we will distribute the complexity.

- [ ] Passed the CI test.


lib/plugins/helper/paginator.js
```diff
@@ -1,96 +1,136 @@
...
+};
 
-    // Display the current page
-    result += currentPage;
+function paginatorHelper(options = {}) {
+  options = Object.assign({
+    base: this.page.base || '',
+    current: this.page.current || 0,
+    format: `${this.config.pagination_dir}/%d/`,
+    total: this.page.total || 1,
```
JLHwung: Will `this.page.total` be zero?

segayuu: Since the initial initial value was 1, it is being handed over.

PR #3378, opened 2018-12-01
chore(package): specify files to shrink package size

JLHwung:
Specified size so that we will not publish package lock files to npm.

Before: 86.2 kB (308.5 kB)
After: 54.2 kB (199.5 kB)

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,10 @@
     "lib": "./lib",
     "bin": "./bin"
   },
+  "files": [
+    "lib/",
+    "bin/"
+  ],
   "repository": "hexojs/hexo",
   "homepage": "https://hexo.io/",
   "keywords": [


tomap:
You can see the content of the current package: https://unpkg.com/hexo@3.8.0/ it already includes only what is necessary
not sure your MR is useful

JLHwung:
@tomap Sorry for the late reply. The reason why I prefer to specify `dist` instead of `.npmignore` is that we are pretty sure `dist` is the root directory of our build artifacts and it should be complete and autonomous. If we do not specify `files` in the `package.json`, there will be, from time to time, maintainer have to remember to add any `yarn.lock`, `package-lock.json`, `.foorc` or `bar.config.js` files to the `.npmignore` so that the package size is optimal.

It is not easy and straightforward to have our maintainers remembering these process, especially when we are adding more and more dev infrastrurcture configurations.

tomap:
@JLHwung I agree, your way is better

tomap:
If you push a new commit, it will trigger a new build

YoshinoriN:
Hmm... 
Probably, the reason of failed the Appveyor that not this change.

OK, I do merge this PR.

PR #3404, opened 2018-12-28
feat(toc): seek anchor id from parent nodes when target is not available

chigix:
Signed-off-by: Richard Lea <chigix@zoho.com>

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/helper/toc.js
+++ b/lib/plugins/helper/toc.js
@@ -21,8 +21,14 @@ function tocHelper(str, options = {}) {
   let lastLevel = 0;
 
   headings.each(function() {
+    function getId(ele) {
+      const id = $(ele).attr('id');
+      return id ||
+        ($(ele).parent().is('html') ? null :
+          getId($(ele).parent()));
+    }
     const level = +this.name[1];
-    const id = $(this).attr('id');
+    const id = getId(this);
     const text = escape($(this).text());
 
     lastNumber[level - 1]++;
--- a/lib/plugins/helper/toc.js
+++ b/lib/plugins/helper/toc.js
@@ -20,13 +20,15 @@ function tocHelper(str, options = {}) {
   let firstLevel = 0;
   let lastLevel = 0;
 
+  function getId(ele) {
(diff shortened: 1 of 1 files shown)


YoshinoriN:
@chigix 
Thank you for your contribution.
I'm going to test this PR. Would you please tell me detail ?

chigix:
@YoshinoriN Thanks for reviewing.

This change is mainly focusing on HTML like the case below:

```html
<div id="data-summarise" class="section level2">
<h2>Data Summarise</h2>
<p>...</p>
</div>
```

, where the id is not assigned directly on the heading tag. However, this id anchor is still work for navigating to the target heading. 

So something mocking the template like the case above could be testing case for this change.

lib/plugins/helper/toc.js
```diff
@@ -21,8 +21,14 @@ function tocHelper(str, options = {}) {
   let lastLevel = 0;
 
   headings.each(function() {
+    function getId(ele) {
```
tcrowe: Please bring this `getId` function above `headings.each(...)` scope.

chigix: I see. I'll push tweaks later

tcrowe: Thanks for your patience with us

chigix:
@tcrowe I've pushed correction commits. May I have your review again?

YoshinoriN:
@tcrowe 
Would you like some help? Should I test it locally?

tcrowe:
Hey @YoshinoriN I tested it but I can't remember what the result was.

Do you have an easy way to get PR branches from contributors? Previously I was pulling from their repo but I think we can pull from PR number.

YoshinoriN:
Hi @tcrowe 

How about this?

```sh
git fetch origin pull/3404/head:BRANCHNAME
```

tcrowe:
@YoshinoriN Thanks! I'll try that method

tcrowe:
Thank you @chigix for the enhancement and @YoshinoriN for the PR help 🙌

Issue #3384, opened 2018-12-10
How can I make Hexo fail when partials are missing?


jakubgs:
I just noticed that the build succeeds when partials are missing:
```
r > node_modules/.bin/hexo generate
INFO  Start processing
INFO  Files loaded in 919 ms
WARN  Partial partial/shared-partials/partials/header-nav does not exist. (in partial/header.swig)
...(a lot omitted)...
WARN  Partial partial/shared-partials/partials/after_footer does not exist. (in layout.swig)
INFO  Generated: index.html
INFO  3 files generated in 2 s
```
Why is this just a warning? The page generated when partials are missing is broken. How can I make Hexo fail if even one partial is missing?

YoshinoriN:
I don't know why just a warning. It seems since [initial implement](https://github.com/hexojs/hexo/commit/1c2b23725ea983db9949e5a169207c2f0772b0dc#diff-ee8a1cd1cca459a05b11f850db3412a0R14).
IMHO, error or exception are better than warning.  I will try to create PR.

jakubgs:
I did the simplest thing possible - changing the warning into `throw new Error()` - in https://github.com/hexojs/hexo/pull/3413 since there doesn't seem to be much interest in this issue. Please tell me if anything should be changed.

PR #3413, opened 2019-01-07
throw error on missing partial

jakubgs:
Fixes: https://github.com/hexojs/hexo/issues/3384

### DIFF
--- a/lib/plugins/helper/partial.js
+++ b/lib/plugins/helper/partial.js
@@ -1,7 +1,6 @@
 'use strict';
 
 const { dirname, join } = require('path');
-const chalk = require('chalk');
 
 module.exports = ctx => function partial(name, locals, options = {}) {
   if (typeof name !== 'string') throw new TypeError('name must be a string!');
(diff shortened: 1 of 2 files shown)


tcrowe:
@jakubgs It's good and clean

@YoshinoriN You said it's better to throw than log a message here. I'm not sure so I will defer to your judgement.

YoshinoriN:
@tcrowe 
It is possibility that user can not notice about broken layout, if only a log message. I think an exception is better than log message, in order to notice it to user.

I'm going to test it by my local machine. But I have been busy recently with daily routine. 😓 So, please wait patiently. Thanks :)

tcrowe:
@YoshinoriN No problem! If you want to I can do it. You may hand over any task to me or assign it to me until you are ready to come back.

tomap:
There is an eslint error before validating

tcrowe:
The tests fail here:
https://github.com/hexojs/hexo/blob/master/test/scripts/helpers/partial.js#L46

jakubgs:
Now it seems the only blocker is the ridiculously small reduction in coverage. No idea what to do about that.

Issue #2610, opened 2017-06-13
Permalink not including root


miqmago:
## Environment Info

Node version(`node -v`): 7.8.9

Your site `_config.yml` (Optional):

```yml
root: /blog/
url: http://127.0.0.1:4000/       # Note: don't forget to modify the CNAME file to your url
permalink: :title-:abbrlink/
permalink_defaults:
```

Your theme `_config.yml` (Optional): 
Nothing relevant

Hexo and Plugin version(`npm ls --depth 0`): 

```
├── bootstrap@3.3.7
├── dialog-polyfill@0.4.7
├── gulp@3.9.1
├── gulp-connect-php@1.0.1
├── hexo@3.3.7
├── hexo-abbrlink@2.0.4
├── hexo-autonofollow@1.0.1
├── hexo-browsersync@0.2.0
├── hexo-deployer-git@0.1.0
├── hexo-filter-cleanup@1.0.7
├── hexo-generator-archive@0.1.4
├── hexo-generator-category@0.1.3
├── hexo-generator-index@0.2.1
├── hexo-generator-seo-friendly-sitemap@0.0.20
├── hexo-generator-tag@0.2.0
├── hexo-renderer-ejs@0.2.0
├── hexo-renderer-markdown-it@3.4.1
├── hexo-renderer-stylus@0.3.3
├── hexo-server@0.2.0
├── hexo-toc@1.0.8
├── jquery@3.2.1
├── marked@0.3.6
├── md-date-time-picker@2.3.0
├── moment@2.18.1
└── showdown@1.7.1
```

## For BUG

I've placed `<%= page.permalink %>` in `post.ejs` and the result is `http://127.0.0.1:4000/El-meu-primer-post-ada5698b/`

This link is not working as root is `/blog/`. Also `<%= config.url + config.root + (page.path||'').replace('index.html', '') %>` produces `http://127.0.0.1:4000/blog/El-meu-primer-post-ada5698b/` which is ok.

I've tried to change `permalink` directive in `_config.yml` to include `:root` with no success. Also this change breaks the links in home page. `index.ejs`:

```ejs
<% page.posts.each(function(post){ %>
<div class="post-preview">

    <a href="<%- config.root %><%- post.path %>">
```

Am I missing something?

NoahDragon:
The url should include the root path: `http://127.0.0.1:4000/blog/`

The new config file has mentioned this in comment:
[...]

PR #3604, opened 2019-06-29
Fix to use root to export Permalink

### LINKED ISSUE #3601: Permalink not including root
## Environment Info

This Issue is Continuation of #2610 .
This Issue is Closed, but the BUG is not fixed.

## For BUG

> I've placed <%= page.permalink %> in post.ejs and the result is http://127.0.0.1:4000/El-meu-primer-post-ada5698b/

> This link is not working as root is /blog/. Also <%= config.url + config.root + (page.path||'').replace('index.html', '') %> produces http://127.0.0.1:4000/blog/El-meu-primer-post-ada5698b/ which is ok.

This is very simple BUG, but I think this is IMPORTANT BUG.
Permalink not including root.
Deploy system will use the parameter `url`, `root`, `permalink`.

Now, `post.permalink` not including root : `url` + `permalink`.
It should be `url` + `root` + `permalink`

sKawashima:
## What does it do?
Fix to use root to export permalink.
It fix #3601 .

```js
[Page,Category,Tag].virtual('permalink').get(function() {
// return `${ctx.config.url}/${this.path}`; 
// to
  return `${ctx.config.url}${ctx.config.root}${this.path}`;
});
```

## How to test

```sh
cd hexo
npm install
npm test
```

### DIFF
--- a/lib/models/category.js
+++ b/lib/models/category.js
@@ -38,7 +38,7 @@ module.exports = ctx => {
   });
 
   Category.virtual('permalink').get(function() {
-    return `${ctx.config.url}/${this.path}`;
+    return `${ctx.config.url}${ctx.config.root}${this.path}`;
   });
 
   Category.virtual('posts').get(function() {
--- a/lib/models/page.js
+++ b/lib/models/page.js
@@ -32,7 +32,7 @@ module.exports = ctx => {
   });
 
   Page.virtual('permalink').get(function() {
-    return `${ctx.config.url}/${this.path}`;
+    return `${ctx.config.url}${ctx.config.root}${this.path}`;
   });
 
   Page.virtual('full_source').get(function() {
--- a/lib/models/tag.js
+++ b/lib/models/tag.js
@@ -29,7 +29,7 @@ module.exports = ctx => {
   });
 
   Tag.virtual('permalink').get(function() {
-    return `${ctx.config.url}/${this.path}`;
+    return `${ctx.config.url}${ctx.config.root}${this.path}`;
   });
 
   Tag.virtual('posts').get(function() {


sKawashima:
Codeclimate is not passed, but it was so before I touched it.

Please check.

YoshinoriN:
Sorry, We have many repositories and need to maintain them.
Would you please give us time? Thanks 😃

weyusi:
@https://github.com/hexojs/hexo/issues/2610#issuecomment-308456996
> In this case, why is `root` parameter for? Just for curiosity...
It's used in several tag plugins like include code to link to a local path via `<a href="/root_value/filename/">` instead of `<a href="http://yourweb.com/root_value/filename/">`,
https://github.com/hexojs/hexo/blob/da14a167ae476d3ac0542d1782eaed433139e8a6/lib/plugins/tag/include_code.js#L59

Post model replace the config.root with '/',
https://github.com/hexojs/hexo/blob/f172774d37f8dffe01d9acecd3613a8b6782e22a/lib/models/post.js#L57

The above works because the docs requires root value in both url and root.
https://github.com/hexojs/hexo-starter/blob/e5e9b59a5b66d4468467af4f1e931988aaa7047b/_config.yml#L14-L16

---
Alternatively, we can specify just one root value:
```yml
url: http://example.com
root: /blog/
```

The issue of this approach is that `<%= config.url %>` would be invalid, the proper one should be `<%= config.url + config.root %>`. This can be fixed by additional processing to append config.root to config.url, so config.url variable includes config.root. However, this also means that config.url no longer corresponds to the value in `_config.yml`.

I'm not so worry about the breaking change of the above approach, since v4 is on the horizon.
---

The [docs](https://hexo.io/docs/variables#Page-Variables) mention `The URL of current page without root URL. We usually use url_for(page.path) in theme.`. Should page.path includes config.root?

weyusi:
Currently the config requires root value in both url and root,
https://github.com/hexojs/hexo-starter/blob/e5e9b59a5b66d4468467af4f1e931988aaa7047b/_config.yml#L14-L16

For example,
```yml
url: http://yoursite.com/child
root: /child/
```

then

```js
return `${ctx.config.url}${ctx.config.root}${this.path}`;
```

will return `http://yoursite.com/child/child/path`.

YoshinoriN:
@weyusi 
Thank you for your support :)

`url` & `root` will be fixed following lines.
https://github.com/hexojs/hexo/blob/f172774d37f8dffe01d9acecd3613a8b6782e22a/lib/hexo/load_config.js#L32-L33

I'm going to approve and merge this PR. But, I'm not yet completed check current hexo source code. So, I haven't got confirmation of that yet.

weyusi:
@YoshinoriN 
I just tested the regex,
```js
config.root.replace(/\/*$/, '/');
```
this make sure the value ends with a slash, '/child' -> '/child/'

```js
config.url.replace(/\/+$/, '');
```
this make sure the url does not ends with a slash,
* 'http://example.com/' -> 'http://example.com'
* 'http://example.com/child/ -> 'http://example.com/child'.

Anyhow, 'child' value is still there. I just tested this PR and it definitely results in `http://yoursite.com/child/child/path`.

YoshinoriN:
@weyusi
Just to confirmation...

NOTE: I do not confirm it on the current master source code. Just confirmed on a console.

## Current hexo config & result

> `The URL of current page without root URL. We usually use url_for(page.path) in theme...`

My understanding its means...

### Pattern 1: OK

Settings

```yaml
url: http://yoursite.com
root: /
```

Result

```js
const urlOne = "http://example.com/".replace(/\/+$/, '');
const rootOne = "/".replace(/\/*$/, '/');

console.log('urlOne: ' + urlOne);
console.log('rootOne: ' + rootOne);
console.log('pattern1: ' + urlOne + rootOne + "\n");

// Result
urlOne: http://example.com
rootOne: /
pattern1: http://example.com/
```

### Pattern 2: OK

Settings

```yaml
url: http://yoursite.com
root: /child
```

Result

```js
const urlTwo = "http://example.com".replace(/\/+$/, '');
const rootTwo = "/child".replace(/\/*$/, '/');

console.log('urlTwo: ' + urlTwo);
console.log('rootTwo: ' + rootTwo);
console.log('pattern2: ' + urlTwo + rootTwo + "\n");

// Result
urlTwo: http://example.com
rootTwo: /child/
pattern2: http://example.com/child/
```

### Pattern 3: OK

Settings

```yaml
url: http://yoursite.com/
root: /child
```

Result

```js
const urlThree = "http://yoursite.com/".replace(/\/+$/, '');
const rootThree = "/child".replace(/\/*$/, '/');

console.log('urlThree: ' + urlThree);
console.log('rootThree: ' + rootThree);
console.log('pattern3: ' + urlThree + rootThree + "\n");

// Result
urlThree: http://yoursite.com
rootThree: /child/
pattern3: http://yoursite.com/child/
```

### Pattern 4: NG

Settings

```yaml
url: http://yoursite.com/
root: child
```

```js
const urlFour = "http://yoursite.com/".replace(/\/+$/, '');
const rootFour = "child".replace(/\/*$/, '/');

console.log('urlFour: ' + urlFour);
console.log('rootFour: ' + rootThree);
console.log('pattern4: ' + urlFour + rootFour + "\n");

// Result
urlFour: http://yoursite.com
rootFour: /child/
pattern4: http://yoursite.comchild/ <- it's wrong. But not relate this PR
```

My understanding is above. Is it okay??

## About this PR

This PR is not change `load_config.js`.
Maybe this PR source code will not affect `url + root` path.

So, I think maybe this PR is not breaking change & merge safely.

How do you think? :)

weyusi:
```
url: http://yoursite.com/child
root: /child/
```
is the [current approach](https://github.com/hexojs/hexo-starter/blob/bc3c31767bdbf0d9be2b24014ab8beb19c2acb50/_config.yml#L15). While I don't necessarily agree with that approach, unless the current approach is updated, this PR is a breaking change.

---

My suggestion is to switch to the following approach (a breaking change),
```
url: http://yoursite.com
root: /child
```

Then change,
https://github.com/hexojs/hexo/blob/f172774d37f8dffe01d9acecd3613a8b6782e22a/lib/hexo/load_config.js#L33
to
```js
    config.url = config.url.replace(/\/+$/, '').concat(config.root);
```
This will also fix #3601, without this PR.

weyusi:
Maybe @sKawashima @YoshinoriN haven't realize,
#3601 can also be easily fixed by,

```yml
url: http://yoursite.com/child
root: /child/
```

which is (I [iterate](https://github.com/hexojs/hexo/pull/3604#issuecomment-509901342) [again](https://github.com/hexojs/hexo/pull/3604#issuecomment-510340108)) the current approach. I understand @sKawashima might not like that approach, me neither. Most users either has "/" as `root:` value or grown accustomed to having "/child" in **both** `url:` and `root:`.

The rationale of using `url: http://yoursite.com/child` instead of `url: http://yoursite.com/` (when `root:` is _other than_ "/") is so that users can use `<%= config.url %>` to add a link to their home page, instead of using `<%= config.url + config.root %>`. I'm iterating my [previous comment](https://github.com/hexojs/hexo/pull/3604#issuecomment-509898551) here.

If user wants to use,
```yml
url: http://yoursite.com/
root: /child/
```
_without using_ `<%= config.url + config.root %>`, then hexo needs
```js
config.url = config.url.replace(/\/+$/, '').concat(config.root);
```
in [hexo/lib/hexo/load_config.js](https://github.com/hexojs/hexo/blob/f172774d37f8dffe01d9acecd3613a8b6782e22a/lib/hexo/load_config.js#L33).

---

So, the real question is, should
```yml
url: http://yoursite.com/child
root: /child/
```
be deprecated, in favor of
```yml
url: http://yoursite.com/
root: /child/
```
?

sKawashima:
@weyusi 
So why does `$ {ctx.config.url}`, `$ {ctx.config.root}` and `$ {this.path}` exist in this source code?

I have tried it once, but I sent this PR because I had a problem.
There are the behavior of link generation for share, the file arrangement at the time of generate, and the behavior of link at the time of rss generation in `hexo-generator-feed`.

These three parameters are called from various libraries for each role.
I can not understand why I do not distinguish it.

YoshinoriN:
@weyusi @sKawashima 
It's very sorry. I perfectly misunderstood.
I completely thought `_config.yml` setting is already as following.

```yaml
url: http://yoursite.com/
root: /child/
```

weyusi:
> So why does `$ {ctx.config.url}`, `$ {ctx.config.root}` and `$ {this.path}` exist in this source code?

Good question,
when you have the following config,
```yml
url: http://yoursite.com/child
root: /child/
```

and your page is located at `source/examplepage/index.md`,

Variable | Value
--- | ---
ctx.config.url | `http://yoursite.com/child`
ctx.config.root | `/child/`
page.path | `examplepage/index.html`
page.permalink | `ctx.config.url` + `/` + `page.path` = <br/> `http://yoursite.com/child/examplepage/index.html`

As for the purpose of `ctx.config.root`, as mentioned in my [previous comment](https://github.com/hexojs/hexo/pull/3604#issuecomment-509898551), it serves as another way to link to a same-origin page or assets.

Instead of using permalink which results in `<a href="http://yoursite.com/child/examplepage/index.html">`, hexo [also uses](https://github.com/hexojs/hexo/blob/da14a167ae476d3ac0542d1782eaed433139e8a6/lib/plugins/tag/include_code.js#L59) `ctx.config.root + page.path` which results in `<a href="/child/examplepage/index.html">`. I can't confirm why, but my guess is this approach allows distinguish of local and third-party links.

---

If you can pinpoint the exact issue (e.g. expected urls vs actual urls), it can help us to troubleshoot.

sKawashima:
I'm sorry for the late reply.
Thank you.

PR #3612, opened 2019-07-04
feat(filter): use existing excerpt if possible

ppoffice:
## What does it do?
Skip excerpt generation for posts or pages if it is set in the front matter. This gives users the ability to write down whatever abstract they want instead of just showing begining part of the post.

For example:
```yaml
title: An article about blah blah blah
excerpt: This article is all about blah blah blah blah blah blah
---
Actual markdown content...
```

will generate:

```js
{
    excerpt: 'This article is all about blah blah blah blah blah blah',
    more: 'Actual markdown content...',
    content: 'Actual markdown content...'
}
```

## How to test

```sh
git clone -b BRANCH https://github.com/ppoffice/hexo.git
cd hexo
npm install
npm test
```

## Screenshots

N/A

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/test/scripts/filters/excerpt.js
+++ b/test/scripts/filters/excerpt.js
@@ -119,4 +119,35 @@ describe('Excerpt', () => {
       'baz'
     ].join('\n'));
   });
+
+  it('skip processing if post/page.excerpt is present', () => {
+    const content = [
+      'foo',
+      '<!-- more -->',
+      'bar'
+    ].join('\n');
+
+    const data = {
+      content,
+      excerpt: 'baz'
+    };
+
+    excerpt(data);
+
+    data.content.should.eql([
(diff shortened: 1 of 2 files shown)


weyusi:
I prefer this approach (using it myself) than `<!--more-->`, looks cleaner. `<!--more-->` does have the advantage of supporting formatting though.

test/scripts/filters/excerpt.js
```diff
@@ -119,4 +119,35 @@ describe('Excerpt', () => {
       'baz'
     ].join('\n'));
   });
+
+  it('skip processing if post/page.excerpt is present', () => {
```
weyusi: Suggest "...is present in the front-matter".

PR #3668, opened 2019-08-15
fix(open_graph): remove duplicate twitter card tags

SukkaW:
## What does it do?

Remove `twitter:title` and `twitter:description` which is duplicated with `og:title` and `og:description`

According to the [Twitter Card Docs](https://developer.twitter.com/en/docs/tweets/optimize-with-cards/guides/getting-started.html):

> When using Open Graph protocol to describe data on a page, it is easy to generate a Twitter card without duplicating tags and data. When the Twitter card processor looks for tags on a page, it first checks for the Twitter-specific property, and if not present, falls back to the supported Open Graph property.

SInce the value of `twitter:title` and `twitter:description` is the same as `og:title` and `og:description`, they can be removed.

## How to test

```sh
cd hexo
npm install
npm test
```

## Pull request tasks

- [ ] Passed the CI test.

### DIFF
--- a/lib/plugins/helper/open_graph.js
+++ b/lib/plugins/helper/open_graph.js
@@ -116,10 +116,6 @@ function openGraphHelper(options = {}) {
   }
 
   result += meta('twitter:card', twitterCard);
-  result += meta('twitter:title', title);
-  if (description) {
-    result += meta('twitter:description', description, false);
-  }
 
   if (images.length) {
     result += meta('twitter:image', images[0], false);
--- a/test/scripts/helpers/open_graph.js
+++ b/test/scripts/helpers/open_graph.js
@@ -423,7 +423,7 @@ describe('open_graph', () => {
     result.should.not.contain(meta({property: 'og:updated_time', content: '2016-05-23T21:20:21.372Z'}));
   });
 
-  it('description - do not add /(?:og:|twitter:)?description/ meta tags if there is no description', () => {
+  it('description - do not add /(?:og:)?description/ meta tags if there is no description', () => {
     const result = openGraph.call({
       page: { },
       config: {},
@@ -431,7 +431,6 @@ describe('open_graph', () => {
     }, { });
 
     result.should.not.contain(meta({property: 'og:description'}));
-    result.should.not.contain(meta({property: 'twitter:description'}));
     result.should.not.contain(meta({property: 'description'}));
   });
 
--- a/test/scripts/helpers/open_graph.js
+++ b/test/scripts/helpers/open_graph.js
@@ -36,8 +36,7 @@ describe('open_graph', () => {
         meta({property: 'og:url'}),
         meta({property: 'og:site_name', content: hexo.config.title}),
         meta({property: 'og:updated_time', content: post.updated.toISOString()}),
-        meta({name: 'twitter:card', content: 'summary'}),
-        meta({name: 'twitter:title', content: hexo.config.title})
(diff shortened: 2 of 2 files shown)


curbengh:
please remove relevant lines in "test/scripts/helpers/open_graph.js".

test/scripts/helpers/open_graph.js
```diff
@@ -35,9 +35,9 @@ describe('open_graph', () => {
         meta({property: 'og:title', content: hexo.config.title}),
         meta({property: 'og:url'}),
         meta({property: 'og:site_name', content: hexo.config.title}),
+        meta({property: 'og:locale', content: hexo.config.language}),
```
curbengh: duplicate of https://github.com/hexojs/hexo/pull/3660

curbengh: please remove [this line](https://github.com/hexojs/hexo/blob/86316ddba28125cf3cbe3bcdbfb6ba326186891c/test/scripts/helpers/open_graph.js#L435) and update [this line](https://github.com/hexojs/hexo/blob/86316ddba28125cf3cbe3bcdbfb6ba326186891c/test/scripts/helpers/open_graph.js#L427).

SukkaW: I will rebase and remove that.

lib/plugins/helper/open_graph.js
```diff
@@ -116,10 +116,6 @@ function openGraphHelper(options = {}) {
   }
 
   result += meta('twitter:card', twitterCard);
-  result += meta('twitter:title', title);
-  if (description) {
-    result += meta('twitter:description', description, false);
-  }
 
   if (images.length) {
     result += meta('twitter:image', images[0], false);
```
YoshinoriN: `twitter:image` can deletable as well. Isn't it?

SukkaW: @YoshinoriN It seems like `og:image` support multiple images while `twitter:image` only support one image. The docs from twitter do not mention multiple images usage.

curbengh: [The doc](https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/markup.html) is not exactly clear.

> URL of image to use in the card. Images must be less than 5MB in size. JPG, PNG, WEBP and GIF formats are supported. Only the first frame of an animated GIF will be used. SVG is not supported.

I lean on single twitter:image and keeping it.

YoshinoriN: @SukkaW @curbengh 


I found following comments from twitter developers forum.

* https://twittercommunity.com/t/for-multiple-images-twitter-is-not-following-the-ogp-spec/15875 
* https://twittercommunity.com/t/twitter-cards-i-have-a-problem-help-me/15320/98
* https://twittercommunity.com/t/wrong-image-appearing-in-twitter-cards-using-wordpress-seo-by-yoast-plugin/15747/10

According to them. The twitter card use last `og:image` if some `og:image` tag exists. And it not following [ogp specification](https://ogp.me/#array).

They are very old comments (4~6 years ago). I could not find more latest. But, IMHO we should use `twitter:image` if we can not find more information about this matter.

I will merge this if other opinion does not exist after wait a while.

curbengh: > According to comments. The twitter card use last og:image if multiple og:image tags exists. It's not following ogp specification.

I figure most users would expect the first image, which is the current behavior of `open_graph.js`. Let's keep `twitter:image` for now.

Issue #3663, opened 2019-08-12
Performance Data of Hexo


SukkaW:
# Check List

Please check followings before submitting a new issue.

- [x] I have already confirmed other issue template and they are not different my want

## Environment & Settings

**Node.js & npm version**

Test with 8.16.0, 10.16.2 and 12.8.0

**Your package.json `package.json`**

https://github.com/theme-suka/hexo-theme-unit-test/blob/master/package.json

## Others

I have ran a performance test for Hexo using Travis CI with [300 posts](https://github.com/SukkaLab/hexo-many-posts)(All posts have 1 category and 5 tags, and all markdown styles and large amounts of code blocks are included in each .md file)

All the css, js & images are deleted before `hexo clean && hexo g` to make sure every file generated is rendered HTML.

| | NodeJS 8 | NodeJS 10 | NodeJS 12 |
|---|:---:|:---:|:---:|
| Enable Fragment Cache & Disable Highlight.js | 26.5s | 21s | 21.5s |
| Disable Fragment Cache | 35s | 29s | 29.5s |
| Enable highlight.js only | 35s | 27s | 27.5s |
| Enable highlight.js & line_number | 38s | 30s | 30s |
| Enable highlight.js & tab_replace | 34.5s | 27s | 28s |
| Enable highlight.js, tab_replace & line_number | 38s | 30s | 31s |

Raw log can be seen [here](https://travis-ci.com/theme-suka/performance-test).

It seems that `Disable Fragment Cache` will make hexo `35%` slower, `Enable highlight.js` will make hexo `45%` slower.
And even with Fragment enabled and Highlight.js disabled, hexo can only render 33 HTML per second.

tcrowe:
I had not heard of that setting so I wanted to help provide some context:
https://hexo.io/docs/helpers#fragment-cache
https://hexo.io/docs/templates#Optimization
https://github.com/search?q=org%3Ahexojs+fragment&type=Code

SukkaW:
@tcatche Fragment Cache was introduced in https://github.com/hexojs/hexo/pull/637 and the feature is depends on theme.
And Fragment Cache will be much more helpful if some time-consuming partial has been cached, seen https://github.com/hexojs/hexo/issues/1769

ppoffice:
@SukkaW Please comment out this line, rerun your tests and see how results are

https://github.com/hexojs/hexo/blob/master/lib/plugins/filter/index.js#L14

SukkaW:
[...]

PR #3670, opened 2019-08-15
perf(open_graph): only require cheerio when content contains '<img'

SukkaW:
## What does it do?

only require cheerio when content contains '<img'

## How to test

```sh
cd hexo
npm install
npm test
```

## Screenshots

https://github.com/hexojs/hexo/issues/3663#issuecomment-521629413

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/helper/open_graph.js
+++ b/lib/plugins/helper/open_graph.js
@@ -28,7 +28,6 @@ function og(name, content, escape) {
 }
 
 function openGraphHelper(options = {}) {
-  if (!cheerio) cheerio = require('cheerio');
 
   const { config, page } = this;
   const { content } = page;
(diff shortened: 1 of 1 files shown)


lib/plugins/helper/open_graph.js
```diff
@@ -59,12 +58,16 @@ function openGraphHelper(options = {}) {
   if (!images.length && content) {
```
curbengh: Is `if (!images.length && content && content.includes('<img'))` safe? I'm wondering if `content` can be null. Let me try with an empty file.

SukkaW: Maybe some syntax error might cause content to be null, I'm not sure.

curbengh:
My preference is on https://github.com/hexojs/hexo/pull/3680

SukkaW:
Superseded by #3680

PR #3673, opened 2019-08-16
feat(open_graph): only require moment when (page.updated)

SukkaW:
## What does it do?

From https://github.com/hexojs/hexo/issues/3663#issuecomment-521563403 we can know `open_graph` helper will slow down generating speed by 28%, and #3670 makes `open_graph` helper require `cheerio` only when needed. So I come up with this PR, which require `moment` when `page.updated` is given.

## How to test

```sh
cd hexo
npm install
npm test
```

## Screenshots

**Test A**

latest `hexojs/hexo#master` with hexo-theme-landscape and meta_generator set to false

image

**Test B**

image

`sukkaw/hexo#lazy-moment` with hexo-theme-landscape and meta_generator set to false

It is 1.4% faster.

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/helper/open_graph.js
+++ b/lib/plugins/helper/open_graph.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const urlFn = require('url');
-const moment = require('moment');
+
 const { escapeHTML, htmlTag, stripHTML } = require('hexo-util');
 let cheerio;
 
@@ -110,6 +110,8 @@ function openGraphHelper(options = {}) {
   });
 
   if (updated) {
+    const moment = require('moment');
+
     if ((moment.isMoment(updated) || moment.isDate(updated)) && !isNaN(updated.valueOf())) {
       result += og('og:updated_time', updated.toISOString());
     }


curbengh:
would it be beneficial if moment is lazy-loaded just like cheerio?

SukkaW:
> would it be beneficial if moment is lazy-loaded just like cheerio?

`moment` is not slow, but in `open_graph` helper it is only needed when `page.updated` is given. It isn't sound like `lazy`, much more like `conditional`.

After `conditional moment`, generate speed is `1.4%` faster. I think it worth a shot.

curbengh:
I'm considering adding [`article:published_time`](https://ogp.me/#type_article) tag, so a global moment may be needed since all pages/posts would have `date` key.

SukkaW:
@curbengh  Since most of the posts will have `post.date`, only page might not have date specified. I think conditional loading `moment` might have only tiny improvements. I will close this PR then.

curbengh:
When preparing https://github.com/hexojs/hexo/pull/3674, I found out there is an option to disable `og:updated` tag, so the PR also include an option to disable `article:published_time` as well.

Since each of the time-related meta tags can be disabled, so there may be situations where `moment` is not needed. At the same time, I figure only a handful of users would disable it, so on-demand `moment` would only benefits them, and even then (as you mentioned) the benefit may not be much.

Personally, I don't mind if moment is loaded like this,

``` js
let moment;

function openGraphHelper() {
  if (published) {
    if (!moment) moment = require('moment');
    result += moment(published)
  }

  if (updated) {
    if (!moment) moment = require('moment');
    result += moment(updated)
  }
}
```

SukkaW:
Since it might still have some benefit, maybe you can just include this way of loading moment in #3674

PR #3680, opened 2019-08-19
refactor(open_graph): drop cheerio and use regex

SukkaW:
## What does it do?

cheerio is slow and we only use that to parse and list img in post.content. This PR drop cheerio and use regex to do the job.

## How to test

```sh
cd hexo
npm install
npm test
```

## Screenshots
 
https://github.com/hexojs/hexo/pull/3680/commits/913f5bed0d39cacda66479bd58345992b64cfb8d

|:---:|:---:|:---:|:---:|
| cheerio@0.22 (hexojs/hexo#master) | **7.12** | 6.96 | 6.86 |
| regex (This PR) | 7.15 | 6.96 | **6.78** |
| dom-parser (#3679) | 7.2 | 6.93 | 6.82 |
| conditional cheerio (#3670) | 7.15 | **6.92** | **6.78** |

https://github.com/hexojs/hexo/pull/3680/commits/4f86ddef8f8f70193c4b590ac35d0eaad9028237

|:---:|:---:|:---:|:---:|
| cheerio@0.22 (hexojs/hexo#master) | 7.03 | 6.59 | 6.51 |
| regex (This PR) | 7.01 | 6.57 | 6.45 |
| dom-parser (#3679) | **6.98** | **6.56** | 6.41 |
| conditional cheerio (#3670) | 7.00 | 6.60 | **6.38** |

https://github.com/hexojs/hexo/pull/3680/commits/5c309eb109c55f78fd5c4c27f070e932c9150b03

|:---:|:---:|:---:|:---:|
| cheerio@0.22 (hexojs/hexo#master) | 6.96 | 6.65 | 6.85 |
| regex (This PR) | 6.90 | 6.68 | **6.73** |
| dom-parser (#3679) | 6.88 | 6.67 | **6.73** |
| conditional cheerio (#3670) | **6.72** | **6.60** | 6.80 |

https://github.com/hexojs/hexo/pull/3680/commits/c11ded4735d702890cc0f42298fcf1cde53b4c15

|:---:|:---:|:---:|:---:|
| cheerio@0.22 (hexojs/hexo#master) | 7.29 | 6.54 | 6.93 |
| regex (This PR) | 7.21 | **6.47** | 6.84 |
| dom-parser (#3679) | 7.16 | 6.50 | 6.87 |
| conditional cheerio (#3670) | **7.06** | 6.50 | **6.81** |

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/helper/open_graph.js
+++ b/lib/plugins/helper/open_graph.js
@@ -3,7 +3,6 @@
 const urlFn = require('url');
 const moment = require('moment');
 const { escapeHTML, htmlTag, stripHTML } = require('hexo-util');
-let cheerio;
 
 function meta(name, content, escape) {
   if (escape !== false && typeof content === 'string') {
@@ -28,7 +27,6 @@ function og(name, content, escape) {
 }
 
 function openGraphHelper(options = {}) {
-  if (!cheerio) cheerio = require('cheerio');
 
   const { config, page } = this;
   const { content } = page;
@@ -59,12 +57,18 @@ function openGraphHelper(options = {}) {
   if (!images.length && content) {
     images = images.slice();
 
-    const $ = cheerio.load(content);
+    if (content.includes('<img')) {
+      const imgPattern = /<img.*?src="(.*?)".*?\/?>/gi;
+      const srcPattern = /src=['"]?([^'"]*)['"]?/i;
+      const imgArray = content.match(imgPattern);
+      for (let img of imgArray) {
+        const src = img.match(srcPattern);
+        if (src[1]) {
+          images.push(src[1]);
+        }
+      }
+    }
 
-    $('img').each(function() {
-      const src = $(this).attr('src');
-      if (src) images.push(src);
-    });
   }
 
   let result = '';
--- a/lib/plugins/helper/open_graph.js
+++ b/lib/plugins/helper/open_graph.js
@@ -63,9 +63,7 @@ function openGraphHelper(options = {}) {
       const imgArray = content.match(imgPattern);
       for (let img of imgArray) {
         const src = img.match(srcPattern);
-        if (src[1]) {
-          images.push(src[1]);
-        }
+        if (src) images.push(src[1]);
       }
     }
 
--- a/lib/plugins/helper/open_graph.js
+++ b/lib/plugins/helper/open_graph.js
@@ -58,12 +58,10 @@ function openGraphHelper(options = {}) {
     images = images.slice();
 
     if (content.includes('<img')) {
-      const imgPattern = /<img.*?>/gi;
-      const srcPattern = /src=['"]?([^'"]*)['"]?/i;
-      const imgArray = content.match(imgPattern);
-      for (let img of imgArray) {
-        const src = img.match(srcPattern);
-        if (src) images.push(src[1]);
+      let img;
+      const imgPattern = /<img [^>]*src=['"]([^'"]+)([^>]*>)/gi;
+      while ((img = imgPattern.exec(content)) !== null) {
+        images.push(img[1]);
       }
     }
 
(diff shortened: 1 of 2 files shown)


SukkaW:
For regex, I have test those pattern with [regex101](https://regex101.com/) to match [this post](https://blog.skk.moe/post/how-verizon-knocked-large-parts-of-the-internet-offline/):

```
/<img.*?src="(.*?)".*?\/?>/gi
```

6491 steps, 9 matches, ~3ms: https://regex101.com/r/N1Rv9w/1

```
/<img.*?(?:>|\/>)/gi
```

8858 steps, 9 matches, ~5ms: https://regex101.com/r/N1Rv9w/2

```
/<img(.*?)>/gi
```

6527 steps, 9 matches, ~4ms: https://regex101.com/r/N1Rv9w/3

```
<img.*?>
```

4313 steps, 9 matched, ~3ms: https://regex101.com/r/N1Rv9w/4

```
<img[^>]*src[="'\s]+[^.]*\/([^.]+)\.[^"']+["']?[^>]*>
```

4627 steps, 9 matched, ~4ms: https://regex101.com/r/N1Rv9w/5

Currently in this PR I am using the first one.

lib/plugins/helper/open_graph.js
```diff
@@ -59,12 +57,18 @@ function openGraphHelper(options = {}) {
...
     images = images.slice();
 
-    const $ = cheerio.load(content);
+    if (content.includes('<img')) {
+      const imgPattern = /<img.*?>/gi;
+      const srcPattern = /src=['"]?([^'"]*)['"]?/i;
+      const imgArray = content.match(imgPattern);
+      for (let img of imgArray) {
+        const src = img.match(srcPattern);
+        if (src[1]) {
```
segayuu: Since the img element requires the src attribute, the null check of String # match () will not be necessary in the normal system.
If there is an img element without src, it should be treated as abnormal, but there is room for consideration in error handling.

SukkaW: Since it is using regex rather than a dom parser to match the `` tag, so an `` tag without src could happened. If there is an `` tag matches `imgPattern` but without `src` attribute, `img.match(srcPattern)` will return `null`. So should I use `if (src)` rather than `if (src[1])`?

segayuu: If you want to ignore an `img` element with an empty `src`, the following changes will eliminate unnecessary checks.
`const srcPattern = /src=['"]?([^'"]+)['"]?/i;`
`if (src) images.push(src[1]); `

SukkaW: OK, I will change to that.

segayuu: Thank you for the correction.
However, with just this one line, even if the value of `src` is empty, it will be pushed to `images`.
If the value of `src` is intended to be ignored if it is empty, you need to change the definition of the `srcPattern` variable.

SukkaW:
I think we still need to dig into those PR: #3679 #3670 and this PR. They all increase performance, but not much.

lib/plugins/helper/open_graph.js
SukkaW: I have pushed a new commit with a new regex pattern:

```
/<img [^>]*src=['"]([^'"]+)([^>]*>)/gi
```

This one will only match the `` tag with non-empty `src` attribute.

curbengh:
Since this PR only uses native API, naturally it supersedes #3679 #3670. This is unless there is possibility of some edge cases, which I currently don't foresee.

SukkaW:
@segayuu @curbengh Should I add some test cases for `` tags with empty `src` attribute?

SukkaW:
I have update the final benchmark for this PR & #3670

|:---:|:---:|:---:|:---:|
| cheerio@0.22 (hexojs/hexo#master) | 8.75 | 7.0 | 7.82 |
| regex (This PR) | **6.51** | **5.52** | **6.44** |
| conditional cheerio (#3670) | 7.68 | 6.49 | 7.51 |

The last benchmark shows this PR increase performance by nearly `20%`.

I will merge this PR and close #3670 then.

PR #3685, opened 2019-08-22
perf(external_link): drop cheerio and use regex

SukkaW:
## What does it do?

Drop cheerio in `external_link()` helper for better performance.

## How to test

```sh
git clone -b drop-cheerio-for-external-link https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Screenshots

|:---:|:---:|:---:|:---:|
| cheerio@0.22 (hexojs/hexo#master) | 7.64 | 6.62 | 7.54 |
| regex (This PR) | **7.60** | **6.57** | **7.40** |

|:---:|:---:|:---:|:---:|
| cheerio@0.22 (hexojs/hexo#master) | 6.55 | 5.53 | 6.15 |
| regex (This PR) | **6.48** | **5.42** | **6.10** |

|:---:|:---:|:---:|:---:|
| cheerio@0.22 (hexojs/hexo#master) | 6.12 | 5.92 | 6.57 |
| regex (This PR) | **6.03** | **5.88** | **6.44** |

The performance increased by nearly `1%`.

## Pull request tasks

- [x] Passed the CI test.


SukkaW:
@segayuu I am thinking is it worthy.

In this PR I drop cheerio, completely rewrite this part use regex and `replace()`, but only result in less than `1%` performance increases.

curbengh:
This PR drops cheerio (along with its deps) and replace it with even less LOC than before. The way I see it, this is equivalent to >99% reduction of LOC. In this case, I'm certain regex+replace can perform at least as fast as cheerio.

LOC reduction + no loss in performance = 💯% worth it.

test/scripts/filters/external_link.js
```diff
@@ -26,13 +26,19 @@ describe('External link', () => {
       '# External link test',
       '1. External link',
       '<a href="https://hexo.io/">Hexo</a>',
-      '2. Internal link',
+      '2. External link with "rel" Attribute',
```
curbengh: suggest to add another test with `<a rel="noopener" href="https://hexo.io/">Hexo</a>` to be certain there would no duplicate `noopener`.

SukkaW: OK, I will add that test cases.

curbengh: I might have missed line 32. Anyway, it's resolved.

test/scripts/filters/external_link.js
```diff
@@ -26,13 +26,23 @@ describe('External link', () => {
...
+      '<a href="https://hexo.io/" rel="noopenner">Hexo</a>',
+      '<a rel="external noopenner" href="https://hexo.io/">Hexo</a>',
+      '<a href="https://hexo.io/" rel="external noopenner">Hexo</a>',
+      '3. External link with Other Attributes',
+      '<a class="img" href="https://hexo.io/">Hexo</a>',
+      '<a href="https://hexo.io/" class="img">Hexo</a>',
+      '4. Internal link',
       '<a href="/archives/foo.html">Link</a>',
-      '3. Ignore links have "target" attribute',
+      '5. Ignore links have "target" attribute',
```
curbengh: My preference is to have "noopener" when target is "_blank", and ignore links with target value other than "_blank".

Maybe this is a way to disable "rel=noopener" specifically without disabling "config.external_link"?

SukkaW: It can disable `rel=noopener but I don't know if it is designed to do that. Since this PR only dropped the cheerio, I just keep that behavior.

curbengh: Surprise feature huh 🤣 . I think it's ok to have keep it.

If the behavior is undesired, user can always remove `target="_blank"` and enable `external_link`.

PR #3696, opened 2019-08-31
refactor(meta_generator): use template string

SukkaW:
## What does it do?

Only a minor changes, use template string instead of `replace()`.

## How to test

```sh
git clone -b meta-generator https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/filter/meta_generator.js
+++ b/lib/plugins/filter/meta_generator.js
@@ -5,8 +5,7 @@ function hexoMetaGeneratorInject(data) {
   if (!config.meta_generator ||
   data.match(/<meta\s+name=['|"]?generator['|"]?/i)) return;
 
-  const hexoGeneratorTag = '<meta name="generator" content="Hexo %s">'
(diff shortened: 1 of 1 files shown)


curbengh:
btw, I just found an edge case for this helper. `<title>` is [also used](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title) in `<svg>`, so it can issue when inline embed svg.

A more robust approach is to only append to `</title>` _inside_ `<head></head>`.

SukkaW:
@curbengh 

```javascript
/<head>.*?<title>.*?(<\/title>).*?<\/head>/
```

I think this pattern will do the job. I will open another PR.

SukkaW:
```javascript
<head>(?!<\/head>).+?<\/head>
```

If we have to back to replace `</head>`, we might use this pattern to avoid `<head></head>` issue.

PR #3697, opened 2019-08-31
refactor(meta_generator): update regex and avoid edge case

SukkaW:
## What does it do?

After #3671 & #3696 , meta_generator filter will replace `</title>` with `<meta name="generator" content="Hexo ${hexo.version}"></title>`.

@curbengh [has mentioned](https://github.com/hexojs/hexo/pull/3696#issuecomment-526819260) some [edge cases](https://github.com/hexojs/hexo/pull/3696#issuecomment-526822365) will cause unwanted behavior:

```html
<html>
<head>
<!-- A inline svg in css might include <title> tag -->
<style>
  .foo {
    background: url(data:image/svg+xml;utf8,<svg width="500" height="300" xmlns="http://www.w3.org/2000/svg"><g><title>svg</title></g></svg>);
  }    
</style>
<title>TITLE</title>
</head>
...
</html>
```

This PR introduce replacing `</head>` with `<meta name="generator" content="Hexo ${hexo.version}"></head>` with [a pattern](https://regex101.com/r/4yIsOT/1/) that [ignore empty `<head>` tag](https://github.com/hexojs/hexo/pull/3315).

## How to test

```sh
git clone -b meta-generator https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

A test case about `ignore empty head tag` has been added in this PR as well. 

## Screenshots

According to [benchmark](https://travis-ci.com/theme-suka/performance-test), use regex to match `</head>` is about `1%` slower than replace `</title>` directly.

|:---:|:---:|:---:|:---:| 
| Disable `meta_generator` | 6.15s | 5.18s | 5.85s  |
| Replace `</title>` directly (current master branch) | 6.25s | 5.25s | 5.94s |
| This PR | 6.3s | 5.30s | 5.94s |

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/test/scripts/filters/meta_generator.js
+++ b/test/scripts/filters/meta_generator.js
@@ -7,7 +7,7 @@ describe('Meta Generator', () => {
   const cheerio = require('cheerio');
 
   it('default', () => {
-    const content = '<head><title>foo</title></head>';
+    const content = '<head><link></head>';
     const result = metaGenerator(content);
 
     const $ = cheerio.load(result);
@@ -16,7 +16,7 @@ describe('Meta Generator', () => {
   });
 
   it('disable meta_generator', () => {
-    const content = '<head><title>foo</title></head>';
+    const content = '<head><link></head>';
     hexo.config.meta_generator = false;
     const result = metaGenerator(content);
 
@@ -25,7 +25,7 @@ describe('Meta Generator', () => {
   });
 
   it('no duplicate generator tag', () => {
-    const content = '<head><title>foo</title>'
+    const content = '<head><link>'
       + '<meta name="generator" content="foo"></head>';
     hexo.config.meta_generator = true;
     const result = metaGenerator(content);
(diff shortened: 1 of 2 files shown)


test/scripts/filters/meta_generator.js
```diff
@@ -32,4 +33,22 @@ describe('Meta Generator', () => {
     const resultType = typeof result;
     resultType.should.eql('undefined');
   });
+
+  it('ignore empty head tag', () => {
```
curbengh: Other tests need to be updated, as the filter no longer seeks `<title>`.

test/scripts/filters/meta_generator.js
```diff
@@ -32,4 +33,22 @@ describe('Meta Generator', () => {
...
+    const content = '<head></head>'
+      + '<head><link></head>'
+      + '<head><link></head>'
```
curbengh: Looks like there are two tests here; one is "ignore empty head tag", the other is "add to the first non-empty head tag". Suggest to split them up.

SukkaW: 'apply to first non-empty head tag only' is now separated from 'ignore empty head tag'.

test/scripts/filters/meta_generator.js
SukkaW: Updated.

PR #3704, opened 2019-09-06
fix post_link, asset_link when title contains <,>," charaters

dailyrandomphoto:
## What does it do?

When the title contains `<,>,"` characters, it will cause some errors on browsers.

e.g.
the titles are
- `this is a title with <a tag>.`
- `this is a title with " class="badname`
```
- {% post_link test1 %}
- {% post_link test2 %}
```
generates to
```
<li><a href="/posts/test1/" title="this is a title with <a tag>.">this is a title with </a><a tag="">.</a></li>
<li><a href="/posts/test2/" title="this is a title with " class="badname">this is a title with " class="badname</a>
</li>
```

hexo-2019090602

### Solution:
use `util.escapeHTML` function escape the titles.

how to use `post_link` with `escape` option:
```javascript
/**
 * Post link tag
 *
 * Syntax:
 *   {% post_link slug [title] [escape] %}
 */
escape = true or omit; // escape
escape = false; // do not escape
```

e.g.
```ejs
# escape title
- {% post_link test1 %}
- {% post_link test1 '"special" custom <title>' %}
- {% post_link test1 true %}
- {% post_link test1 '"special" custom <title>' true %}

# do not escape title
- {% post_link test2 false %}
- {% post_link test2 '<b>bold custom title</b>' false %}
```

## How to test

```sh
git clone -b fix-post_link-title https://github.com/dailyrandomphoto/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/tag/asset_link.js
+++ b/lib/plugins/tag/asset_link.js
@@ -7,7 +7,7 @@ const { escapeHTML } = require('hexo-util');
  * Asset link tag
  *
  * Syntax:
- *   {% asset_link slug [escape] [title] %}
+ *   {% asset_link slug [title] [escape] %}
  */
 module.exports = ctx => {
   const PostAsset = ctx.model('PostAsset');
@@ -19,20 +19,16 @@ module.exports = ctx => {
     const asset = PostAsset.findOne({post: this._id, slug});
     if (!asset) return;
 
-    let escape = args[0];
+    let escape = args[args.length - 1];
     if (escape === 'true' || escape === 'false') {
-      args.shift();
+      args.pop();
     } else {
       escape = 'true';
     }
 
     let title = args.length ? args.join(' ') : asset.slug;
-    let attrTitle;
-    if (escape === 'true') {
-      attrTitle = title = escapeHTML(title);
-    } else {
-      attrTitle = escapeHTML(title);
-    }
+    const attrTitle = escapeHTML(title);
+    if (escape === 'true') title = attrTitle;
 
     return `<a href="${url.resolve(ctx.config.root, asset.path)}" title="${attrTitle}">${title}</a>`;
   };
--- a/lib/plugins/tag/asset_link.js
+++ b/lib/plugins/tag/asset_link.js
@@ -1,8 +1,7 @@
 'use strict';
 
 const url = require('url');
-const util = require('hexo-util');
-const escapeHTML = util.escapeHTML;
+const { escapeHTML }  = require('hexo-util');
 
 /**
  * Asset link tag
--- a/lib/plugins/tag/asset_link.js
+++ b/lib/plugins/tag/asset_link.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const url = require('url');
-const { escapeHTML }  = require('hexo-util');
+const { escapeHTML } = require('hexo-util');
 
 /**
  * Asset link tag
--- a/lib/plugins/tag/asset_link.js
+++ b/lib/plugins/tag/asset_link.js
@@ -23,7 +23,7 @@ module.exports = ctx => {
     if (escape === 'true' || escape === 'false') {
       args.shift();
     } else {
-      escape = 'false';
+      escape = 'true';
     }
 
     let title = args.length ? args.join(' ') : asset.slug;
--- a/lib/plugins/tag/asset_link.js
+++ b/lib/plugins/tag/asset_link.js
@@ -7,7 +7,7 @@ const { escapeHTML } = require('hexo-util');
  * Asset link tag
  *
  * Syntax:
- *   {% asset_link slug [title] %}
+ *   {% asset_link slug [escape] [title] %}
  */
 module.exports = ctx => {
   const PostAsset = ctx.model('PostAsset');
@@ -19,8 +19,21 @@ module.exports = ctx => {
     const asset = PostAsset.findOne({post: this._id, slug});
     if (!asset) return;
 
-    const title = escapeHTML(args.length ? args.join(' ') : asset.slug);
+    let escape = args[0];
+    if (escape === 'true' || escape === 'false') {
+      args.shift();
+    } else {
+      escape = 'false';
+    }
 
-    return `<a href="${url.resolve(ctx.config.root, asset.path)}" title="${title}">${title}</a>`;
+    let title = args.length ? args.join(' ') : asset.slug;
+    let attrTitle;
+    if (escape === 'true') {
+      attrTitle = title = escapeHTML(title);
+    } else {
+      attrTitle = escapeHTML(title);
+    }
+
+    return `<a href="${url.resolve(ctx.config.root, asset.path)}" title="${attrTitle}">${title}</a>`;
   };
 };
--- a/lib/plugins/tag/asset_link.js
+++ b/lib/plugins/tag/asset_link.js
@@ -1,6 +1,8 @@
 'use strict';
 
 const url = require('url');
+const util = require('hexo-util');
+const escapeHTML = util.escapeHTML;
 
 /**
  * Asset link tag
@@ -18,7 +20,7 @@ module.exports = ctx => {
     const asset = PostAsset.findOne({post: this._id, slug});
     if (!asset) return;
 
-    const title = args.length ? args.join(' ') : asset.slug;
+    const title = escapeHTML(args.length ? args.join(' ') : asset.slug);
 
     return `<a href="${url.resolve(ctx.config.root, asset.path)}" title="${title}">${title}</a>`;
   };
--- a/test/scripts/tags/asset_link.js
+++ b/test/scripts/tags/asset_link.js
@@ -44,6 +44,10 @@ describe('asset_link', () => {
     assetLink('bar Hello world').should.eql('<a href="/foo/bar" title="Hello world">Hello world</a>');
   });
 
+  it('title with tag', () => {
+    assetLink('bar Hello <world>').should.eql('<a href="/foo/bar" title="Hello &lt;world&gt;">Hello &lt;world&gt;</a>');
+  });
+
   it('with space', () => {
     // {% asset_link "spaced asset" "spaced title" %}
     assetLinkTag.call(post, ['spaced asset', 'spaced title'])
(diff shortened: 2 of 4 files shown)


lib/plugins/tag/asset_link.js
```diff
@@ -1,6 +1,8 @@
 'use strict';
 
 const url = require('url');
+const util = require('hexo-util');
```
curbengh: suggest to refactor L4-5 to `const { escapeHTML } = require('hexo-util');`. Apply to "lib/plugins/tag/post_link.js " too.

test/scripts/tags/asset_link.js
```diff
@@ -44,6 +44,10 @@ describe('asset_link', () => {
     assetLink('bar Hello world').should.eql('<a href="/foo/bar" title="Hello world">Hello world</a>');
   });
 
+  it('title with tag', () => {
```
curbengh: suggest `should escape tag in title`.

test/scripts/tags/post_link.js
```diff
@@ -22,6 +27,14 @@ describe('post_link', () => {
     postLink(['foo', 'test']).should.eql('<a href="/foo/" title="test">test</a>');
   });
 
+  it('default with tag', () => {
```
curbengh: "should escape tag in title"

test/scripts/tags/post_link.js
```diff
@@ -22,6 +27,14 @@ describe('post_link', () => {
...
+    postLink(['title-with-tag']).should.eql('<a href="/title-with-tag/" title="Hello &lt;new world&gt;!">Hello &lt;new world&gt;!</a>');
+  });
+
+  it('title with tag', () => {
```
curbengh: "should escape tag in custom title"

curbengh:
There is a possible breaking change with ejs theme layout, when user wants to markup the title in unescaped form, e.g. `<%- post.title %>`.

I suggest to escape in [`post_link`](https://github.com/hexojs/hexo/blob/master/lib/plugins/tag/post_link.js) tag plugin instead.

lib/plugins/tag/asset_link.js
dailyrandomphoto: @curbengh thank you for review. I have fixed them.

dailyrandomphoto:
> Just realized there is a possible breaking change with ejs theme layout, i.e. when user wants to markup the title, like
> 
> ```yaml
> ---
> title: This is a <b>Bold</b> statement
> ---
> ```
> 
> , so need to display it in unescaped form. `<%- post.title %>`.
> 
> I suggest to escape in [`post_link`](https://github.com/hexojs/hexo/blob/master/lib/plugins/tag/post_link.js) tag plugin instead.

@curbengh `post_link` is a tag using in the Markup. (`{% post_link test1 %}`)
To escape `post_link` tag, `<%= post_link test1 %>` or `{%= post_link test1 %}` does not work.

And, as I know most themes escape titles by default now. `<%= post.title %>`.
Users can't use markuped title without modify the themes.

dailyrandomphoto:
@curbengh I have refactored this feature.
Keep the behavior of the original code,
use `escape` option to escape title.

how to use `post_link` with `escape` option:
```javascript
/**
 * Post link tag
 *
 * Syntax:
 *   {% post_link slug [escape] [title] %}
 */
escape = true; // escape
escape = false or omit; // do not escape
```

e.g.
```ejs
# escape title
- {% post_link test1 true %}
- {% post_link test1 true custom title %}

# do not escape title
- {% post_link test2 false %}
- {% post_link test2 false custom title %}
- {% post_link test2 %}
- {% post_link test2 custom title %}
```

curbengh:
what I meant is 
https://github.com/hexojs/hexo/blob/d403b70dc507ac617584613be9142548015f6f61/lib/plugins/tag/post_link.js#L21

becomes

``` js
return `<a href="${ctx.config.root}${post.path}" title="${escapeHTML(title)}">${escapeHTML(title)}</a>`;
```

> And, as I know most themes escape titles by default now. `<%= post.title %>`.
you're right. As such, having `escape` option is not necessary.
I need to test whether `<%= post.title %>` will get double escaped (most probably not, just want to make sure).

dailyrandomphoto:
@curbengh 

> have you tested whether `<%= post.title %>` will double escape (most probably not, just want to make sure)?

Yes, `<%= post.title %>` will double escape title when post.title was escaped before in some filter or somewhere else.
e.g.
```
<b>Bold</b>
=>
&lt;b&gt;Bold&lt;&#x2F;b&gt;
=>
&amp;lt;b&amp;gt;Bold&amp;lt;&amp;#x2F;b&amp;gt;
```
<b>Bold</b>
=>
&lt;b&gt;Bold&lt;&#x2F;b&gt;
=>
&amp;lt;b&amp;gt;Bold&amp;lt;&amp;#x2F;b&amp;gt;

curbengh:
I see, then we can only escape in tag plugins. Another tag plugin that also need to escape is `{% asset_img %}`, but that can be done https://github.com/hexojs/hexo-util/blob/master/lib/html_tag.js (PR welcome 😄).

I suggest to remove `escape` option, I believe title should always be escaped (in the tag plugins).

lib/plugins/tag/asset_link.js
```diff
@@ -18,8 +19,21 @@ module.exports = ctx => {
...
+      escape = 'false';
+    }
 
-    return `<a href="${url.resolve(ctx.config.root, asset.path)}" title="${title}">${title}</a>`;
+    let title = args.length ? args.join(' ') : asset.slug;
+    let attrTitle;
+    if (escape === 'true') {
+      attrTitle = title = escapeHTML(title);
+    } else {
+      attrTitle = escapeHTML(title);
```
curbengh: In retrospect, I think it's fine to disable escaping the `title`, but `attrTitle` should _always_ be escaped.

curbengh:
@dailyrandomphoto I would like to merge this as part of hexo v4, consider this PR high priority.

dailyrandomphoto:
@curbengh 
Instead of removing the `escape` option, I have changed default value of `escape` option to `true`.

Is there anything else I need to fix?

### how to use `post_link` with `escape` option:
```javascript
/**
 * Post link tag
 *
 * Syntax:
 *   {% post_link slug [escape] [title] %}
 */
escape = true or omit; // escape
escape = false; // do not escape
```

e.g.
```ejs
# escape title
- {% post_link test1 %}
- {% post_link test1 '"special" custom <title>' %}
- {% post_link test1 true %}
- {% post_link test1 true '"special" custom <title>' %}

# do not escape title
- {% post_link test2 false %}
- {% post_link test2 false '<b>bold custom title</b>' %}
```

lib/plugins/tag/asset_link.js
```diff
@@ -1,12 +1,13 @@
...
 
 const url = require('url');
+const { escapeHTML } = require('hexo-util');
 
 /**
  * Asset link tag
  *
  * Syntax:
- *   {% asset_link slug [title] %}
+ *   {% asset_link slug [escape] [title] %}
```
curbengh: try to put (if possible) `escape` as the last since it's a new parameter.

lib/plugins/tag/asset_link.js
```diff
@@ -18,8 +19,21 @@ module.exports = ctx => {
...
+      args.shift();
+    } else {
+      escape = 'true';
+    }
 
-    return `<a href="${url.resolve(ctx.config.root, asset.path)}" title="${title}">${title}</a>`;
+    let title = args.length ? args.join(' ') : asset.slug;
+    let attrTitle;
+    if (escape === 'true') {
+      attrTitle = title = escapeHTML(title);
```
curbengh: `if (escape === 'true') title = escapeHTML(title);`, no need `else`.

lib/plugins/tag/asset_link.js
```diff
@@ -18,8 +19,21 @@ module.exports = ctx => {
...
+    let escape = args[0];
+    if (escape === 'true' || escape === 'false') {
+      args.shift();
+    } else {
+      escape = 'true';
+    }
 
-    return `<a href="${url.resolve(ctx.config.root, asset.path)}" title="${title}">${title}</a>`;
+    let title = args.length ? args.join(' ') : asset.slug;
+    let attrTitle;
```
curbengh: `const attrTitle = escapeHTML(title);`

curbengh:
> Instead of removing the `escape` option, I have changed default value of `escape` option to `true`.

Thanks for your prompt response.

dailyrandomphoto:
@curbengh 
the `escape` option is the last parameter now.

### how to use `post_link` with `escape` option:
```javascript
/**
 * Post link tag
 *
 * Syntax:
 *   {% post_link slug [title] [escape] %}
 */
escape = true or omit; // escape
escape = false; // do not escape
```

e.g.
```ejs
# escape title
- {% post_link test1 %}
- {% post_link test1 '"special" custom <title>' %}
- {% post_link test1 true %}
- {% post_link test1 '"special" custom <title>' true %}

# do not escape title
- {% post_link test2 false %}
- {% post_link test2 '<b>bold custom title</b>' false %}
```

test/scripts/tags/asset_link.js
```diff
@@ -49,11 +49,11 @@ describe('asset_link', () => {
   });
 
   it('should escape tag in title', () => {
-    assetLink('bar true "Hello" <world>').should.eql('<a href="/foo/bar" title="&quot;Hello&quot; &lt;world&gt;">&quot;Hello&quot; &lt;world&gt;</a>');
+    assetLink('bar "Hello" <world> true').should.eql('<a href="/foo/bar" title="&quot;Hello&quot; &lt;world&gt;">&quot;Hello&quot; &lt;world&gt;</a>');
```
curbengh: the purpose is to test default option, so shouldn't need to specify `true`.

dailyrandomphoto: there are 3 test cases
- option omitted(default): should escape tag in title by default
- true: should escape tag in title
- false: should not escape tag in title

curbengh: sorry, missed that.

PR #3754, opened 2019-10-08
using Class

deepnum:


SukkaW:
The PR looks good, but would you mind making eslint happy? It seems `L349` has broke the eslint.

curbengh:
@deepnum 

Removing the semicolon didn't resolve it?

``` diff
- .then(locals => { route.set(path, createLoadThemeRoute(generatorResult, locals, this)); })
+ .then(locals => { route.set(path, createLoadThemeRoute(generatorResult, locals, this)) })

We would like to have this PR.

curbengh:
@deepnum 

What happened? Do you still want to re-attempt?
We would like to have this refactor.

curbengh:
Since this is refactor (not a bugfix), we'll re-visit after v4.

deepnum:
You can uses lebab to automatically do this

curbengh:
Thanks, will keep that in mind.

PR #3665, opened 2019-08-15
Solve "Out of memory" during generation by limiting the number of concurrent tasks

ppoffice:
## What does it do?
Allow users to set the number of files to be generated in parallel.
Users who are struggling with the "out of memory" problem (e.g., #3350 #3311) can limit Hexo's memory usage by `hexo g --concurrency <number>`.

## How to test
1. `hexo init` a new blog.
2. change the `hexo` dependency to `ppoffice/hexo#master` in `package.json`
3. `npm install`
4. Add some posts `git clone https://github.com/SukkaLab/hexo-many-posts.git source/_posts/tests `
5. `hexo g -c <number of concurrent tasks>`

### Environment

- **hexo**: https://github.com/ppoffice/hexo#master
- **theme**: landscape
- **hexo-cli**: 2.0.0
- **os**: Linux 5.0.0-21-generic linux x64
- **node**: 12.4.0
- **node_flags**: --max-old-space-size=8192

### Test 1: `hexo clean && hexo g`
- No concurrency limit
- 659 files generated in 24 s
- Memory usage: 500 ~ 660 MB
mem-graph

### Test 2: `hexo clean && hexo g -c 20`
- 20 concurrent tasks at most
- 659 files generated in 30 s
- Memory usage: 450 ~ 550 MB
mem-graph

### Test 3: `hexo clean && hexo g -c 10`
- 10 concurrent tasks at most
- 659 files generated in 34 s
- Memory usage: 480 ~ 550 MB
mem-graph

### Test 4: `hexo clean && hexo g -c 5`
- 5 concurrent tasks at most
- 659 files generated in 34 s
- Memory usage: 420 ~ 550 MB
mem-graph

## Pull request tasks

- [x] Add test cases for the changes.
- [x] Passed the CI test.

Issue #3311, opened 2018-10-26
Why this error happens?


dazkarieh:
xxximac:ian root# hexo clean && hexo d -g
INFO  Deleted database.
INFO  Deleted public folder.
INFO  Start processing
INFO  Files loaded in 5.14 s

<--- Last few GCs --->

[69389:0x104800600]   665523 ms: Scavenge 1292.1 (1438.7) -> 1283.2 (1439.2) MB, 9.6 / 0.0 ms  (average mu = 0.238, current mu = 0.159) allocation failure
[69389:0x104800600]   665583 ms: Scavenge 1296.6 (1439.7) -> 1288.0 (1441.2) MB, 8.0 / 0.0 ms  (average mu = 0.238, current mu = 0.159) allocation failure
[69389:0x104800600]   666470 ms: Mark-sweep 1300.0 (1441.2) -> 1285.9 (1439.2) MB, 856.1 / 0.0 ms  (average mu = 0.268, current mu = 0.297) allocation failure scavenge might not succeed

<--- JS stacktrace --->

==== JS stack trace =========================================

    0: ExitFrame [pc: 0x375315841bd]
    1: StubFrame [pc: 0x3753159213d]
Security context: 0xefd8c31e589 <JSObject>
    2: importAssets [0xefdebfe44f1] [/Users/dazkarieh/Public/ian/node_modules/hexo-prism-plugin/src/index.js:~132] [pc=0x37531ea148b](this=0xefdcd388b41 <Hexo map = 0xefd674efa19>,code=0xefd5c2fda89 <Very long string[229539]>,data=0xefda27cca59 <Object map = 0xefd462f7211>)
    3: tryCatcher [0xefdea5b5c99] [/Users/dazkarie...

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: node::Abort() [/usr/local/bin/node]
 2: node::OnFatalError(char const*, char const*) [/usr/local/bin/node]
 3: v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [/usr/local/bin/node]
 4: v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [/usr/local/bin/node]
 5: v8::internal::Heap::UpdateSurvivalStatistics(int) [/usr/local/bin/node]
[...]

Issue #1769, opened 2016-02-21
hexo generate slow with 400 posts


iMilnb:
```
$ hexo version
hexo: 3.1.1
os: NetBSD 7.0_STABLE netbsd x64
http_parser: 2.5.0
node: 4.1.1
v8: 4.5.103.33
uv: 1.7.4
zlib: 1.2.3
ares: 1.10.1-DEV
modules: 46
openssl: 1.0.2d
$ hexo clean
INFO  Deleted database.
INFO  Deleted public folder.
$ hexo g
[...]
INFO  1071 files generated in 3.68 min
# without cleaning
$ hexo g
INFO  Files loaded in 4.14 s
INFO  0 files generated in 4 min
```

Is this a normal generation time for 400 posts? seems awfully slow to me. Every change I make on a layout or a css takes ages to be generated. `hexo g -w` is barely unusable.

FWIW:
```
highlight:
  enable: true
  line_number: true
  auto_detect: false
  tab_replace:
```
Similar results with `enable: false`.

Theme is `light`.

leesei:
Did you tried `hexo server`?

iMilnb:
I did, it is indeed more usable and faster, but only for pre-visualization / debugging purposes

leesei:
I see. I was trying to suggest you to use it during development, so that `generate` is used sparingly.
What about not cleaning before every generate? It should be able to pick up changes in posts (a theme change will need to be cleaned though).

iMilnb:
Sure, I don't run `clean` everytime, just for testing, but TBH, even `hexo g` without a previous `clean` takes ages...
```
$ hexo g
INFO  Files loaded in 4.14 s
INFO  0 files generated in 4 min
```

iMilnb:
For the sake of it, I've tried to `hexo g` on a `tmpfs` mounted directory, and the result is pretty much the same. Meaning it's not about filesystem access but really processing time.

Xuanwo:
I don't think it's normal.
As in my blog: https://travis-ci.org/Xuanwo/xuanwo.github.io/builds/111696580 ,

```
INFO  Files loaded in 4.46 s
INFO  718 files generated in 22 s
```

The curve of time/files is strange from 22/718(about 0.03) to 240/1071(about 0.224).

Maybe you can try set `linenumber` to `false` to disable highlight.js totally.

iMilnb:
Well:
```
highlight:
  enable: false
  line_number: false
  auto_detect: false
[...]

Issue #1763, opened 2016-02-20
Memory usage when generating large site


leesei:
https://github.com/hexojs/hexo/issues/1735
https://github.com/hexojs/hexo/issues/1529

I'll try to recreate the environment first.

leesei:
Generate stub files with [leesei/hexo-genstubs](https://github.com/leesei/hexo-genstubs).

With `genstubs.js -N 5000 -P 100`:
```
229M    source/_posts
128M    stub-posts.zip
```

This is enough to generate OOM on my Arch Linux with 8G RAM.
The Hexo site is generated by `hexo init`.

```
hexo-cli@0.1.9
hexo@3.1.1
hexo-generator-archive@0.1.4
hexo-generator-category@0.1.3
hexo-generator-index@0.2.0
hexo-generator-tag@0.2.0
hexo-renderer-ejs@0.1.1
hexo-renderer-marked@0.2.10
hexo-renderer-stylus@0.3.1
hexo-server@0.1.3
```

```sh
$ hexo generate

<--- Last few GCs --->

    7958 ms: Scavenge 1408.7 (1457.0) -> 1408.7 (1457.0) MB, 17.1 / 0 ms (+ 33.0 ms in 1 steps since last GC) [allocation failure] [incremental marking delaying mark-sweep].
    8984 ms: Mark-sweep 1408.7 (1457.0) -> 1405.9 (1454.0) MB, 1026.3 / 0 ms (+ 967.1 ms in 2405 steps since start of marking, biggest step 33.0 ms) [last resort gc].
   10009 ms: Mark-sweep 1405.9 (1454.0) -> 1405.8 (1457.0) MB, 1024.3 / 0 ms [last resort gc].

FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - process out of memory
```

tommy351:
Thanks for the test. :100: :+1:

xieguigang:
My site have over 8000 pages, and encounter this problem too:
```
$ hexo generate
INFO  Start processing
FATAL Something's wrong. Maybe you can find the solution here: http://hexo.io/docs/troubleshooting.html
Template render error: (unknown path) [Line 4, Column 929]
  expected variable end
    at Object.exports.prettifyError (D:\gcmodeller-hexo\docs\node_modules\nunjucks\src\lib.js:34:15)
    at Obj.extend.render (D:\gcmodeller-hexo\docs\node_modules\nunjucks\src\environment.js:468:27)
[...]

PR #3756, opened 2019-10-09
perf(cache): Improve cache, Reduce Memory Usages

dailyrandomphoto:
## What does it do?

Hexo uses a lot of memory to make a huge site.
This PR will reduce memory usages by modifying the usage of the unnecessary cache.
It can help with the "out of memory" related issues. (e.g., #3350 #3311 #1769 #1763 #1735 #1529)

There are two kinds of caches that hexo using
-  Cache Of Rendered HTML: introduced in https://github.com/hexojs/hexo/commit/e8e45ed0f379f975147a214fbf52b2c28a6c806a#diff-b9bb6fa7bb069bab9948daba90c6c3b2
- Fragment Cache: introduced in https://github.com/hexojs/hexo/pull/637 

I've found something that can be improved a little bit to reduce memory usages.
 
### Before

Mode | Use Cache Of Rendered HTML | Use Fragment Cache
--- | --- | ---
generate | true | true
generate + watch | false | false
server(watch) | false | false

1. In `generate` mode, **Use Cache Of Rendered HTML** should be `false`, since each route just be called once. The unnecessary cache uses a lot of memory.
On the other hand, in `server` mode, **Use Cache Of Rendered HTML** should be `true`, since each route will be called several times. After enabling the cache, the route will respond more quickly from second requests.

2. In `generate + watch` mode, generating a huge site will be slower than `generate` mode since  **Use Fragment Cache** is disabled. **Use Fragment Cache** should always be enabled to speed up generating.

### After
This PR will change the usages of caches to 

Mode | Use Cache Of Rendered HTML | Use Fragment Cache
--- | --- | ---
generate | false | true
generate + watch | false | true
server(watch) | true | true

## How to test

```sh
git clone -b improve-cache https://github.com/dailyrandomphoto/hexo.git
cd hexo
npm install
npm test
```

### Before
```sh
git clone https://github.com/dailyrandomphoto/hexo-huge-site-test.git
cd hexo-huge-site-test
npm install
npm run g
```

**Environment & Result**
- hexo: hexojs/hexo
- theme: landscape
- concurrency: 1 (Introduced in https://github.com/hexojs/hexo/pull/3665)
- 3059 files generated in 2.13 min
- Memory usage: 944MB
- Rendered HTML strings hold a lot of memory. see screenshots. 

### After
change the hexo dependency to dailyrandomphoto/hexo#improve-cache in package.json
```sh
npm install dailyrandomphoto/hexo#improve-cache
npm run g
```

**Environment & Result**
- hexo: dailyrandomphoto/hexo#improve-cache
- theme: landscape
- concurrency: 1 (Introduced in https://github.com/hexojs/hexo/pull/3665)
- 3059 files generated in 2.03 min
- Memory usage: 445MB
- Rendered HTML strings were garbage collected. see screenshots. 

## Pull request tasks

- [x] Passed the CI test.


SukkaW:
LGTM. I will run a benchmark on this.

SukkaW:
With fragment cache enabled:

|:---:|:---:|:---:|:---:|
| Hexo (master) | 10s | 9.39s | 9.4s |
| This PR | 11s | 9.33s | 9.38s |

With fragment cache disabled:

|:---:|:---:|:---:|:---:|
| Hexo (master) | 15.5s | 15s | 15s |
| This PR | 16s | 15s | 14.5s |

I can see this PR has its generation speed more or less as same as the current master branch. I am wonder why @tommy351 implement [`Cache Of Rendered HTML`](https://github.com/hexojs/hexo/commit/e8e45ed0f379f975147a214fbf52b2c28a6c806a#diff-b9bb6fa7bb069bab9948daba90c6c3b2) at the beginning.

curbengh:
Fragment cache reminds me of this issue https://github.com/hexojs/hexo/issues/2503, specifically this [comment](https://github.com/hexojs/hexo/issues/2503#issuecomment-364626648) (I highlighted the last line):

> Can you confirm whether you have a layout file that is using the {cache:true} option set on the partial that loads the open_graph tag? I had the same issue, and then realized I was caching the partial that was loading my header which then caused this issue. Take a look, that might also have been your issue! **hexo s must not respect the cache:true?**

By enabling Fragment Cache for every mode, I believe this PR is also an improvement to the `Different result between hexo g and hexo s` of [roadmap](https://github.com/hexojs/hexo/issues/2492). If Fragment Cache was the sole cause, this PR can even fully fix it; though I suspect there might be other cause.

lib/plugins/helper/fragment_cache.js
```diff
@@ -1,7 +1,12 @@
 'use strict';
 
 module.exports = ctx => {
-  const cache = {};
+  let cache = {};
+
+  ctx.on('generateBefore', function() {
```
curbengh: I think can use arrow function?

dailyrandomphoto:
**Fragment Cache**(introduced in #637) can improve performance a lot.
as @tommy351 said,
> The effect of fragment cache is remarkable. I used fragment_cache helper in sidebar to reduce database queries. It only took less than 10 seconds to generate 1000+ static files. The original elapsed time is 10 minutes, which means is at least 60x faster.

However, some partial which output is different between pages should turn it off.
e.g. 
- head: title, og tags, meta tags should diffrent
- toc
- etc.

Like these partial(widget) can turn it on:
- footer
- tag, tagcloud
- category
- archives
- etc.


I found that `landscape` theme set `true` for `'_partial/header'` when `config.relative_link` is false.
This `cache` option should be removed on L7. 
https://github.com/hexojs/hexo-theme-landscape/blob/master/layout/layout.ejs#L5-L17
```html
  <div id="container">
    <div id="wrap">
      <%- partial('_partial/header', null, {cache: !config.relative_link}) %>
      <div class="outer">
        <section id="main"><%- body %></section>
        <% if (theme.sidebar && theme.sidebar !== 'bottom'){ %>
          <%- partial('_partial/sidebar', null, {cache: !config.relative_link}) %>
        <% } %>
      </div>
      <%- partial('_partial/footer', null, {cache: !config.relative_link}) %>
    </div>
    <%- partial('_partial/mobile-nav', null, {cache: !config.relative_link}) %>
    <%- partial('_partial/after-footer') %>
```

PR #3782, opened 2019-10-21
feat: meta_generator helper

SukkaW:
## What does it do?

This is a continuation of #3669 

## How to test

```sh
git clone -b meta_generator https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/helper/index.js
+++ b/lib/plugins/helper/index.js
@@ -43,6 +43,8 @@ module.exports = ctx => {
   helper.register('list_tags', require('./list_tags'));
   helper.register('list_posts', require('./list_posts'));
 
+  helper.register('meta_generator', require('./meta_generator'));
+
   helper.register('open_graph', require('./open_graph'));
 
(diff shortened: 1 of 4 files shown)


curbengh:
What's the benefit of having this instead of `meta_generator: true|false`?

SukkaW:
@curbengh 
If a theme want to add `meta_generator` tag this helper will be helpful. With theme using this helper, the `meta_generator` filter can be disabled and speed up generation.

curbengh:
Fair enough.

Then,

https://github.com/hexojs/hexo/blob/3a3e101d80599b2976133af7478c5e0a5222f52e/lib/plugins/filter/after_render/meta_generator.js#L8

can be refactored to `const hexoGeneratorTag = metaGeneratorTag();` to de-duplicate.

SukkaW:
@curbengh Since it is only a simple string, I see no necessary replacing it with a function.

Issue #3774, opened 2019-10-17
post.path contains "(double quotation marks) will generate wrong URL


walfud:
## Check List

Please check followings before submitting a new feature request.

- [x] I have already read [Docs page](https://hexo.io/docs/)
- [x] I have already searched existing issues

## Others

If post file named: _js 里面的 "相等性" 你了解多少_, hexo will generate:

```html
<a class="post-title-link" href="/js/js 里面的 " 相等性"="" 你了解多少="" "="" itemprop="url">
            <span itemprop="name">....</span>
          </a>
```

`href` need to be encoded I think.

curbengh:
May I know where did you get the link? Archive? In a post?

curbengh:
Maybe related to https://github.com/hexojs/hexo/pull/3704, can you try in hexo v4?

walfud:
> May I know where did you get the link?
> 
> Which theme are you using?
> 
> I just tested the filename, it's encoded on archive page, so I need to know the source of the unencoded `<a>`.

![image](https://user-images.githubusercontent.com/9317897/67181838-538e4e80-f410-11e9-901f-41cbd9cf66b3.png)

walfud:
> Maybe related to #3704, can you try in hexo v4?

Hi, v4 works fine. Thank you for helping.

PR #3789, opened 2019-10-23
docs: add compatibility with Node.js

YoshinoriN:
## What does it do?

Currently, we have to update `ISSUE_TEMPLATE` and [site requirements section](https://hexo.io/docs/#Requirements) whenever we drop older Node.js support.

I suggest create compatibility document and that `ISSUE_TEMPLATE` and [site requirements section](https://hexo.io/docs/#Requirements) refers to it.

It may help users. Users will be able to know what version they need.

---

Engines option introduce [this commit](https://github.com/hexojs/hexo/commit/5a9a76efe9d1e59fe4509d0516bcc332a02f3948). And it's released `3.6.0`.

## How to test

Nothing

## Screenshots

Nothing

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/docs/nodejs-compatiblility.md
+++ b/docs/nodejs-compatiblility.md
@@ -1,0 +1,17 @@
+# Node.js compatibility
+
+Hexo runs on the [Node.js](https://nodejs.org). Sometimes we may drop compatibility with old Node.js, to use new features of Node.js and keep up to date of dependencies.
+
+## Compatibility with Node.js
+
+Following is a list of minimum required Node.js version in each version of Hexo.
+
+|Hexo version|Minimum required Node.js version|
+|---|---|
+|4.x|8.6|
+|3.6.x - 3.9.x|6.9|
+|0.0.1 - 3.5.x|-|
+
+## References
+
+> [Node.js release](https://github.com/nodejs/Release)


SukkaW:
IMHO, we can just mention it at nodejs installation guide and it will be fine.

curbengh:
> Currently, we have to update `ISSUE_TEMPLATE` and [site requirements section](https://hexo.io/docs/#Requirements) whenever we drop older Node.js support.

This PR doesn't help reducing maintenance, it merely shift to updating `nodejs-compatiblility.md` instead.

I still prefer to retain Node version in `ISSUE_TEMPLATE` to indicate that we don't support older Node (neither should we).

I would suggest to add `Status`,

``` md
We strongly recommend to upgrade to the latest Hexo version, whenever possible.

|Hexo version|Minimum required Node.js version|Status
|---|---|
|4.1+|8.10|Current
|4.0|8.6|
...
```

> IMHO, we can just mention it at nodejs installation guide of docs and it will be fine.

+1 on moving it to the website for visibility, but I prefer to have the table at the bottom section of the guide (while still retaining the `(Should be at least Node.js x.x...)`. The compatibility table should _only_ be applicable to users who couldn't upgrade to newer Node; it should be easy to find, but not the focus.

YoshinoriN:
> This PR doesn't help reducing maintenance, it merely shift to updating nodejs-compatiblility.md instead.

I think half is yes and half is no.

Currently, 2 repositories 10 files have to fix. So, I thought we can focus on only update `nodejs-compatiblility.md` when I submitted this PR.

> IMHO, we can just mention it at nodejs installation guide of docs and it will be fine.

Ok, I agree with you. Because, I was very hard to decide whether should I submit it to hexo repo or website repo. Finally, I submitted it to this repo for the reasons mentioned at first. But, as you say website is better than this repo for visibility.

> I would suggest to add Status,

What is the purpose for? IMHO if we publish like nightly or pre or beta version, yes it's needed. But, currently, we have not published it.

> I still prefer to retain Node version in ISSUE_TEMPLATE to indicate that we don't support older Node (neither should we).

Hmm... I think users may notice their hexo and Node.js compatibility when before submit a new issue if we add reference to it on ISSUE_TEMPLATE.

In any case, I agree with this PR move to website.

curbengh:
> What is the purpose for? IMHO if we publish like nightly or pre or beta version, yes it's needed. But, currently, we have not published it.

It's to emphasize we only support the current minimum Node version (and above). But if this Pr moves to the website instead, plus mentioning "We strongly recommend to upgrade...", then `Status` is not needed.

> I think users may notice their hexo and Node.js compatibility when before submit a new issue

I think having the current checklist gives stronger nudge to user to upgrade. Actually, I wanted to add another checklist "Hexo version is 4.0" to the ISSUE_TEMPLATE to minimize bug report like #3774, but maybe it's too harsh. To be frank, we're not going to release a bugfix to v3 anyway.

curbengh:
> In any case, I agree with this PR move to website.

Added https://github.com/hexojs/site/pull/1189/commits/0d6a1ceb3b330f5e6e67352789374b1da998953e

YoshinoriN:
> To be frank, we're not going to release a bugfix to v3 anyway.

No objections

PR #2631, opened 2017-07-03
feat: Ignore files from processing that are not required.

bhaskarmelkani:
### This is a proposal for the issue [#2632](https://github.com/hexojs/hexo/issues/2632)
Example _config.yml entry.
```yaml
ignore: ['**/themes/**/node_modules/**', '**/themes/**/node_modules', '**/themes/**/.git', '**/themes/**/.git/**']
```
### TODO:
- [x] Add test cases for the changes.
- [x] Passed the CI test.

Issue #2090, opened 2016-08-11
Error: EMFILE: too many open files (ulimit does not work on Windows 10)


8ctopotamus:
In the [Hexo troubleshooting guide](https://hexo.io/docs/troubleshooting.html) they say to try

```
$ ulimit -n 10000
```

**...but this does not work on Windows.**
## Environment Info

Node version

```
v5.6.0
```

My site `_config.yml`

```
# Hexo Configuration
# Hexo Configuration
## Docs: https://hexo.io/docs/configuration.html
## Source: https://github.com/hexojs/hexo/

# Site
title: Cool New Site
subtitle:
description:
author:  8ctopotamus
language:
timezone:

# URL
## If your site is put in a subdirectory, set url as 'http://yoursite.com/child' and root as '/child/'
url: http://yoursite.com
root: /
permalink: :year/:month/:day/:title/
permalink_defaults:

# Directory
source_dir: source
public_dir: public
tag_dir: tags
archive_dir: archives
category_dir: categories
code_dir: downloads/code
i18n_dir: :lang
skip_render:

# Writing
new_post_name: :title.md # File name of new posts
default_layout: post
titlecase: false # Transform title into titlecase
external_link: true # Open external links in new tab
filename_case: 0
render_drafts: false
post_asset_folder: false
relative_link: false
future: true
highlight:
  enable: true
  line_number: true
  auto_detect: false
  tab_replace:

# Category & Tag
default_category: uncategorized
category_map:
tag_map:

# Date / Time format
## Hexo uses Moment.js to parse and display date
## You can customize the date format as defined in
## http://momentjs.com/docs/#/displaying/format/
date_format: YYYY-MM-DD
time_format: HH:mm:ss

# Pagination
## Set per_page to 0 to disable pagination
per_page: 10
pagination_dir: page

# Extensions
## Plugins: https://hexo.io/plugins/
## Themes: https://hexo.io/themes/
theme: trio-theme

# Deployment
## Docs: https://hexo.io/docs/deployment.html
deploy:
  type:

```

My theme `_config.yml`

```
# Hexo Configuration
# Header
menu:
  VISION: /vision
  DESIGN: /design
  EXECUTION: /execution
  REALIZATION: /realization

rss: /atom.xml

# Content
excerpt_link: Read More

# Sidebar
sidebar: right
widgets:
- category
- tag
- tagcloud
- archive
- recent_posts

# display widgets at the bottom of index pages (pagination == 2)
index_widgets:
- category
[...]

Issue #1316, opened 2015-06-06
Exclude a folder from files to be checked/watched


diessica:
`hexo server` command was taking too much to execute in terminal, then I added the `--debug` flag and have found out that Hexo is actually looking for changes in some folders unnecessarily. I want to exclude them from the files to be checked when I `hexo server` - folders just like *node_modules*:

```bash
[...]
03:53:33.612 DEBUG Unchanged: themes/[...]/node_modules/[...]
03:53:33.612 DEBUG Unchanged: themes/[...]/node_modules/file[...]
03:53:33.612 DEBUG Unchanged: themes/[...]/node_modules/file3[...]
03:53:34.097 DEBUG Theme config loaded.
03:53:34.097 DEBUG Processed: _config.yml
[...]
```
**Is there a setting for ignoring folders?** I was thinking about `skip_render` or `exclude_generator`, but those didn't work. I guess the documentation wasn't that clear about their roles.

Thank you!

diessica:
**TL;DR**
Hexo is watching files I don't want it to watch. How can I ignore those folders in my blog folder?

Xuanwo:
https://github.com/hexojs/hexo/issues/1146#issuecomment-88380140

Will this be useful?

Cobertos:
Got here from Google, I think this feature was added in https://github.com/hexojs/hexo/pull/2631 from issue https://github.com/hexojs/hexo/issues/2632. This adds a global `ignore:` parameter to exclude folders from watching/processing, but I could only get it to work from the projects config and not the theme config (though you can still ignore theme files).

Adding the following code will stop it from watching/processing these files during generation or serving (from the themes folder). It uses minimatch globs.
```
ignore: ['**/themes/**/node_modules/**', '**/themes/**/node_modules', '**/themes/**/.git', '**/themes/**/.git/**']
```

If you're curious about the actual code, [check out this location in box.js](https://github.com/hexojs/hexo/blob/2ed17cd105768df379dad8bbbe4df30964fe8f2d/lib/box/index.js#L108)

PR #3797, opened 2019-10-24
Disable fs.watch() for dirs/files specified by "ignore" option

### LINKED ISSUE #2179: Feature Request: Allow turning off watch() for files inside node_modules
## Environment Info

Hexo version 3.2.2

Plugin version(`npm ls --depth 0`)

```
+-- hexo@3.2.2
+-- hexo-deployer-git@0.2.0
+-- hexo-generator-archive@0.1.4
+-- hexo-generator-category@0.1.3
+-- hexo-generator-feed@1.2.0
+-- hexo-generator-index@0.2.0
+-- hexo-generator-tag@0.2.0
+-- hexo-renderer-ejs@0.2.0
+-- hexo-renderer-marked@0.2.11
+-- hexo-renderer-stylus@0.3.1
+-- hexo-server@0.2.0
```
## Feature Request

Allow turning off watch() for files inside node_modules
## Use Case

Right now it's easy to get too many open files error. The main reason is node_modules often contributes a lot of files. While understanding sometimes this is useful when updating code, this is generally unnecessary. Users should be allowed to filter out certain directories to be watched.

seaoak:
# What does it do?

Disable `fs.watch()` for dirs/files which are specified by `ignore` option in `_config.yml`.

This patch is supplement to #2631.

Fix #2179, #1316, hexojs/hexo-server#45.
And also, maybe mitigate #2090, hexojs/hexo-server#21.

## How to test

```sh
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/box/index.js
+++ b/lib/box/index.js
@@ -30,6 +30,10 @@ function Box(ctx, base, options) {
   this.Cache = ctx.model('Cache');
   this.File = this._createFileClass();
   this.ignore = ctx.config.ignore;
+  if (ctx.config.ignore) {
+    const targets = Array.isArray(ctx.config.ignore) ? ctx.config.ignore : [ctx.config.ignore];
+    this.options.ignored = (this.options.ignored || []).concat(targets.map(s => toRegExp(ctx, s)).filter(x => x));
+  }
 }
 
 require('util').inherits(Box, EventEmitter);
@@ -50,6 +54,20 @@ function getHash(path) {
       })
       .on('error', reject);
   });
+}
+
+function toRegExp(ctx, arg) {
+  if (!arg) return null;
+  if (typeof arg !== 'string') {
+    ctx.log.warn('A value of "ignore:" section in "_config.yml" is not invalid (not a string)');
+    return null;
+  }
+  const result = micromatch.makeRe(arg);
+  if (!result) {
+    ctx.log.warn('A value of "ignore:" section in "_config.yml" can not be converted to RegExp:' + arg);
+    return null;
+  }
+  return result;
 }
 
 Box.prototype._createFileClass = function() {
--- a/test/scripts/box/box.js
+++ b/test/scripts/box/box.js
@@ -437,6 +437,78 @@ describe('Box', () => {
     });
   });
 
+  it('watch() - update with simple "ignore" option', () => {
+    const box = newBox('test', {ignore: '**/ignore_me'});
+    const path1 = 'a.txt';
+    const path2 = 'b.txt';
+    const src1 = pathFn.join(box.base, path1);
+    const src2 = pathFn.join(box.base, 'ignore_me', path2);
+    const cacheId1 = 'test/' + path1;
+    const cacheId2 = 'test/ignore_me/' + path2;
+    const Cache = box.Cache;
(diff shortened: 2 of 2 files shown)


seaoak:
SAMPLE:
To exclude `node_modules` in theme directories, add lines into your `_config.yml` as follow:

```yaml
ignore:
  - /**/node_modules
```

My setting is:

```yaml
ignore:
  - /**/node_modules
  - /**/.git
  - /**/bower_components
```

seaoak:
NOTE:
To prohibit Hexo from watching dirs/files,
the option `ignored` should be passed to `chokidar.watch()`.
(Not `ignore`)
This option can be an array of RegExp.

The option `box.options.ignore` (which is set by this patch) is passed through `hexo-fs.watch()`.

curbengh:
`ignore:` option is currently [not documented](https://hexo.io/docs/configuration). In my previous testing, [`exclude:`](https://hexo.io/docs/configuration#Include-Exclude-Files-or-Folders) option doesn't work on theme folder.

Does `ignore:` work on all folders? specifically, can it cover the use case of `exclude`? If that's the case, we need to add some notes in the docs:

1. In the `Include/Exclude` section of the doc, mention "To exclude files/folders in the `themes`, use `ignore:` option".
2. In the `ignore:` option (does not exist yet), mention "While this option applies to all subdirectories of your hexo working folder, we recommend to use this option for ignoring files/folders in `themes` only."

---

> maybe some hexo plugins ot themes might require external source from node_modules
I believe plugins are still able to access the folder through Node api, even when it's ignored by Hexo.

> If we can ignore node_modules and .gitby default we might significantly reduce generation time.
+1 on ignoring them by default. I would recommend array appending them (instead of object assign) so that the defaults don't get replaced. Suggest to implement in a separate PR for easier testing.

seaoak:
My understanding that three options (`exclude:` and `include:` and `skip_render:`) affect to handle dirs/files under `source/`.
On the other hand, `ignore:` option is not restricted within there. That is, it can apply to dirs under `themes/`.

This behavior is not written in the "Configuration" document.
https://hexo.io/docs/configuration

This is TODO.

Are there already-existing Issue/PR about documentation?

seaoak:
I agree with adding `.node_modules` and `.git` into the default config of Hexo.
But it is necessary to discuss, test for existing themes and update document.
So I think I will not include it in this PR.

curbengh:
> On the other hand, `ignore:` option is not restricted within there. That is, it can apply to dirs under `themes/`.

I just tested the option, it does apply to every folder including `source/`. So, `ignore` does cover use case of `exclude`. While `exclude` seems extraneous, having smaller scope (i.e. `source/`) can be useful, for example if a folder/file has same name in theme but user doesn't want to exclude the one in theme.

> Are there already-existing Issue/PR about documentation?

https://github.com/hexojs/site/issues/1095

> So I think I will not include it in this PR.

Agree, we'll need to test globbing for that, which is out of scope for this PR.

SukkaW:
FYI: https://github.com/hexojs/hexo/commit/4eaec8ac2219eacddb1222a1374543738e983c8f

`node_modules` under themes dir is already ignored by default.

Issue #3753, opened 2019-10-08
Replace lodash with native API


SukkaW:
## Feature Request

The idea came from https://github.com/hexojs/site/pull/1139#discussion_r332336052 & #3677.

There is [an article](https://codeburst.io/why-you-shouldnt-use-lodash-anymore-and-use-pure-javascript-instead-c397df51a66) suggests we don't need lodash anymore.

According to benchmark below we can see es6 performance more or less as same as Lodash, in term of CPU, Memory or Handling time, at least in browser.

- https://dustinpfister.github.io/2017/11/20/lodash_foreach/
- https://hackernoon.com/es6-lodash-native-javascript-nodejs-example-tutorial-map-function-functional-programming-performance-speed-69645e613049
- https://www.reddit.com/r/javascript/comments/8j2hmq/performance_lodash_vs_es6_map/

There are also some benchmark of lodash vs es6 and es6 is definitely much faster in browser:

- https://blog.bitsrc.io/you-dont-need-lodash-or-how-i-started-loving-javascript-functions-3f45791fa6cd
- https://codeburst.io/why-you-shouldnt-use-lodash-anymore-and-use-pure-javascript-instead-c397df51a66

IMHO, we might consider replace at least some lodash with native js.

A guide about how to replace lodash with es6: https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore

A list of files that requires lodash:

**hexo**

- [x] [lib/plugins/console/deploy.js](https://github.com/hexojs/hexo/blob/master/lib/plugins/console/deploy.js)
- [x] [lib/theme/view.js](https://github.com/hexojs/hexo/blob/master/lib/theme/view.js)
- [x] [lib/plugins/filter/template_locals/i18n.js](https://github.com/hexojs/hexo/blob/master/lib/plugins/filter/template_locals/i18n.js)
- [x] [lib/theme/index.js](https://github.com/hexojs/hexo/blob/master/lib/theme/index.js)
- [x] [lib/hexo/load_config.js](https://github.com/hexojs/hexo/blob/master/lib/hexo/load_config.js)
- [x] [lib/plugins/filter/post_permalink.js](https://github.com/hexojs/hexo/blob/master/lib/plugins/filter/post_permalink.js)
- [x] [lib/plugins/helper/toc.js](https://github.com/hexojs/hexo/blob/master/lib/plugins/helper/toc.js)
[...]

Issue #2164, opened 2016-09-15
Major performance issues


pronvit:
Trying to investigate why Hexo is so slow, I found an interesting thing. Hexo always renders ALL files, even if no files are changed, and then rendering result for unchanged files is discarded.

Moreover, when rendering each file, it loads ejs templates and partials from the theme again and binds all the helpers, i.e. no template caching exists.

For example, for my not very bid site these two problems together lead to various ejs templates being processed 4500+ times and helpers being bound (https://github.com/hexojs/hexo/blob/fc4ca635b207fc2b223c7efab2b976af1a389b07/lib/theme/view.js#L78) 217000+ times (!) - always, even if no files are changed.

Any comments on this and plans to fix?

leesei:
Thanks for your effort in pointing out the bottlenecks.
Indeed Hexo's rendering pipeline can be improved. Yet without **documentation** of how it works it's hard to have an overall optimization.

That said, Hexo do have [fragment_cache](https://hexo.io/docs/helpers.html#fragment-cache). But some themes are not using it properly.

SukkaW:
For a layout for article page, there might be only one partial but it will be used to render a lot of posts. Those posts have different `locals`. The layout itself may not change at all, but your post content or your config, which will affect `locals`, might change.

In order to use hexo's built-in helpers in the theme layout, all those helpers have to bind on each `locals`.

stevenjoezhang:
I will close this issue because the cache is controlled by the Hexo theme. The theme developer decides which rendered content is cached and reused. For example, if every page has the same footer element, then the footer should be rendered faster using `fragment_cache`.

PR #3807, opened 2019-10-27
refactor: drop lodash for lib/theme

SukkaW:
## What does it do?

This PR is a part of #3753

## How to test

```sh
git clone -b drop-lodash-lib-theme https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/theme/index.js
+++ b/lib/theme/index.js
@@ -5,7 +5,6 @@ const util = require('util');
 const Box = require('../box');
 const View = require('./view');
 const I18n = require('hexo-i18n');
-const _ = require('lodash');
 
 function Theme(ctx) {
   Reflect.apply(Box, this, [ctx, ctx.theme_dir]);
@@ -29,7 +28,7 @@ function Theme(ctx) {
 
   this.i18n = new I18n({
     // languages: Array.from(new Set(languages.filter(Boolean)))
-    languages: _(languages).compact().uniq().value()
+    languages: [...new Set(languages.filter(Boolean))]
   });
 
   const viewFn = function(path, data) {
--- a/lib/theme/index.js
+++ b/lib/theme/index.js
@@ -5,7 +5,6 @@ const util = require('util');
 const Box = require('../box');
 const View = require('./view');
 const I18n = require('hexo-i18n');
-const _ = require('lodash');
 
 function Theme(ctx) {
   Reflect.apply(Box, this, [ctx, ctx.theme_dir]);
@@ -28,7 +27,7 @@ function Theme(ctx) {
   languages.push('default');
 
   this.i18n = new I18n({
-    languages: _(languages).compact().uniq().value()
+    languages: [...new Set(languages.filter(Boolean))]
   });
 
   const viewFn = function(path, data) {
--- a/lib/theme/index.js
+++ b/lib/theme/index.js
@@ -5,6 +5,7 @@ const util = require('util');
 const Box = require('../box');
 const View = require('./view');
 const I18n = require('hexo-i18n');
+const _ = require('lodash');
 
 function Theme(ctx) {
   Reflect.apply(Box, this, [ctx, ctx.theme_dir]);
@@ -27,7 +28,8 @@ function Theme(ctx) {
   languages.push('default');
 
   this.i18n = new I18n({
-    languages: Array.from(new Set(languages.filter(Boolean)))
+    // languages: Array.from(new Set(languages.filter(Boolean)))
+    languages: _(languages).compact().uniq().value()
   });
 
   const viewFn = function(path, data) {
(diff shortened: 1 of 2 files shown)


SukkaW:
According to benchmark performed by Travis CI, this PR makes generation speed 3x slower.

lib/theme/view.js
```diff
@@ -61,7 +67,9 @@ View.prototype.renderSync = function(options = {}) {
 };
 
 View.prototype._buildLocals = function(locals) {
-  return assignIn({}, locals, omit(this.data, ['layout', '_content']), {
+  // eslint-disable-next-line no-unused-vars
```
curbengh: can use the following function (which re-implement lodash.omit) to avoid unused vars ([source](https://github.com/Semantic-Org/Semantic-UI-React/pull/860)),

``` js
Object.keys(props).reduce((acc, prop) => {
    if (handledProps.indexOf(prop) === -1) acc[prop] = props[prop]
    return acc
  }, {})
```

lib/theme/view.js
```diff
@@ -1,11 +1,17 @@
...
 const { dirname, extname, join } = require('path');
-const assignIn = require('lodash/assignIn');
-const omit = require('lodash/omit');
 const yfm = require('hexo-front-matter');
 const Promise = require('bluebird');
 
+const assignIn = (target, ...sources) => {
+  let source = [];
+  sources.forEach(src => {
+    source = source.concat([src, Object.getPrototypeOf(src)]);
```
curbengh: would `Array.push()` resolve the regression?

SukkaW: Maybe. According to [this post](https://dev.to/uilicious/javascript-array-push-is-945x-faster-than-array-concat-1oki) `push()` is 945x faster than `concat()`.

I might try to re-implement `agginIn()` using `push()`.

curbengh: Looks like to https://github.com/hexojs/hexo/pull/3807/commits/f093749cde8e6973713547678662e15519c3c7e5 resolved the regression.

lib/theme/index.js
```diff
@@ -28,7 +27,7 @@ function Theme(ctx) {
   languages.push('default');
 
   this.i18n = new I18n({
-    languages: _(languages).compact().uniq().value()
+    languages: [...new Set(languages.filter(Boolean))]
```
curbengh: what about `Array.from(new Set(languages.filter(Boolean)))`?

SukkaW: `value()` is a lodash method, which is used to get value from lodash wrapped prototype chain.

SukkaW: I have make a benchmark at jsPerf: https://jsperf.com/lodash-uniq-vs-javascript-set/1

It seems the diffrence between `Array.from` & spread syntax is negligible, at least in browser.

curbengh: I see.

The benchmark showed the spread syntax is the source of regression. I noticed you reverted the last commit; to confirm the source?

curbengh: > It seems the diffrence between Array.from & spread syntax is negligible, at least in browser.

Similar result in travis too. Can you try revert back to lodash just for this line? to test whether `Set()` is the culprit.

curbengh: Found another alternative is `languages.filter((v, i, a) => a.indexOf(v) === i)` ([source](https://blog.bitsrc.io/you-dont-need-lodash-or-how-i-started-loving-javascript-functions-3f45791fa6cd)). It's <10% slower in Chrome while faster in FF,

SukkaW: I have revert this line back to lodash and found out it is not the culprit. So the `view.js` is the smoking gun.

During the generation process, the `view.js` will be called every time when a page is rendered (#2164). We might need tohandle it carefully.

curbengh: But that doesn't explain why lodash is faster in `view.js`.

SukkaW: @curbengh I have done through some investigation about performance of `Object.assign`. It seems that `Object.assign` will meet performance issue when facing large object. Even Node.js itself is still using the deprecated `util._extend`, because it is still a lot more faster than `Object.assign`.

SukkaW: A bit more interesting... Even if I replace Object.assign with lodash.assign, the impact is still there. So maybe the problem is at `Object.getPrototypeOf`..

curbengh: That seems like the last suspect now. what about `return Object.assign({}, locals, locals.prototype, data, {` ? ([source](https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_extend))

SukkaW: @curbengh In fact I have tried `local.prototype` at very beginning but it won't pass the test.

PR #3813, opened 2019-10-27
refactor: drop lodash for new-post-path filter

SukkaW:
## What does it do?

This PR is part of #3753 

## How to test

```sh
git clone -b drop-lodash-for-new-post-path https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/filter/new_post_path.js
+++ b/lib/plugins/filter/new_post_path.js
@@ -2,7 +2,6 @@
 
 const { join, extname } = require('path');
 const moment = require('moment');
-const defaults = require('lodash/defaults');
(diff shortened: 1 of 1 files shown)


SukkaW:
@YoshinoriN 

Maybe `Object.assign({}, permalinkDefaults, filenameData)` should be used? Then `permalinkDefaults` will become a `source` and `null` will be ignored.

YoshinoriN:
> Maybe Object.assign({}, permalinkDefaults, filenameData) should be used?

I think it's better than set default value :)

PR #3845, opened 2019-11-08
refactor: drop lodash for lib/hexo/index.js

SukkaW:
## What does it do?

This PR is part of #3753 

This PR only replace `cloneDeep` and `debounce` with native while `lodash` itself retains exposed to global.

## How to test

```sh
git clone -b drop-lodash-for-index https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Screenshots

The performance is more or less the same after this PR.

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/hexo/index.js
+++ b/lib/hexo/index.js
@@ -28,7 +28,7 @@ const resolve = require('resolve');
 const full_url_for = require('../plugins/helper/full_url_for');
 const { inherits } = require('util');
 
-const { cloneDeep, debounce } = _;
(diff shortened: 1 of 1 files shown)


curbengh:
Please verify the benchmark (rebase first) before merging.

SukkaW:
@curbengh Last time I check, the benchmark seems no difference. I will trigger a rebuild to see if there is any regression.

PR #3865, opened 2019-11-17
Exclude some languages in code highlighting

corvofeng:
Signed-off-by: corvofeng <corvofeng@gmail.com>

## What does it do?

Only wrap some languages' code with `<pre>` and `<code>`, and do no render all tags(span, and br) in content, so that user could process those languages ​​freely.

## How to test

```sh
git clone -b exclude_highlight_lang https://github.com/corvofeng/hexo.git
cd hexo
npm install
npm test
```

## How to use?

`config.yaml`:

```yaml
highlight:
  enable: true
  line_number: true
  auto_detect: false
  exclude_languages:
    - flowchartjs-code
```

Blog code:

\`\`\`flowchartjs-code
st=>start: Start:>http://www.bing.com[blank]
e=>end:>http://www.bing.com
op1=>operation: My Operation
sub1=>subroutine: My Subroutine
cond=>condition: Yes
or No?:>http://www.google.com
io=>inputoutput: catch something...
para=>parallel: parallel tasks

st->op1->cond
cond(yes)->io->e
cond(no)->para
para(path1, bottom)->sub1(right)->op1
para(path2, top)->op1
\`\`\`

You will get:

```html
<code class="highlight flowchartjs-code">st=&gt;start: Start:&gt;http://www.bing.com[blank]
e=&gt;end:&gt;http://www.bing.com
op1=&gt;operation: My Operation
sub1=&gt;subroutine: My Subroutine
cond=&gt;condition: Yes
or No?:&gt;http://www.google.com
io=&gt;inputoutput: catch something...
para=&gt;parallel: parallel tasks

st-&gt;op1-&gt;cond
cond(yes)-&gt;io-&gt;e
cond(no)-&gt;para
para(path1, bottom)-&gt;sub1(right)-&gt;op1
para(path2, top)-&gt;op1</code>
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/filter/before_post_render/backtick_code_block.js
+++ b/lib/plugins/filter/before_post_render/backtick_code_block.js
@@ -89,6 +89,13 @@ function backtickCodeBlock(data) {
         }
       }
 
+      if (Array.isArray(hljsCfg.exclude_languages) && hljsCfg.exclude_languages.includes(options.lang)) {
+        // Only wrap with <pre><code class="lang"></code></pre>
+        options.wrap = false;
+        options.gutter = false;
+        options.autoDetect = false;
+      }
+
       content = highlight(content, options);
     }
 
--- a/test/scripts/filters/backtick_code_block.js
+++ b/test/scripts/filters/backtick_code_block.js
@@ -241,6 +241,26 @@ describe('Backtick code block', () => {
       data.content.should.eql('<hexoPostRenderCodeBlock>' + expected + '</hexoPostRenderCodeBlock>');
     });
 
+    it('only wrap with pre and code', () => {
+      hexo.config.highlight.exclude_languages = ['js'];
+      hexo.config.highlight.hljs = true;
+      const data = {
+        content: [
+          '``` js',
+          code,
+          '```'
+        ].join('\n')
+      };
+      const expected = highlight(code, {
+        lang: 'js',
+        gutter: false,
+        hljs: true,
+        wrap: false
+      });
+      codeBlock(data);
+      data.content.should.eql('<hexoPostRenderCodeBlock>' + expected + '</hexoPostRenderCodeBlock>');
+    });
+
     it('line number false, don`t care first_line_number inilne', () => {
       hexo.config.highlight.line_number = false;
       hexo.config.highlight.first_line_number = 'inilne';
--- a/lib/hexo/default_config.js
+++ b/lib/hexo/default_config.js
@@ -46,6 +46,7 @@ module.exports = {
     line_number: true,
     tab_replace: '',
     wrap: true,
+    exclude_languages: [],
     hljs: false
   },
   prismjs: {


lib/plugins/filter/before_post_render/backtick_code_block.js
```diff
@@ -48,6 +49,14 @@ function backtickCodeBlock(data) {
           }
         }
       }
+
+      if (exclude_languages.includes(args)) {
+        // Only wrap with <pre><code class="lang"></code></pre>
+        options.wrap = false;
+        options.hljs = true;
```
curbengh: I haven't test this PR, my understanding is that, while enabling `hljs` also enables `wrap`, it is also possible to disable both at the same time.

test/scripts/filters/backtick_code_block.js
```diff
@@ -195,6 +195,27 @@ describe('Backtick code block', () => {
...
+    const expected = highlight(code, {
+      lang: 'js',
+      gutter: false,
+      hljs: true,
+      wrap: false
+    });
+
+    codeBlock(data);
+    data.content.should.eql('<escape>' + expected + '</escape>');
+  });
```
curbengh: need to reset `hexo.config.highlight.exclude_languages` at the end of this test.

lib/plugins/filter/before_post_render/backtick_code_block.js
```diff
@@ -12,6 +12,7 @@ function backtickCodeBlock(data) {
   if (!config.enable) return;
   data.content = data.content.replace(rBacktick, ($0, start, $2, _args, content, end) => {
     const args = _args.split('=').shift();
+    const exclude_languages = config.exclude_languages || [];
```
curbengh: need to move this into L17-22 for consistency.
add a null checking `if (options.exclude_languages)` before `if (exclude_languages.includes(args)) {}`.

lib/plugins/filter/before_post_render/backtick_code_block.js
```diff
@@ -48,6 +49,14 @@ function backtickCodeBlock(data) {
           }
         }
       }
+
+      if (exclude_languages.includes(args)) {
```
curbengh: `args` may include both language and caption, should use `options.lang` instead.

curbengh:
Requires https://github.com/hexojs/hexo/pull/3827

lib/plugins/filter/before_post_render/backtick_code_block.js
curbengh: `exclude_languages` also need to be added to `lib/hexo/default_config.js`.
see https://github.com/hexojs/hexo/pull/3827 for example.

test/scripts/filters/backtick_code_block.js
corvofeng: Well, `beforeEach()` has made a deep clone from the default config, I don't think it is necessary to reset that value.

lib/plugins/filter/before_post_render/backtick_code_block.js
corvofeng: To be honest, I have read the code in [hexo-util](https://github.com/hexojs/hexo-util), but it seems that there have been some changes in `wrap` and `gutter` in  function `highlightUtil` [lib/highlight.js](https://github.com/hexojs/hexo-util/blob/master/lib/highlight.js):
```javascript
// newer version:

if (useHljs && !gutter) wrap = false;
if (gutter && !wrap) wrap = true; // arbitrate conflict ("gutter:true" takes priority over "wrap:false")

const before = useHljs ? `<pre><code class="${classNames}">` : '<pre>';
const after = useHljs ? '</code></pre>' : '</pre>';

if (!wrap) return `<pre><code class="${classNames}">${data.value}</code></pre>`;
```

```javascript
// older version
const before = useHljs ? `<pre><code class="hljs ${options.lang}">` : '<pre>';
const after = useHljs ? '</code></pre>' : '</pre>';

if (!wrap) return useHljs ? before + data.value + after : data.value;
```

 I also wanna make sure the generated code will have`<pre>` and `<code>`, not only a `<pre>`

lib/plugins/filter/before_post_render/backtick_code_block.js
corvofeng: Fixed.

lib/plugins/filter/before_post_render/backtick_code_block.js
corvofeng: Fixed.

lib/plugins/filter/before_post_render/backtick_code_block.js
curbengh: > I also wanna make sure the generated code will have`<pre>` and `<code>`, not only a `<pre>`

shouldn't `hljs: false` + `wrap: false` also generate `<pre><code>`? That was the purpose of https://github.com/hexojs/hexo-util/pull/112.

lib/plugins/filter/before_post_render/backtick_code_block.js
```diff
@@ -47,6 +48,14 @@ function backtickCodeBlock(data) {
...
           }
         }
+
+        if (options.exclude_languages && options.exclude_languages.includes(options.lang)) {
+          // Only wrap with <pre><code class="lang"></code></pre>
+          options.wrap = false;
+          options.hljs = true;
+          options.gutter = false;
+          options.autoDetect = false;
+        }
```
curbengh: needs to be refactored into

``` js
if (options.exclude_languages) {
  if (options.exclude_language.includes(options.lang)) {
    ...
  }
}
```

I'm aware that `exclude_languages:` defaults to `[]` as shown in `default_config.js`, but `options.exclude_languages.includes()` would error if the `exclude_languages`is null,

``` yml
_config.yml
highlight:
  exclude_languages:
```

`Object.assign()` only handle non-existent key, not null value.

lib/plugins/filter/before_post_render/backtick_code_block.js
corvofeng: Yes, after this PR [hexojs/hexo-util#112](https://github.com/hexojs/hexo-util/pull/112), the `hljs` is not needed anymore, but in the older version of that code, `hljs` is needed for `<code>`. Well, because there are two projects for `hexo` and `hexo-utils`, I think it is much better to make some compatibility codes in case the user mixed-use the new version hexo and old version hexo-utils.

lib/plugins/filter/before_post_render/backtick_code_block.js
corvofeng: Thanks for pointing this out, Fixed.

lib/plugins/filter/before_post_render/backtick_code_block.js
curbengh: But enabling `hljs` wouldn't work with most theme which doesn't use the newer `hljs-*` syntax though.

I also created [a PR](https://github.com/hexojs/hexo-util/pull/138) to decouple `hljs:` and `wrap:`, i.e. enabling `hljs` will not disable `wrap`.

> in case the user mixed-use the new version hexo and old version hexo-utils.

we actually prefer plugins/themes to support older hexo (if the required workaround is minimal), than supporting older plugins.

corvofeng: Updated.

lib/plugins/filter/before_post_render/backtick_code_block.js
```diff
@@ -52,8 +52,10 @@ function backtickCodeBlock(data) {
         if (options.exclude_languages) {
           if (options.exclude_languages.includes(options.lang)) {
             // Only wrap with <pre><code class="lang"></code></pre>
+            // If you can't get the proper output.
+            // Please refer to: https://github.com/hexojs/hexo-util/blob/master/lib/highlight.js
+            // or update hexo-util to the latest version.
```
curbengh: I forgot to mention that hexo-util is tightly integrated into hexo (#3888, #3829, #3728), as in they are always upgraded together.

Since this feature will only be implemented (I don't mind, but I don't know if maintainer will approve) in future version and not backported (thus, https://github.com/hexojs/hexo-util/pull/138 will be included), you don't have to worry about compatibility with older hexo-util. In fact, some hexo components may even error out if latest hexo-util is not used.

corvofeng: Well, I delete those comments and rebase with this repo's master. Do I need to compress the commits?

curbengh: that's fine, I can do that when merging.

curbengh:
I just started testing this,

> Only wrap some languages' code with `<pre>` and `<code>`, and do no render all tags(span, and br) in content, so that user could process those languages ​​freely.

By "render", you mean html escape?

I tried

``` js
if (options.exclude_languages) {
  if (options.exclude_languages.includes(options.lang)) {
    options.wrap = false;
<br>
    options.gutter = false;
    options.autoDetect = false;
  }
}
```

`<br>` is still escaped, even when `highlight.enable` is disabled.

Looks like `exclude_wrap: []` option is more accurate?

lib/plugins/filter/before_post_render/backtick_code_block.js
```diff
@@ -48,6 +49,15 @@ function backtickCodeBlock(data) {
             options.caption += `<a href="${match[3]}">${match[4] ? match[4] : 'link'}</a>`;
           }
         }
+
+        if (options.exclude_languages) {
+          if (options.exclude_languages.includes(options.lang)) {
```
curbengh: if `exclude_languages` is a string, `c` will match `cpp`. need string -> array conversion, but don't write back to options.exclude_languages, use

```js
let { exclude_languages } = options;
```

corvofeng:
> I just started testing this,
> 
> > Only wrap some languages' code with `<pre>` and `<code>`, and do no render all tags(span, and br) in content, so that user could process those languages ​​freely.
> 
> By "render", you mean html escape?
> 
> I tried
> 
> ```js
> if (options.exclude_languages) {
>   if (options.exclude_languages.includes(options.lang)) {
>     options.wrap = false;
> <br>
>     options.gutter = false;
>     options.autoDetect = false;
>   }
> }
> ```
> 
> `<br>` is still escaped, even when `highlight.enable` is disabled.
> 
> Looks like `exclude_wrap: []` option is more accurate?

Well, in my blogs, when I work with my latest changes and hexo-util@1.8.0, I could get the proper result which does not include `<br>`. Would you like to use `hexo clean` firstly, and then check the result.

Like this:

\`\`\`flowchartjs-code
st=>start: Start:>http://www.bing.com[blank]
e=>end:>http://www.bing.com
op1=>operation: My Operation
sub1=>subroutine: My Subroutine
cond=>condition: Yes
or No?:>http://www.google.com
io=>inputoutput: catch something...
para=>parallel: parallel tasks

st->op1->cond
cond(yes)->io->e
cond(no)->para
para(path1, bottom)->sub1(right)->op1
para(path2, top)->op1
\`\`\`

And set:

```yaml
highlight:
  enable: true
  line_number: true
  auto_detect: false
  exclude_languages:
    - flowchartjs-code
```

You will get:

```html
<code class="highlight flowchartjs-code">st=&gt;start: Start:&gt;http://www.bing.com[blank]
e=&gt;end:&gt;http://www.bing.com
op1=&gt;operation: My Operation
sub1=&gt;subroutine: My Subroutine
cond=&gt;condition: Yes
or No?:&gt;http://www.google.com
io=&gt;inputoutput: catch something...
para=&gt;parallel: parallel tasks

st-&gt;op1-&gt;cond
cond(yes)-&gt;io-&gt;e
cond(no)-&gt;para
para(path1, bottom)-&gt;sub1(right)-&gt;op1
para(path2, top)-&gt;op1</code>
```

lib/plugins/filter/before_post_render/backtick_code_block.js
corvofeng: Well, I update the code to make sure the exclude_languages will be an array, but it is only for a defensive-programming requirement. It will be better to add a schema checker for `hexo` config.

stevenjoezhang:
There will be other code highlighting engines added to Hexo: https://github.com/hexojs/hexo/pull/4119
I suggest you implement this feature using custom tags, e.g.
```
{% flowchartjscode %}
st=>start: Start:>http://www.bing.com[blank]
e=>end:>http://www.bing.com
{% endflowchartjscode %}
```
This looks more clear than
\`\`\`flowchartjs-code
st=>start: Start:>http://www.bing.com[blank]
e=>end:>http://www.bing.com
\`\`\`

corvofeng:
> There will be other code highlighting engines added to Hexo: #4119
> I suggest you implement this feature using custom tags, e.g.
> 
> ```
> {% flowchartjscode %}
> st=>start: Start:>http://www.bing.com[blank]
> e=>end:>http://www.bing.com
> {% endflowchartjscode %}
> ```
> 
> This looks more clear than
> ```flowchartjs-code
> st=>start: Start:>[http://www.bing.com[blank]](http://www.bing.com%5Bblank%5D)
> e=>end:>http://www.bing.com
> ```

Well, I have downloaded that branch and tested it, and it may somehow achieve the function I requested. But I have a question: why not let the hexo user decide when the code should be rendered(in the hexo building or in the browser loading).

SukkaW:
@stevenjoezhang 

Will, we could merge #4119 first, and then I will bring up the feature in another PR.

corvofeng:
@SukkaW @stevenjoezhang @curbengh 

I have rebased this PR based on the latest master branch, could anyone merge it or give me some advice? I don’t want to be forgotten after a long time, because I really need this feature.

botreview (approved):
LGTM

stevenjoezhang (approved):

SukkaW (approved):

SukkaW:
The failed test is fixed in https://github.com/hexojs/hexo/pull/4902, not caused by the PR itself. The change LGTM.

D-Sketon:
I try to exclude java in highlight like this:
highlight:
  enable: true
  line_number: true
  auto_detect: false
  tab_replace: ''
  wrap: true
  hljs: false
  exclude_languages:
    - java 
But I find that <span> tags still exist, is this expected behavior?🤔🤔

PR #3878, opened 2019-11-20
Enable to use "ignore:" option for individual files (fix #3873)

### LINKED ISSUE #3873: 'ignore:' option doesn't work on invidual file
1. Create more than one dummy files in `source/foo/` folder
2. Configure `ignore:`,

``` yml
_config.yml
ignore:
  - '**/foo/dummy1.txt'
```

### Expected behavior

`public/foo/dummy1.txt` should not exist

### Actual behavior

`public/foo/dummy1.txt` is there

### Workaround

Use `exclude:`, but it doesn't work in `themes/` folder.

---

`ignore` currently works on folder level only,

``` yml
_config.yml
ignore:
  - '**/foo'
```

Noticed this limitation while working on https://github.com/hexojs/hexo-generator-feed/pull/110.

I haven't dive into source code for this issue yet, @seaoak can you help?

seaoak:
## What does it do?

Fix issue #3873

When an individual (non-directory) file matches the glob in `ignore:` option,
it will be ignored (not process, not copy to `publish/` folder).

`fs.watch()` already ignores it by the PR #3797.
But Hexo still processes it.

This patch fixes this behavior.

## How to test

```sh
git clone -b feature/enable_ignore_option_for_individual_file https://github.com/seaoak/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/box/index.js
+++ b/lib/box/index.js
@@ -120,12 +120,17 @@ Box.prototype._readDir = function(base, fn, prefix = '') {
   const { ignore } = this;
 
   if (base && ignore && ignore.length && micromatch.isMatch(base, ignore)) {
-    return Promise.resolve('Ignoring dir.');
+    return Promise.resolve([]);
   }
 
   return fs.readdir(base).map(path => fs.stat(join(base, path)).then(stats => {
+    const fullpath = join(base, path);
     if (stats.isDirectory()) {
-      return this._readDir(join(base, path), fn, `${prefix + path}/`);
+      return this._readDir(fullpath, fn, `${prefix + path}/`);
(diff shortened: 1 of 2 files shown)


seaoak:
Sorry!  This patch has critical bug.
All test cases pass, but both `hexo generate` and `hexo server` throw an exception.
I'm investigating, so please wait reviewing this patch.

seaoak:
I finished to rewrite this patch (mainly adjust the type of return value of `Box.prototype._readDir()`).

Please review!

Issue #3897, opened 2019-12-01
Enhancement of fragment_cache


SukkaW:
## Check List

Please check followings before submitting a new feature request.

- [x] I have already read [Docs page](https://hexo.io/docs/)
- [x] I have already searched existing issues

## Feature Request

Fragment cache was introduced in #637. It stores the function's output (usually `partial()`'s) into a big Object and read value from it.

Before we could completely understand hexo's rendering & generation process, fragment cache was probably the best way to solve the issue like #2164 (binding locals again and again for thousands of times) and #2991 (renders everything and only do comparing when writing files to drive).

Let's looks into #1769 again. Before utilizing `fragment_cache()`, although files are loaded less than 5 seconds, but rendering process has lasted for 4 min and no files has changed and written to drive. This issue is definitely the result of #2164 - rendering everything but nothing generated. But after just enable fragment_cache for sidebar partial the rendering process drops from 4 min to 6.41s.

Users could have significant performance boost if theme developer could utilize `fragment_cache()` correctly, but currently most of the theme developers don't know how to use `fragment_cache()` correctly: #3071 (`fragment_cache` could be enabled directly in `partial()`) & #3738 (fragment cache shouldn't be enabled in a partial when its locals could changes).

Recently I have noticed https://github.com/ppoffice/hexo-theme-icarus/commit/958138f2140010e4953cbecb923945fce85b642c : every partial will be cached in fragment cache with `cacheId` as md5 of `locals`. The a partial can be found in the fragment cache if its `locals` has not changed since the first time call. In this way every partial could enable fragment cache without theme developers enables it manually.

And there is a problem: if a partial has its `locals` changed frequently, then fragment cache will consume memory very, very quickly. So if we are going to address this feature in hexo, we have to first refactoring fragment cache from a big Object to LRU.

We could name this feature as `enhance_cache` and enables by default.

cc @hexojs/core

SukkaW:
I suggest if `enhance_cache` is enabled, all `partial()` should be cached by `fragment_cache` unless `{ cache:false }` is set.

SukkaW:
I have bring up the implementation of this proposal in #3901 & #3092. 

In #3902, I add `enhance_cache` option (default: `false`) option.

When a theme developer has specified `{ cache: true }` or `{ cache: [cacheId] }`, the `fragment_cache` will be enabled no matter `enhance_cache` is enabled or not. For the case a theme developer want to cache this partial even the locals has changed, I believe this approach will be helpful.

[...]

PR #3901, opened 2019-12-02
[WIP] feat(cache): bring up lfu

SukkaW:
## What does it do?

This PR is a part of proposal #3897, bring up LFU cache for `fragment_cache()`.

## How to test

```sh
git clone -b fragment_cache_lru https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.


lib/plugins/helper/fragment_cache.js
```diff
@@ -1,16 +1,18 @@
 'use strict';
 
+const LRU = require('node-lfu-cache');
+
 module.exports = ctx => {
-  let cache = {};
+  const cache = new LRU(50);
```
SukkaW: I have run some test by using `Object.keys(cache).length` to know the size of current themes.

| Theme | Current cache size |
|:---:|:---:|
| [hexo-theme-next](https://github.com/theme-next/hexo-theme-next) | 18 |
| [hexo-theme-suka](https://github.com/SukkaW/hexo-theme-suka) | 18 |
| [hexo-theme-cake](https://github.com/jiangtj/hexo-theme-cake) | 17 |

> Those data are from one-language site. Multi-language site could have larger size of cache.

jiangtj: It is best to be exposed to the outside for configuration, If some theme has larger cache length, it can adjust it.

SukkaW: Most of users have no idea about the cache size of their themes.

jiangtj: > Most of users have no idea about the cache size of their themes.

But as a theme developer. If he knew clearly that he needed 50+, how to achieve it?

SukkaW: @jiangtj In fact he/she doesn't. It is better to cache as less as possible because larger size of cache means higher memory usage. It is not a wise idea to cache everything in the `fragment_cache`. That's how LFU works, by only caching the most usefuls.

PR #3926, opened 2019-12-10
fix(postPermalinkFilter): permalink_defaults should not overwrite values

dailyrandomphoto:
## What does it do?

`permalink_defaults` option does not work correctly from 4.1.0.
According https://hexo.io/docs/permalinks, set `_config.yml` as
```yaml
permalink: :lang/:title/
permalink_defaults:
  lang: en
```

```
$ hexo new "Hello World" --lang tw
```
and the URL will be:
```
# expected is
http://localhost:4000/tw/hello-world/

# but actual is 
http://localhost:4000/en/hello-world/
```

Compare v4.0.0 and v4.1.0,
I found the related change is https://github.com/hexojs/hexo/commit/e0f2478ef431cccb1d30d92d2d0c47d6f195675b#diff-dbb03cb234189f8e709e7c5f9921aa67L45  (#3790) 
```diff
- return permalink.stringify(defaults(meta, config.permalink_defaults));
+ return permalink.stringify(Object.assign(meta, config.permalink_defaults));
```

If change to
```js
Object.assign([], config.permalink_defaults, meta)
```
it causes "Maximum call stack size exceeded" error, so use for loop to copy values.

## How to test

```sh
git clone -b fix-permalinkDefaults https://github.com/dailyrandomphoto/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/filter/post_permalink.js
+++ b/lib/plugins/filter/post_permalink.js
@@ -6,18 +6,19 @@ let permalink;
 
 function postPermalinkFilter(data) {
   const { config } = this;
+  const { id, _id, slug, title, date } = data;
   const meta = {
-    id: data.id || data._id,
-    title: data.slug,
-    name: typeof data.slug === 'string' ? basename(data.slug) : '',
-    post_title: slugize(data.title, {transform: 1}),
-    year: data.date.format('YYYY'),
-    month: data.date.format('MM'),
-    day: data.date.format('DD'),
-    hour: data.date.format('HH'),
-    minute: data.date.format('mm'),
-    i_month: data.date.format('M'),
-    i_day: data.date.format('D')
+    id: id || _id,
+    title: slug,
+    name: typeof slug === 'string' ? basename(slug) : '',
+    post_title: slugize(title, {transform: 1}),
+    year: date.format('YYYY'),
+    month: date.format('MM'),
+    day: date.format('DD'),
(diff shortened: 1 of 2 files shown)


lib/plugins/filter/post_permalink.js
```diff
@@ -6,18 +6,19 @@ let permalink;
 
 function postPermalinkFilter(data) {
   const { config } = this;
+  const { id, _id, slug, title, date } = data;
```
curbengh: :+1:

lib/plugins/filter/post_permalink.js
```diff
@@ -42,7 +43,15 @@ function postPermalinkFilter(data) {
     Object.defineProperty(meta, key, Object.getOwnPropertyDescriptor(data, key));
   }
 
-  return permalink.stringify(Object.assign(meta, config.permalink_defaults));
+  const keys2 = Object.keys(config.permalink_defaults);
+
+  for (const key of keys2) {
+    if (Object.prototype.hasOwnProperty.call(meta, key)) continue;
+
+    meta[key] = config.permalink_defaults[key];
```
SukkaW: Try [`deepMerge()`](https://github.com/hexojs/hexo-util#deepmergetarget-source) to see if `Maximum call stack size exceeded` still occurs.

PR #3963, opened 2019-12-14
feat(cache): option for cache rendered content

SukkaW:
## What does it do?

In #3756, @dailyrandomphoto disabled `cache rendered content` for `generate` and enabled for `hexo s`.

Enabling cache rendered content for `hexo s` will respond more quickly from second requests which is helpful when use `hexo-server` in production, but it also makes more difficult when developing theme or previewing other changes. So the `config.server.cache` is introduced in this PR. The documents for `hexo-server` will be updated after this PR is merged and released with `hexo@4.3`.

## How to test

```sh
git clone -b server-development https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/hexo/default_config.js
+++ b/lib/hexo/default_config.js
@@ -62,6 +62,9 @@ module.exports = {
   pagination_dir: 'page',
   // Extensions
   theme: 'landscape',
+  server: {
+    cache: false
+  },
   // Deployment
   deploy: {},
 
--- a/lib/hexo/index.js
+++ b/lib/hexo/index.js
@@ -311,7 +311,11 @@ class Hexo {
 
   watch(callback) {
     let useCache = false;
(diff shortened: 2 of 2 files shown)


curbengh:
> Enabling cache rendered content for hexo s will respond more quickly from second requests which is helpful when use hexo-server in production, but it also makes more difficult when developing theme or previewing other changes.

This PR disable it by default though, shouldn't it be enabled for most users?

SukkaW:
@curbengh 

Most users and theme developers are using hexo-server as a live preview. Only few of them are using hexo-server to serve their website.

curbengh:
Possibly supersedes https://github.com/hexojs/hexo-server/pull/76

PR #3996, opened 2019-12-20
refactor: utilize hexo-util

SukkaW:
## What does it do?

Utilize hexo-util's `CacheStream()` & `Cache()`;

## How to test

```sh
git clone -b utilize-hexo-util-cache https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/test/scripts/hexo/locals.js
+++ b/test/scripts/hexo/locals.js
@@ -24,10 +24,10 @@ describe('Locals', () => {
     locals.set('foo', () => 'foo');
 
     // cache should be clear after new data is set
-    should.not.exist(locals.cache.foo);
+    should.not.exist(locals.cache.cache.foo);
     locals.get('foo').should.eql('foo');
     // cache should be saved once it's get
-    locals.cache.foo.should.eql('foo');
+    locals.cache.cache.foo.should.eql('foo');
   });
 
   it('set() - not function', () => {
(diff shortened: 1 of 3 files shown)


test/scripts/hexo/locals.js
```diff
@@ -24,10 +24,10 @@ describe('Locals', () => {
     locals.set('foo', () => 'foo');
 
     // cache should be clear after new data is set
-    should.not.exist(locals.cache.foo);
+    should.not.exist(locals.cache.cache.foo);
```
SukkaW: Currently `Cache()` has no `dump()` function, but the cached content could be find through `.cache` property.

Will be refactor `cache.has()` later.

curbengh:
Slightly off-topic to this PR, would caching helps external_link filter? If assuming it's possible to cache all `<a>` in all pages (including posts), since there bounds to be repetitive links (especially navigation links), maybe it can help (even just a tiny bit, if memory usage doesn't increase).

SukkaW:
@curbengh Cache is already been enabled in hexo-util's `isExternalLink()`.

curbengh:
what about caching the every `<a></a>`, so it returns processed `<a></a>`. I'm assuming object cache is faster than String.replace().

SukkaW:
@curbengh I will give it a try in a separate PR.

PR #4017, opened 2019-12-28
fix(meta_generator): avoid an edge case

SukkaW:
## What does it do?

Current regex pattern will only matches `<meta name="generator" content="Hexo 4.2.0">` but not`<meta content="Hexo 4.2.0" name="generator">`,

## How to test

```sh
git clone -b meta_generator_fix https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/test/scripts/filters/meta_generator.js
+++ b/test/scripts/filters/meta_generator.js
@@ -25,13 +25,11 @@ describe('Meta Generator', () => {
   });
 
   it('no duplicate generator tag', () => {
-    const content = '<head><link>'
-      + '<meta name="generator" content="foo"></head>';
     hexo.config.meta_generator = true;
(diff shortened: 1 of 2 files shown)


test/scripts/filters/meta_generator.js
```diff
@@ -25,13 +25,11 @@ describe('Meta Generator', () => {
   });
 
   it('no duplicate generator tag', () => {
-    const content = '<head><link>'
-      + '<meta name="generator" content="foo"></head>';
     hexo.config.meta_generator = true;
-    const result = metaGenerator(content);
+    const resultType = str => typeof metaGenerator(str);
```
curbengh: 👍

Issue #3984, opened 2019-12-18
Improve test coverage


curbengh:
## Check List

Please check followings before submitting a new feature request.

- [x] I have already read [Docs page](https://hexo.io/docs/)
- [x] I have already searched existing issues

## Feature Request

This repo's unit test coverage currently stands at [![Coverage Status](https://coveralls.io/repos/hexojs/hexo/badge.svg?branch=master)](https://coveralls.io/r/hexojs/hexo?branch=master).

For anyone who is interested in improving it, easiest way to start is to search for ["missing-unit-test"](https://github.com/hexojs/hexo/search?q=missing-unit-test&type=Code) to locate functions that haven't been covered. Otherwise, you could also click on the badge link and it shows all the relevant lines.

[hexo-util](https://github.com/hexojs/hexo-util)  [![Coverage Status](https://coveralls.io/repos/hexojs/hexo-util/badge.svg?branch=master)](https://coveralls.io/r/hexojs/hexo-util?branch=master) also needs help.

seaoak:
@curbengh Thank you for nice waking-up call!

I checked a "missing-unit-test" point of `hexo-util`.
https://github.com/hexojs/hexo-util/blob/master/test/highlight.spec.js#L172

I found that `highlight()` of `hexo-util` might not work well for illegal syntax codes.
`highlight()` of `highlightjs` package will throw an exception if source text is illegal syntax codes.
https://highlightjs.readthedocs.io/en/latest/api.html#highlight-languagename-code-ignore-illegals-continuation

But `hexo-util/lib/highlight.js` might not consider that case.

I continue to investigate this issue (and make new issue in the repo of `hexo-util`).

SukkaW:
After #4422, the coverage has been increased from 97.03% to 98.26%.

stevenjoezhang:
```
npx nyc report --reporter=lcov --reporter=text
```

File                               | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s           
-----------------------------------|---------|----------|---------|---------|-----------------------------
All files                          |    97.3 |    92.55 |   98.22 |   98.32 |                             
 box                               |   92.81 |    83.64 |   96.23 |   94.78 |                             
  file.js                          |     100 |      100 |     100 |     100 |                             
  index.js                         |   92.09 |    83.64 |   95.83 |   94.17 | 155,230-231,235-236,249-250 
 extend                            |   97.55 |    93.69 |     100 |   99.06 |                             
[...]

PR #4050, opened 2020-01-02
test: improve coverage

SukkaW:
## What does it do?

This PR is a part of #3984

## How to test

```sh
git clone -b test-coverage-improve-1 https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/test/scripts/console/deploy.js
+++ b/test/scripts/console/deploy.js
@@ -18,6 +18,12 @@ describe('deploy', () => {
 
   after(() => fs.rmdir(hexo.base_dir));
 
+  it('no deploy config', () => {
+    delete hexo.config.deploy;
+
+    should.not.exist(deploy({test: true}));
+  });
+
   it('single deploy setting', () => {
     hexo.config.deploy = {
       type: 'foo',
--- a/test/scripts/console/config.js
+++ b/test/scripts/console/config.js
@@ -96,6 +96,11 @@ describe('config', () => {
     should.not.exist(config.language);
   });
 
+  it('write config: undefined', async () => {
+    const config = await writeConfig('meta_generator', 'undefined');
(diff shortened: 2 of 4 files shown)


lib/extend/tag.js
```diff
@@ -214,12 +214,7 @@ class Tag {
     this.env.addExtension(name, tag);
   }
 
-  render(str, options, callback) {
-    if (!callback && typeof options === 'function') {
-      callback = options;
-      options = {};
-    }
-
+  render(str, options = {}) {
```
SukkaW: `Tag.render()` just doesn't have callback, so I remove those lines and `callback` parameter directly.

test/scripts/console/deploy.js
```diff
@@ -18,6 +18,12 @@ describe('deploy', () => {
 
   after(() => fs.rmdir(hexo.base_dir));
 
+  it('no deploy config', () => {
+    delete hexo.config.deploy;
+
+    return deploy({test: true});
```
curbengh: What is the expected return value of `deploy()`, after it resolved?
suggest to refactor to async/await (including other tests in this PR).

lib/extend/tag.js
curbengh: Seems like a callback is promisified?

https://github.com/hexojs/hexo/blob/640b91e4b13d37a9fbdaf518a296354345bdf7ff/lib/extend/tag.js#L229

test/scripts/console/deploy.js
SukkaW: > What is the expected return value of `deploy()`, after it resolved?

No value at all. When there is no config, `console.log()` will be used to display an error message.

Issue #4048, opened 2020-01-02
Provide another filter to handle the HTML(after all rendered)


jiangtj:
## Check List

Please check followings before submitting a new feature request.

- [x] I have already read [Docs page](https://hexo.io/docs/)
- [x] I have already searched existing issues

## Feature Request

I tested adding the following filters to the initial example.

```js
let index = 0;
hexo.extend.filter.register('after_render:html', (data) => {
  console.log('--------------------:' + (index++))
  console.log(data.substr(0,10))
});
```

This filter will be executed many times, such as md to html or ejs to html, etc.

![image](https://user-images.githubusercontent.com/15902347/71648958-a1df9800-2d45-11ea-9a2a-9ab7e6582318.png)

I noticed that there are some built-in filters in hexo, but they are expected to be executed after the template engine to html rendering. If you can provide more accurate filters, you can reduce the number of regular matches and improve performance

https://github.com/hexojs/hexo/blob/a9bbc42219f522a0423f6dfb366f8a160821e48c/lib/plugins/filter/after_render/index.js#L6-L7

~Here are two ideas~
~- Provide `after_render:${input}-${output}`~
~- Provide `after_render:generate`~

SukkaW:
Hexo use `hexo.render.getOutput()` to find the `ext` of rendered output. So `after_render:html` will execute after every partial rendering.

jiangtj:
> Hexo use `hexo.render.getOutput()` to find the `ext` of rendered output. So `after_render:html` will execute after every partial rendering.

Or provide another filter to handle the HTML(after all rendered), `after_render:html` isn't very accurate, due to `md to html` or `partial` will trigger this filter. But expect to filter the final rendered HTML

SukkaW:
https://github.com/hexojs/hexo/issues/4048#issuecomment-570119228

> So `after_render:html` will execute after every rendering of post, partial, tag plugins.

Firstly, `hexo.render.render` and `hexo.render.renderSync` will trigger `hexo.extend.filter.execFilter` and `hexo.extend.filter.execFilterSync`.

https://github.com/hexojs/hexo/blob/0bebf604dda615cbaad86d86f9d18b464dcbe421/lib/hexo/render.js#L83-L89

`partial()` will trigger filter execution during precompile:

https://github.com/hexojs/hexo/blob/a9bbc42219f522a0423f6dfb366f8a160821e48c/lib/theme/view.js#L116-L126

[...]

PR #4051, opened 2020-01-04
feat(filter): add after_route_render filter

jiangtj:
## What does it do?

#4048 See discussion here, add `after_route_render` implement it.

## How to test

```sh
git clone -b filter https://github.com/jiangtj/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/test/scripts/hexo/hexo.js
+++ b/test/scripts/hexo/hexo.js
@@ -419,15 +419,15 @@ describe('Hexo', () => {
   });
 
   it('_generate() - after_route_render filter', () => {
-    const hook = sinon.spy(result => result.replace('</head>', 'content</head>'));
+    const hook = sinon.spy(result => '1');
     hexo.extend.filter.register('after_route_render', hook);
-    hexo.theme.setView('test.swig', '<head></head>');
+    hexo.theme.setView('test.swig', '0');
     hexo.extend.generator.register('test', () => ({
       path: 'test',
       layout: 'test'
     }));
     return hexo._generate()
-      .then(() => checkStream(route.get('test'), '<head>content</head>'))
+      .then(() => checkStream(route.get('test'), '1'))
       .then(() => hook.called.should.be.true);
   });
 
--- a/test/scripts/hexo/hexo.js
+++ b/test/scripts/hexo/hexo.js
@@ -419,15 +419,15 @@ describe('Hexo', () => {
   });
 
   it('_generate() - after_route_render filter', () => {
-    const hook = sinon.spy(result => '1');
+    const hook = sinon.spy(result => result.replace('foo', 'bar'));
     hexo.extend.filter.register('after_route_render', hook);
-    hexo.theme.setView('test.swig', '0');
+    hexo.theme.setView('test.swig', 'foo');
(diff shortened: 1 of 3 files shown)


test/scripts/hexo/hexo.js
```diff
@@ -418,6 +418,19 @@ describe('Hexo', () => {
     });
   });
 
+  it('_generate() - after_route_render filter', () => {
+    const hook = sinon.spy(result => result.replace('</head>', 'content</head>'));
```
SukkaW:
```suggestion
    const hook = sinon.spy(result => result.replace('foo', 'bar'));
```

test/scripts/hexo/hexo.js
```diff
@@ -418,6 +418,19 @@ describe('Hexo', () => {
...
+    hexo.extend.filter.register('after_route_render', hook);
+    hexo.theme.setView('test.swig', '<head></head>');
```
SukkaW:
```suggestion
    hexo.theme.setView('test.swig', 'foo');
```

test/scripts/hexo/hexo.js
```diff
@@ -418,6 +418,19 @@ describe('Hexo', () => {
...
+    hexo.extend.generator.register('test', () => ({
+      path: 'test',
+      layout: 'test'
+    }));
+    return hexo._generate()
+      .then(() => checkStream(route.get('test'), '<head>content</head>'))
```
SukkaW:
```suggestion
      .then(() => checkStream(route.get('test'), 'bar'))
```

test/scripts/hexo/hexo.js
SukkaW: The test case should be as simple as possible.

jiangtj:
@SukkaW modified

BTW, Celebrating 1000 test cases. 😂

curbengh:
> Or provide another filter to handle the HTML(after all rendered), after_render:html isn't very accurate, due to md to html or partial will trigger this filter. But expect to filter the final rendered HTML

So this filter only _filter_ the final html file? In that case, `after_route_render` can be confusing, as [`route`](https://hexo.io/api/router) can refer to any file. Initially I thought `after_route_render` pass every rendered html, js and css to the plugin, but actually only pass the html.

Wouldn't `after_html_render` (as initially suggested in https://github.com/hexojs/hexo/issues/4048) be more accurate?

SukkaW:
> Wouldn't `after_html_render` (as initially suggested in #4048) be more accurate?

@curbengh But it will be confused with `after_render:html` then.

curbengh:
> it will be confused with `after_render:html` then

They're functionally similar (to a plugin dev). We can make a note in the docs to recommend `after_html_render` over `after_render:html`, if the plugin dev doesn't mind backward compatibility.

curbengh:
Would it be possible to make `after_render:html` as an alias of `after_html_render`? If that's possible, this filter can be made as an internal API (e.g. `_after_html_render`), and then existing plugins can take advantage of it without changing filter.

PR #4061, opened 2020-01-06
refactor(helper): minor changes

SukkaW:
## What does it do?

Replace `const url_for = require('./url_for');` with `this.url_for()` to avoid extra require and retain consistent with other helpers.

## How to test

```sh
git clone -b refactor-helper https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/models/category.js
+++ b/lib/models/category.js
@@ -1,8 +1,7 @@
 'use strict';
 
 const { Schema } = require('warehouse');
-const { slugize } = require('hexo-util');
-const full_url_for = require('../plugins/helper/full_url_for');
+const { slugize, full_url_for } = require('hexo-util');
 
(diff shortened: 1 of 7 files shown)


curbengh:
why not have `full_url_for` part of `this` object, just like `this.url_for`.

SukkaW:
> why not have `full_url_for` part of `this` object, just like `this.url_for`.

I only replaced `url_for()` with `this.url_for` inside helpers, since they have same context.

SukkaW:
> Suggest to have `this.full_url_for` (in a separate PR) to avoid `.call()`.

@curbengh Only if they have same context.

PR #4070, opened 2020-01-10
feat: add prop to helper extension

### LINKED ISSUE #3919: For better performance, I need to set meta_generator to be false????
hexo version: 4.1.0
> Hexo 3.9.0+ inserts the tag automatically; to get the performance benefit, meta_generator: option needs to be disabled,
_config.yml
meta_generator: false

### LINKED ISSUE #4056: Use `Cache` to determine if filter is enabled
## Check List

## Feature Request

https://github.com/hexojs/hexo/blob/master/lib/plugins/filter/after_render/meta_generator.js

Currently, the meta_generator filter has performance problems. It is recommended to use a helper, but regular matching will still be used.

Maybe use [Cache](https://github.com/hexojs/hexo-util#cache) is a good solution.

```js
function metaGeneratorHelper() {
  cache.set('meta_generator', true);
  return `<meta name="generator" content="Hexo ${this.env.version}">`;
}
```

```js
function hexoMetaGeneratorInject(data) {
  let flag = cache.get('meta_generator');
  if (flag) return;
  const { config } = this;
  //...
}
```

## Others

#4047 Inject can also do this, if use helpers in the theme, can improve performance.

SukkaW:
## What does it do?

Fix #4056. Close #3919.

Automatically disable `meta_generator` filter if `meta_generator()` is used.

cc @curbengh @jiangtj 

## How to test

```sh
git clone -b extend-helper-prop https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/extend/helper.js
+++ b/lib/extend/helper.js
@@ -3,6 +3,7 @@
 class Helper {
   constructor() {
     this.store = {};
+    this.prop = {};
   }
 
   list() {
@@ -18,6 +19,17 @@ class Helper {
     if (typeof fn !== 'function') throw new TypeError('fn must be a function');
 
     this.store[name] = fn;
+  }
+
+  setProp(name, prop) {
+    if (!name) throw new TypeError('name is required');
+    if (!prop) throw new TypeError('prop is required');
+
+    this.prop[name] = prop;
+  }
+
+  getProp(name) {
+    return this.prop[name];
   }
 }
 
--- a/lib/plugins/filter/after_render/meta_generator.js
+++ b/lib/plugins/filter/after_render/meta_generator.js
@@ -9,7 +9,7 @@ function hexoMetaGeneratorInject(data) {
 
   const needInject = cache.apply('need-inject', () => {
     if (!config.meta_generator
-        || this.extend.helper.getProp('meta_generator')
+        || this.extend.helper.getProp('meta_generator_used')
         || data.match(/<meta([\s]+|[\s]+[^<>]+[\s]+)name=['|"]?generator['|"]?/i)) return false;
 
     return true;
--- a/lib/plugins/filter/after_render/meta_generator.js
+++ b/lib/plugins/filter/after_render/meta_generator.js
@@ -2,9 +2,10 @@
 
 const { Cache } = require('hexo-util');
 const cache = new Cache();
+let metaGeneratorTag = '';
 
 function hexoMetaGeneratorInject(data) {
-  const { config, version } = this;
+  const { config } = this;
 
   const needInject = cache.apply('need-inject', () => {
     if (!config.meta_generator
@@ -15,8 +16,9 @@ function hexoMetaGeneratorInject(data) {
   });
 
   if (!needInject) return;
+  metaGeneratorTag = metaGeneratorTag || `<meta name="generator" content="Hexo ${this.version}">`;
 
-  return data.replace(/<head>(?!<\/head>).+?<\/head>/s, str => str.replace('</head>', `<meta name="generator" content="Hexo ${version}"></head>`));
+  return data.replace(/<head>(?!<\/head>).+?<\/head>/s, str => str.replace('</head>', metaGeneratorTag + '</head>'));
 }
 
 module.exports = hexoMetaGeneratorInject;
--- a/lib/plugins/filter/after_render/meta_generator.js
+++ b/lib/plugins/filter/after_render/meta_generator.js
@@ -1,11 +1,20 @@
 'use strict';
 
+const { Cache } = require('hexo-util');
+const cache = new Cache();
+
 function hexoMetaGeneratorInject(data) {
   const { config, version } = this;
 
-  if (!config.meta_generator
-    || this.extend.helper.getProp('meta_generator')
-    || data.match(/<meta([\s]+|[\s]+[^<>]+[\s]+)name=['|"]?generator['|"]?/i)) return;
+  const needInject = cache.apply('need-inject', () => {
+    if (!config.meta_generator
+        || this.extend.helper.getProp('meta_generator')
+        || data.match(/<meta([\s]+|[\s]+[^<>]+[\s]+)name=['|"]?generator['|"]?/i)) return false;
+
+    return true;
+  });
+
+  if (!needInject) return;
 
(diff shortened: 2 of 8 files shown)


lib/plugins/filter/after_render/meta_generator.js
```diff
@@ -1,13 +1,22 @@
...
+const { Cache } = require('hexo-util');
+const cache = new Cache();
+
 function hexoMetaGeneratorInject(data) {
-  const { config } = this;
-  if (!config.meta_generator
-    || data.match(/<meta([\s]+|[\s]+[^<>]+[\s]+)name=['|"]?generator['|"]?/i)) return;
+  const { config, version } = this;
+
+  const needInject = cache.apply('need-inject', () => {
```
SukkaW: I just assume `meta[name="generator"]` will be inserted to every page if `meta_generator()` is used or `<meta name="generator" content="Hexo [version]">` is manually added in the theme.

lib/extend/helper.js
```diff
@@ -19,6 +20,17 @@ class Helper {
 
     this.store[name] = fn;
   }
+
+  setProp(name, prop) {
+    if (!name) throw new TypeError('name is required');
+    if (!prop) throw new TypeError('prop is required');
```
curbengh: It shouldn't be TypeError as this is not type checking.

lib/plugins/helper/meta_generator.js
```diff
@@ -1,7 +1,8 @@
...
 
-function metaGeneratorHelper() {
-  return `<meta name="generator" content="Hexo ${this.env.version}">`;
-}
-
-module.exports = metaGeneratorHelper;
+module.exports = ctx => {
+  return () => {
+    if (!ctx.extend.helper.getProp('meta_generator')) ctx.extend.helper.setProp('meta_generator', true);
+    return `<meta name="generator" content="Hexo ${ctx.env.version}">`;
```
curbengh: Since this helper is [bind](https://github.com/hexojs/hexo/pull/4070/files#diff-0c88d2f71e5c4ea3ed9b3c0933b15686R46) to ctx, I believe `ctx.version` should work?

test/scripts/extend/helper.js
```diff
@@ -31,18 +34,45 @@ describe('Helper', () => {
...
+
+  it('setProp()', () => {
+    // name prop
+    h.setProp('foo', 'bar');
+
+    h.prop.foo.should.eql('bar');
+
+    // no name
+    try {
+      h.setProp();
```
curbengh: Recently I found `should.fail()` is necessary in `try...catch`. see https://github.com/hexojs/hexo-fs/pull/51#discussion_r364247814

lib/extend/helper.js
SukkaW: L18 uses `TypeError` as well.

lib/plugins/helper/meta_generator.js
SukkaW: @curbengh I will have a test on a dummy site to see if it will cause `undefined` again.

test/scripts/extend/helper.js
SukkaW: Added.

lib/plugins/helper/meta_generator.js
SukkaW: Update to `ctx.version`.

lib/extend/helper.js
curbengh: oh well, `undefined` can be considered as a type.

lib/plugins/helper/meta_generator.js
curbengh: also verified in my test.

test/scripts/filters/meta_generator.js
```diff
@@ -1,11 +1,19 @@
...
 describe('Meta Generator', () => {
   const Hexo = require('../../../lib/hexo');
   const hexo = new Hexo();
-  const metaGenerator = require('../../../lib/plugins/filter/after_render/meta_generator').bind(hexo);
+  let metaGenerator;
+  const metaGeneratorHelper = require('../../../lib/plugins/helper/meta_generator')(hexo);
   const cheerio = require('cheerio');
 
+  beforeEach(async () => {
+    await decache('../../../lib/plugins/filter/after_render/meta_generator');
```
curbengh: I don't think `await` has any effect here. I can't find any promise/async in decache's codebase.

lib/plugins/filter/after_render/meta_generator.js
```diff
@@ -1,13 +1,24 @@
 'use strict';
 
+const { Cache } = require('hexo-util');
+const cache = new Cache();
```
dailyrandomphoto: Using local val is enough.

```js
let needInject;

...
function hexoMetaGeneratorInject(data) {
  const { config } = this;
  if (typeof needInject !== 'boolean') {
      needInject = !(!config.meta_generator
          || this.extend.helper.getProp('meta_generator')
          || data.match(/<meta([\s]+|[\s]+[^<>]+[\s]+)name=['|"]?generator['|"]?/i));
  }
```

lib/plugins/helper/meta_generator.js
```diff
@@ -1,7 +1,11 @@
...
 
-function metaGeneratorHelper() {
-  return `<meta name="generator" content="Hexo ${this.env.version}">`;
-}
+let metaGeneratorTag = '';
 
-module.exports = metaGeneratorHelper;
+module.exports = ctx => {
+  return () => {
+    if (!ctx.extend.helper.getProp('meta_generator')) ctx.extend.helper.setProp('meta_generator', true);
```
dailyrandomphoto: Suggest `use_meta_generator_helper` as the property name.
If `meta_generator` is used, we don't know what this means.

lib/plugins/filter/after_render/meta_generator.js
SukkaW: IMHO, use `cache.apply` here made code more readable since `cache.apply` support caching return value of given function.

lib/plugins/helper/meta_generator.js
SukkaW: Change to `meta_generator_helper_used`

Issue #3890, opened 2019-11-29
The proposal of installing theme through npm


SukkaW:
The original discussion is at #2471 

Here is my proposal: 

1. We should not treat `hexo-theme-` prefixed package as a hexo plugin. See @tomap 's commit:  https://github.com/hexojs/hexo/compare/master...tomap:feature/themeAsPackagediff-3d0f701643104f3b0e388b8f6a9a7f3eR38
2. Users are still required to specify their theme's name in `_config.yml`, because...
3. Hexo will read `themes/[name]` directory first, then search `hexo-theme-[name]` directory under `node_modules`. It means the theme under `themes` directory should have higher priority. In this way no extra `theme_from_npm: true` configuration is needed. 
4. `theme_config` in `_config.yml` looks good, but it is better to have a separate theme config. We could introduce `_config.[name].yml`. Currently, `theme_config` has higher priority than `_config.yml` under theme directories. After bring up `_config.[name].yml`, I suggest the `theme_config in _config.yml > _config.[name].yml > _config.yml under theme directories` which will retain highest priority for `theme_config`.
5. ~~Also, we need to fix the issue that some plugins can not read theme config from `theme_config`: https://github.com/hexojs/hexo/issues/2471#issuecomment-557877081~~ 

IMHO, we could easily bring up this feature without any Breaking Changes: the theme structure, current configurations and setup will not be affected. Yes, I mean we can even bring up this feature during Hexo 4.x development.

cc @hexojs/core @jiangtj

SukkaW:
FYI: 
Currently there are more than 50 packages on npm has `hexo-theme-` prefixed: https://www.npmjs.com/search?q=hexo-theme

stevenjoezhang:
> We should not treat hexo-theme- prefixed package as a hexo plugin.

Is there a good way to handle this situation?
Plugin: https://www.npmjs.com/package/hexo-theme-next-anchor
Theme: https://www.npmjs.com/package/hexo-theme-next

tomap:
> > We should not treat hexo-theme- prefixed package as a hexo plugin.
> 
> Is there a good way to handle this situation?
> Plugin: https://www.npmjs.com/package/hexo-theme-next-anchor
> Theme: https://www.npmjs.com/package/hexo-theme-next

We could contact all owners of `hexo-theme-*` on npm.org when they are not themes

The package hexo-theme-next-anchor is indeed a plugin. 
We should suggest the owner to change it's name

[...]

PR #4112, opened 2020-01-31
feat(load_config): support theme_dir in node_modules

SukkaW:
## What does it do?

The part of #3890

> Hexo will read themes/[name] directory first, then search hexo-theme-[name] directory under node_modules. It means the theme under themes directory should have higher priority. In this way no extra theme_from_npm: true configuration is needed.

## How to test

```sh
git clone -b theme_dir_load https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/hexo/load_config.js
+++ b/lib/hexo/load_config.js
@@ -4,64 +4,60 @@ const { sep, resolve, join, parse } = require('path');
 const tildify = require('tildify');
 const Theme = require('../theme');
 const Source = require('./source');
-const { exists, readdir, existsSync } = require('hexo-fs');
+const { exists, readdir } = require('hexo-fs');
 const { magenta } = require('chalk');
 const { deepMerge } = require('hexo-util');
 
-module.exports = ctx => {
+module.exports = async ctx => {
   if (!ctx.env.init) return;
 
   const baseDir = ctx.base_dir;
   let configPath = ctx.config_path;
 
-  return exists(configPath).then(exist => {
-    return exist ? configPath : findConfigPath(configPath);
-  }).then(path => {
-    if (!path) return;
+  const path = await exists(configPath) ? configPath : await findConfigPath(configPath);
+  if (!path) return;
+  configPath = path;
 
-    configPath = path;
-    return ctx.render.render({path});
-  }).then(config => {
-    if (!config || typeof config !== 'object') return;
+  let config = await ctx.render.render({ path });
+  if (!config || typeof config !== 'object') return;
 
-    ctx.log.debug('Config loaded: %s', magenta(tildify(configPath)));
+  ctx.log.debug('Config loaded: %s', magenta(tildify(configPath)));
 
-    ctx.config = deepMerge(ctx.config, config);
-    config = ctx.config;
+  ctx.config = deepMerge(ctx.config, config);
+  config = ctx.config;
 
-    ctx.config_path = configPath;
+  ctx.config_path = configPath;
 
-    config.root = config.root.replace(/\/*$/, '/');
-    config.url = config.url.replace(/\/+$/, '');
+  config.root = config.root.replace(/\/*$/, '/');
+  config.url = config.url.replace(/\/+$/, '');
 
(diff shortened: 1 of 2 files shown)


curbengh:
I don't see any bluebird functions. can the whole `load_config` be refactored to async, so that `existsSync` can be avoided?

SukkaW:
@curbengh Have been refactored in async / await syntax

lib/hexo/load_config.js
```diff
@@ -4,54 +4,60 @@ const { sep, resolve, join, parse } = require('path');
...
+  const themeDirFromNodeModules = join(ctx.plugin_dir, 'hexo-theme-' + theme) + sep; // base_dir/node_modules/hexo-theme-[config.theme]/
+
+  // themeDirFromThemes has higher priority than themeDirFromNodeModules
+  if (await exists(themeDirFromThemes)) {
+    ctx.theme_dir = themeDirFromThemes;
+  } else if (await exists(themeDirFromNodeModules)) {
+    ctx.theme_dir = themeDirFromNodeModules;
+  }
+  ctx.theme_script_dir = join(ctx.theme_dir, 'scripts') + sep;
+  ctx.theme = new Theme(ctx);
```
jiangtj:
```suggestion
  ctx.theme = new Theme(ctx);
  ctx.theme.ignore = ['**/node_modules/hexo-theme-*/node_modules/**', '**/node_modules/hexo-theme-*/.git/**']
```

see: https://github.com/hexojs/hexo/blob/master/lib/box/index.js#L11

jiangtj: It is used in `_readDir`
https://github.com/hexojs/hexo/blob/master/lib/box/index.js#L88

jiangtj: Another problem is that whether it is ignored or not, it will affect the performance of watch.

My theme tried to do the same thing in index.js https://github.com/jiangtj/hexo-theme-cake/blob/master/index.js, but when I debug using yarn command (yarn link), Because there are more develop dependencies, `hexo cl && hexo s` will take longer

May be better solution is to listen only to the specified directory, like `source` `scripts` `layout` `languages`

stevenjoezhang: This is more symmetrical
```js
  let ignoreCfg;
  if (await exists(themeDirFromThemes)) {
    ctx.theme_dir = themeDirFromThemes;
    ignoreCfg = ['**/themes/*/node_modules/**', '**/themes/*/.git/**'];
  } else if (await exists(themeDirFromNodeModules)) {
    ctx.theme_dir = themeDirFromNodeModules;
    ignoreCfg = ['**/node_modules/hexo-theme-*/node_modules/**', '**/node_modules/hexo-theme-*/.git/**'];
  }
  ctx.theme_script_dir = join(ctx.theme_dir, 'scripts') + sep;
  ctx.theme = new Theme(ctx);
```

And https://github.com/hexojs/hexo/blob/master/lib/box/index.js#L11 can be removed

SukkaW: @jiangtj @stevenjoezhang This could be done in another PR then.

stevenjoezhang: I agree that another PR can be used to deal with the efficiency issues caused by `watch()`. After resolving the conflict, this PR can be merged

SukkaW:
Rebased to latest master.

lib/hexo/load_config.js
SukkaW: @stevenjoezhang Rebased.

sergeyzwezdin:
@SukkaW Does it support scoped package name? I.e. `@somescope/hexo-theme-mytheme`.

SukkaW:
> Does it support scoped package name

Currently, no.

sergeyzwezdin:
Any plans to introduce it?

SukkaW:
> Any plans to introduce it?

You can opened a new issue for Feature Request so that we can track it.

sergeyzwezdin:
> You can opened a new issue for Feature Request so that we can track it.

Added #4496

PR #4115, opened 2020-02-08
Refactor test

segayuu:
## How to test

```sh
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/test/scripts/box/box.js
+++ b/test/scripts/box/box.js
@@ -270,7 +270,7 @@ describe('Box', () => {
     await writeFile(join(box.base, 'foo.txt'), 'foo');
     await box.process();
 
-    data.should.to.have.all.keys(['foo.txt']);
+    data.should.have.all.keys(['foo.txt']);
 
     await rmdir(box.base);
   });
@@ -289,7 +289,7 @@ describe('Box', () => {
     ]);
     await box.process();
 
-    data.should.to.have.all.keys(['foo.txt']);
+    data.should.have.all.keys(['foo.txt']);
 
     await rmdir(box.base);
   });
@@ -309,7 +309,7 @@ describe('Box', () => {
     ]);
     await box.process();
 
-    data.should.to.have.all.keys(['foo.txt']);
+    data.should.have.all.keys(['foo.txt']);
 
     await rmdir(box.base);
   });
@@ -328,7 +328,7 @@ describe('Box', () => {
     ]);
     await box.process();
 
-    data.should.to.have.all.keys(['foo.txt']);
+    data.should.have.all.keys(['foo.txt']);
 
     await rmdir(box.base);
   });
@@ -348,7 +348,7 @@ describe('Box', () => {
     ]);
     await box.process();
 
-    data.should.to.have.all.keys(['foo.txt']);
+    data.should.have.all.keys(['foo.txt']);
 
     await rmdir(box.base);
   });
--- a/test/scripts/box/box.js
+++ b/test/scripts/box/box.js
@@ -267,9 +267,7 @@ describe('Box', () => {
       data[file.path] = file;
     });
 
-    await Promise.all([
-      writeFile(join(box.base, 'foo.txt'), 'foo')
-    ]);
+    await writeFile(join(box.base, 'foo.txt'), 'foo');
     await box.process();
 
     data.should.to.have.all.keys(['foo.txt']);
@@ -363,7 +361,7 @@ describe('Box', () => {
 
     box.addProcessor(processor);
 
-    await Promise.all([writeFile(src, 'a')]);
+    await writeFile(src, 'a');
     await box.watch();
     box.isWatching().should.to.be.true;
     await Promise.delay(500);
--- a/test/scripts/box/box.js
+++ b/test/scripts/box/box.js
@@ -83,13 +83,7 @@ describe('Box', () => {
 
     await box.process();
 
-    const keys = Object.keys(data);
-    let key, item;
-
-    for (let i = 0, len = keys.length; i < len; i++) {
-      key = keys[i];
-      item = data[key];
-
+    for (const [key, item] of Object.entries(data)) {
       item.path.should.eql(key);
       item.source.should.eql(join(box.base, key));
       item.type.should.eql('create');
@@ -278,10 +272,7 @@ describe('Box', () => {
     ]);
     await box.process();
 
-    const keys = Object.keys(data);
-
-    keys.length.should.eql(1);
-    keys[0].should.eql('foo.txt');
+    data.should.to.have.all.keys(['foo.txt']);
 
     await rmdir(box.base);
   });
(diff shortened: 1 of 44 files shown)


test/scripts/box/box.js
```diff
@@ -251,7 +236,7 @@ describe('Box', () => {
     ]);
     await box.process();
 
-    processor.calledOnce.should.eql(true);
+    processor.calledOnce.should.to.be.true;
```
SukkaW:
```suggestion
    processor.calledOnce.should.be.true;
```

Since we are using `chai` to make our unit test into human language, we should consider about grammar things. There is no `to` after `should` in English, and chai could still work without `.to`.

test/scripts/filters/external_link.js
```diff
@@ -21,8 +21,7 @@ describe('External link', () => {
 
     hexo.config.external_link.enable = false;
 
-    const result = typeof externalLink(content);
-    result.should.eql('undefined');
+    should.equal(externalLink(content), undefined);
```
SukkaW:
```suggestion
    should.not.exist(externalLink(content));
```

chai has `should.not.exist()` syntax.

test/scripts/box/box.js
segayuu: My personal preference is to reduce unnecessary chains, so I would like to agree with your review, but if you use the should interface, you should follow the grammar as much as possible.

test/scripts/filters/external_link.js
segayuu: I don't understand the obvious reasons for obscuring null and undefined.
However, I think that should.equal () was not used in previous test code, and I think we should only use should.exist () to avoid confusion.

SukkaW: @segayuu `to.be.` should be used with chai's `expect` syntax, not with `should` syntax.

test/scripts/box/box.js
SukkaW: @segayuu 

From chai's documents, I can not see any example of `should.to.be`. It seems that `to.be` should only be used in `expect()` syntax (`expect().to.be`). And with `should` syntax, `should.be` will be enough.

test/scripts/filters/external_link.js
```diff
@@ -33,8 +32,7 @@ describe('External link', () => {
 
     hexo.config.external_link.field = 'post';
 
-    const result = typeof externalLink(content);
-    result.should.eql('undefined');
+    should.equal(externalLink(content), undefined);
```
SukkaW:
```suggestion
    should.not.exist((externalLink(content));
```

test/scripts/filters/external_link.js
```diff
@@ -97,9 +95,7 @@ describe('External link', () => {
 
     hexo.config.external_link = false;
 
-    const result = typeof externalLink(content);
-    result.should.eql('undefined');
-
+    should.equal(externalLink(content), undefined);
```
SukkaW:
```suggestion
    should.not.exist((externalLink(content));
```

test/scripts/filters/meta_generator.js
```diff
@@ -20,16 +20,15 @@ describe('Meta Generator', () => {
     hexo.config.meta_generator = false;
     const result = metaGenerator(content);
 
-    const resultType = typeof result;
-    resultType.should.eql('undefined');
+    should.equal(result, undefined);
```
SukkaW:
```suggestion
    should.not.exist(result);
```

test/scripts/helpers/meta_generator.js
```diff
@@ -8,9 +8,8 @@ describe('meta_generator', () => {
 
   it('default', () => {
     const { version } = hexo;
-    const versionType = typeof version;
 
-    versionType.should.not.eql('undefined');
+    should.not.equal(version, undefined);
```
SukkaW:
```suggestion
    should.not.exist(version);
```

test/scripts/filters/meta_generator.js
```diff
@@ -20,16 +20,15 @@ describe('Meta Generator', () => {
...
   });
 
   it('no duplicate generator tag', () => {
     hexo.config.meta_generator = true;
-    const resultType = str => typeof metaGenerator(str);
+    const result = str => metaGenerator(str);
```
SukkaW:
```suggestion
```

test/scripts/filters/meta_generator.js
```diff
@@ -20,16 +20,15 @@ describe('Meta Generator', () => {
...
 
-    resultType('<head><link><meta name="generator" content="foo"></head>').should.eql('undefined');
-    resultType('<head><link><meta content="foo" name="generator"></head>').should.eql('undefined');
+    should.equal(result('<head><link><meta name="generator" content="foo"></head>'), undefined);
```
SukkaW:
```suggestion
    should.not.exist(metaGenerator('<head><link><meta name="generator" content="foo"></head>'));
```

test/scripts/filters/meta_generator.js
```diff
@@ -20,16 +20,15 @@ describe('Meta Generator', () => {
...
+    should.equal(result('<head><link><meta content="foo" name="generator"></head>'), undefined);
```
SukkaW:
```suggestion
    should.not.exist(metaGenerator('<head><link><meta content="foo" name="generator"></head>'));
```

test/scripts/processors/data.js
```diff
@@ -51,7 +49,7 @@ describe('data', () => {
       path: 'users.yaml'
     });
 
-    typeOf(pattern.match('users.json')).should.eql('undefined');
+    should.equal(pattern.match('users.json'), undefined);
```
SukkaW:
```suggestion
    should.not.exist(pattern.match('users.json'));
```

test/scripts/processors/data.js
```diff
@@ -141,6 +139,6 @@ describe('data', () => {
       data: {foo: 'bar'}
     });
     await process(file);
-    typeOf(Data.findById('users')).should.eql('undefined');
+    should.equal(Data.findById('users'), undefined);
```
SukkaW:
```suggestion
    should.not.exist(Data.findById('users'));
```

test/scripts/box/box.js
segayuu: Certainly, the scene used in the should interface on the official website could not be confirmed. Although it is incomplete in English, the difference will be changed along the official website.

segayuu:
Note that `should(.not).exist()` is [deprecated on the official site](https://www.chaijs.com/api/bdd/#method_exist) (because it is better to treat null and undefined separately).

SukkaW:
@segayuu For `hexo.extend.filter`, it will treat `null` and `undefined` return value as not doing anything. So it is fine to use `should.not.exist` in `filter` related test cases.

test/scripts/filters/meta_generator.js
```diff
@@ -11,7 +11,7 @@ describe('Meta Generator', () => {
     const result = metaGenerator(content);
 
     const $ = cheerio.load(result);
-    $('meta[name="generator"]').length.should.eql(1);
+    $('meta[name="generator"]').should.to.have.lengthOf(1);
```
SukkaW:
```suggestion
    $('meta[name="generator"]').should.have.lengthOf(1);
```

Not only `should.to.be` should be replaced with `should.be` but also `should.to.have`  should be replaced with  `should.have`.

segayuu:
@SukkaW Thanks!

PR #4120, opened 2020-02-09
feat(load_theme_config): support alternate theme config

SukkaW:
## What does it do?

The part of #3890

> `theme_config` in `_config.yml` looks good, but it is better to have a separate theme config. We could introduce `_config.[name].yml`. Currently, `theme_config` has higher priority than `_config.yml` under theme directories. After bring up `_config.[name].yml`, I suggest the `theme_config in _config.yml > _config.[name].yml > _config.yml under theme directories` which will retain highest priority for `theme_config`.

## How to test

```sh
git clone -b theme-config-dedicated https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/test/scripts/hexo/load_theme_config.js
+++ b/test/scripts/hexo/load_theme_config.js
@@ -1,0 +1,65 @@
+'use strict';
+
+const { join } = require('path');
+const { mkdirs, unlink, writeFile, rmdir } = require('hexo-fs');
+
+describe('Load alternate theme config', () => {
+  const Hexo = require('../../../lib/hexo');
+  const hexo = new Hexo(join(__dirname, 'config_test'), {silent: true});
+  const loadThemeConfig = require('../../../lib/hexo/load_theme_config');
+
+  hexo.env.init = true;
+
+  before(() => mkdirs(hexo.base_dir).then(() => hexo.init()));
+
+  after(() => rmdir(hexo.base_dir));
+
+  beforeEach(() => {
+    hexo.config.theme_config = { foo: { bar: 'ahhhhhh' } };
+    hexo.config.theme = 'test_theme';
+  });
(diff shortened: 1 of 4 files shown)


test/scripts/hexo/load_theme_config.js
```diff
@@ -0,0 +1,65 @@
...
+  it('_config.[theme].yml exists', () => {
+    const configPath = join(hexo.base_dir, '_config.test_theme.yml');
+
+    return writeFile(configPath, 'bar: 1').then(() => loadThemeConfig(hexo)).then(() => {
+      hexo.config.theme_config.bar.should.eql(1);
+    }).finally(() => unlink(configPath));
+  });
+
+  it('_config.[theme].json exists', () => {
+    const configPath = join(hexo.base_dir, '_config.test_theme.json');
```
YoshinoriN: Is this feature will allow other than `YAML` format?

SukkaW: Since I have just port the feature from `load_config`. So if Hexo support `_config.json`, then `_config.test_theme.json` should be supported as well (That's why I include this in test cases)

stevenjoezhang:
Currently `theme_config` only affects template local

https://github.com/hexojs/hexo/blob/6d29793e3792ccd84d51c8118046ba90b2eaf6be/lib/hexo/index.js#L360

Which means something like

```js
hexo.extend.filter.register('', () => {
  let { key } = hexo.theme.config;
  // If the key is defined in `theme_config`, it would be null here
}, 0);
```

does not work.

Is it possible to merge `config.theme_config` into `theme.config` after `ctx.theme` loads?

PR #4134, opened 2020-02-19
perf(external_link): cache config

SukkaW:
## What does it do?

This PR add cache for 2 values:

- `EXTERNAL_LINK_(POST|SITE)_ENABLED`

Since hexo will only run filter for `post` or `site`, it is important to return earlier for another one.

- `EXTERNAL_LINK_(POST|SITE)_CONFIG`

Currently, hexo handles `hexo.config.external_link` inside the filter. With config being cached there will be no need to run `Object.assign()` again and again.

## How to test

```sh
git clone -b optimize-external-filter https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/filter/after_post_render/external_link.js
+++ b/lib/plugins/filter/after_post_render/external_link.js
@@ -1,23 +1,33 @@
 'use strict';
 
 const { isExternalLink } = require('hexo-util');
+let EXTERNAL_LINK_POST_CONFIG;
+let EXTERNAL_LINK_POST_ENABLED = true;
 
 function externalLinkFilter(data) {
+  if (!EXTERNAL_LINK_POST_ENABLED) return;
+
   const { config } = this;
 
-  if (typeof config.external_link === 'undefined' || typeof config.external_link === 'object'
-    || config.external_link === true) {
-    config.external_link = Object.assign({
-      enable: true,
-      field: 'site',
-      exclude: []
-    }, config.external_link);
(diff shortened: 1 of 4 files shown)


lib/plugins/filter/after_post_render/external_link.js
```diff
@@ -1,23 +1,35 @@
...
+  if (typeof EXTERNAL_LINK_POST_CONFIG === 'undefined') {
+    if (typeof config.external_link === 'undefined' || typeof config.external_link === 'object'
+      || config.external_link === true) {
+      EXTERNAL_LINK_POST_CONFIG = Object.assign({
+        enable: true,
+        field: 'site',
+        exclude: []
+      }, config.external_link);
+      // Write config back
+      config.external_link = EXTERNAL_LINK_POST_CONFIG;
```
curbengh: I'm personally ok with this, but user might caught off guard, https://github.com/hexojs/hexo-generator-feed/issues/115

SukkaW: @curbengh Since the original code will write the config back to `config.external_link` directly, so I retain this approach in this PR.

However, I do agree to remove this line since it seems not break anything.

SukkaW:
@YoshinoriN 

IMHO, we have no necessary to do that.
We already split `isExternalLink()` into `hexo-util`. The rest of the code is very specific and no need to split into util.

PR #4143, opened 2020-02-21
ci: add GitHub Actions to run linter

Himself65:
example sceenshot:

image

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -1,0 +1,27 @@
+name: Linter
+
+on: [push, pull_request]
+
+jobs:
+  linter:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Use Node.js 12.x
+        uses: actions/setup-node@v1
+        with:
+          node-version: '12.x'
+      - name: Cache NPM dependencies
+        uses: actions/cache@v1
+        with:
+          path: node_modules
+          key: ${{ runner.OS }}-npm-cache
+          restore-keys: |
+            ${{ runner.OS }}-npm-cache
+      - name: Install Dependencies
+        run: npm install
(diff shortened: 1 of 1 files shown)


.github/workflows/linter.yml
```diff
@@ -0,0 +1,27 @@
...
+          node-version: '12.x'
+      - name: Cache NPM dependencies
+        uses: actions/cache@v1
+        with:
+          path: node_modules
+          key: ${{ runner.OS }}-npm-cache
+          restore-keys: |
+            ${{ runner.OS }}-npm-cache
+      - name: Install Dependencies
+        run: yarn --frozen-lockfile --non-interactive
```
SukkaW: Although `yarn` should be considered as faster, we prefer using `npm` in CI (Like what we do in Travis)

.github/workflows/linter.yml
```diff
@@ -0,0 +1,27 @@
...
+        with:
+          path: node_modules
+          key: ${{ runner.OS }}-npm-cache
+          restore-keys: |
+            ${{ runner.OS }}-npm-cache
+      - name: Install Dependencies
+        run: npm install
+      - name: Lint
+        run: |
+          yarn eslint
```
SukkaW:
```suggestion
          npm run eslint
```

Will it works?

.github/workflows/linter.yml
```diff
@@ -0,0 +1,27 @@
...
+  linter:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Use Node.js 12.x
+        uses: actions/setup-node@v1
+        with:
+          node-version: '12.x'
+      - name: Cache NPM dependencies
+        uses: actions/cache@v1
```
curbengh: 👍

PR #4161, opened 2020-02-26
fix(post): use non-greedy regular expressions

stevenjoezhang:
## What does it do?

Currently, the regular expression `rSwigFullBlock` used in the `escapeAllSwigTags` method in `lib/hexo/post.js` may incorrectly match the Swig / Nunjucks block. For example, the user writes the following in a markdown document

```md
{% note danger %}note text, note text, note text{% endnote %}

## Title

{% note danger %}note text, note text, note text{% endnote %}
```

Then, this regular expression will match all three lines, starting from `{% note danger %}` in the first line to the ending `{% endnote %}` in the third line, which means that `## Title` in the second line is only processed by Nunjucks but not by the Markdown renderer, so it appears as `## Title` instead of `<h2>Title</h2>` in the output HTML.

截屏2020-02-27上午10 46 09

By using non-greedy regular expression, the first and third line will be matched separately to get the correct results.

I'm not an expert in regular expressions, so I can't guarantee whether this change will cause other bugs. Suggestions are welcome, thank you!

Related issues:
https://github.com/iissnan/hexo-theme-next/issues/1674
https://github.com/iissnan/hexo-theme-next/issues/1678
https://github.com/iissnan/theme-next-docs/issues/163

## How to test

```sh
git clone -b stevenjoezhang-patch-1 https://github.com/stevenjoezhang/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [ ] Passed the CI test.

### DIFF
--- a/lib/hexo/post.js
+++ b/lib/hexo/post.js
@@ -43,7 +43,7 @@ class PostRenderCache {
     const rSwigVar = /\{\{[\s\S]*?\}\}/g;
     const rSwigComment = /\{#[\s\S]*?#\}/g;
     const rSwigBlock = /\{%[\s\S]*?%\}/g;
-    const rSwigFullBlock = /\{% *(.+?)(?: *| +.*)%\}[\s\S]+?\{% *end\1 *%\}/g;
+    const rSwigFullBlock = /\{% *(.+?)(?: *| +.*?)%\}[\s\S]+?\{% *end\1 *%\}/g;
 
     const escape = _str => _escapeContent(this.cache, _str);
     return str.replace(rSwigFullBlock, escape)
--- a/test/scripts/hexo/post.js
+++ b/test/scripts/hexo/post.js
@@ -888,4 +888,48 @@ describe('Post', () => {
     });
   });
 
+  // test for PR #4161
+  it('render() - adjacent tags', () => {
+    const content = [
+      '{% pullquote %}content1{% endpullquote %}',
+      '',
+      'This is a following paragraph',
+      '',
+      '{% pullquote %}content2{% endpullquote %}'
+    ].join('\n');
+
+    return post.render(null, {
+      content,
+      engine: 'markdown'
+    }).then(data => {
+      data.content.trim().should.eql([
+        '<blockquote class="pullquote"><p>content1</p>\n</blockquote>\n',
+        '<p>This is a following paragraph</p>\n',
+        '<blockquote class="pullquote"><p>content2</p>\n</blockquote>\n'
+      ].join(''));
+    });
+  });
+
+  // test for PR #4161
+  it('render() - adjacent tags with args', () => {
+    const content = [
+      '{% pullquote center %}content1{% endpullquote %}',
+      '',
+      'This is a following paragraph',
+      '',
+      '{% pullquote center %}content2{% endpullquote %}'
+    ].join('\n');
+
+    return post.render(null, {
+      content,
+      engine: 'markdown'
+    }).then(data => {
+      data.content.trim().should.eql([
+        '<blockquote class="pullquote center"><p>content1</p>\n</blockquote>\n',
+        '<p>This is a following paragraph</p>\n',
+        '<blockquote class="pullquote center"><p>content2</p>\n</blockquote>\n'
+      ].join(''));
+    });
+  });
+
 });


SukkaW:
Since this PR has fixed a issue but you are not sure whether your regexp will solve that issue or not, why not adding a related test case?

https://github.com/hexojs/hexo/blob/master/test/scripts/hexo/post.js

SukkaW:
```js
  // test for PR #4161
  it('render() - adjacent tags', () => {
    const content = [
      '{% quote %}',
      'content1',
      '{% endquote %}',
      '{% quote %}',
      'content2',
      '{% endquote %}'
    ].join('\n');

    return post.render(null, {
      content,
      engine: 'swig'
    }).then(data => {
      data.content.trim().should.eql([
        '<blockquote><p>content1</p>\n</blockquote>\n',
        '<blockquote><p>content2</p>\n</blockquote>\n',
      ].join(''));
    });
  });
```

Test case could be designed like this.

stevenjoezhang:
Thanks for reminding. In fact, when Tommy351 created the relevant code five years ago, the test examples given were incomplete: https://github.com/hexojs/hexo/commit/683fd0a19c4f94216dcc13c35970135d7a6ef03e

What's more interesting is that for multiple nested tags with the same name, the regular expression before modification does not actually match the last `{% end %}` tag. The intermediate results rendered by Hexo are confusing, but this does not affect nunjucks rendering correct results.

https://www.regextester.com/15

```regex
/\{% *(.+?)(?: *| +.*)%\}[\s\S]+?\{% *end\1 *%\}/g
```
```md
{% note danger %}
note text, note text, note text

{% note danger %}
note text, note text, note text
{% endnote %}

{% endnote %}
```

截屏2020-02-27上午12 46 02

I am still reading the source code to determine how Nunjucks preprocessing works. I'd appreciate it if you would help.

SukkaW:
@stevenjoezhang 

I noticed this regexp is only used to escape content:

https://github.com/hexojs/hexo/blob/master/lib/hexo/post.js#L48-L52

The whole post render process are looking like this:

1. Exec `before_post_render` filter

https://github.com/hexojs/hexo/blob/a97bd2f11e67ca2ffb6186b5d61521fcb58f4a1a/lib/hexo/post.js#L253

Hexo has built-in backtick code filter which will be executed at this time.

2. Escape swig tags into `<!--\uFFFC[index]-->`

https://github.com/hexojs/hexo/blob/a97bd2f11e67ca2ffb6186b5d61521fcb58f4a1a/lib/hexo/post.js#L259

3. Render post with proper engine. Since swig tag is escaped, there will be no extra `<p>` tags added.

4. After render is finished, restore escaped swig tag based on index:

https://github.com/hexojs/hexo/blob/a97bd2f11e67ca2ffb6186b5d61521fcb58f4a1a/lib/hexo/post.js#L282-L283

5. swig tag to be rendered:

https://github.com/hexojs/hexo/blob/a97bd2f11e67ca2ffb6186b5d61521fcb58f4a1a/lib/hexo/post.js#L288-L289

stevenjoezhang:
@SukkaW Thank you for the explanation. The problem is in the second step

> Escape swig tags into `<!--\uFFFC[index]-->`

https://github.com/hexojs/hexo/blob/a97bd2f11e67ca2ffb6186b5d61521fcb58f4a1a/lib/hexo/post.js#L42-L53

As I said earlier, `rSwigFullBlock` does not match the last `{% end %}` tag, which means that after executing `str.replace(rSwigFullBlock, escape)`, 

https://github.com/hexojs/hexo/blob/a97bd2f11e67ca2ffb6186b5d61521fcb58f4a1a/lib/hexo/post.js#L49

```md
{% note danger %}
note text, note text, note text

{% note danger %}
note text, note text, note text
{% endnote %}

{% endnote %}
```

becomes 

```md
<!-- \uFFFC 1 -->

{% endnote %}
```

`{% endnote %}`  is actually replaced by the next line

https://github.com/hexojs/hexo/blob/a97bd2f11e67ca2ffb6186b5d61521fcb58f4a1a/lib/hexo/post.js#L50

---

The test cases have been updated, but this part is a bit confusing to me.

SukkaW:
> Of course, this is not a bug, it's just a bit confusing.

https://runkit.com/sukkaw/5e57b5ead3f4440013a77529

I have set up a demo to show how it works. I believe it is ok merge this PR then.

PR #4164, opened 2020-02-29
Refactor(test): Useful sinon.assert

segayuu:
## How to test

```sh
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.


SukkaW:
IMHO, since Hexo has complicate unit tests, it is better to retain `test/index.js`. Also, using `*` might causes different behaviors between enviroments.

Issue #2821, opened 2017-10-30
BUG with "post_link" or "post_path" tag


gitforziio:
##### 1. BUG description

The tag `post_link` or `post_path` did not work correctly.

###### Code in one `.md` file

```markdown
## 硕士期间的其他工作

- [基于强化学习的生物行为模拟]({% post_path reinforce-learning-for-creature-behavior-simulation %})（Javascript）
- 心理学教学助手APP（界面设计、宣传视频制作）
- Psychtoolbox 心理学速成编程框架（MATLAB）
......
```

###### Meta of another `.md` file

```markdown
---
title: 基于强化学习的生物行为模拟
permalink: reinforce-learning-for-creature-behavior-simulation
tags:
- featured
categories:
- featured
---
```

###### Result

![p1](https://user-images.githubusercontent.com/10508116/32160523-101c1710-bd8d-11e7-8723-29307df1593b.png)
![p2](https://user-images.githubusercontent.com/10508116/32160524-1054a616-bd8d-11e7-80bd-58eeed29bb8d.png)

#####  3. Log with hexo <command> --debug

no unusual

```
07:49:33.322 INFO  Hexo is running at http://localhost:4000/. Press Ctrl+C to stop.
07:49:33.332 DEBUG Database saved
07:49:37.910 DEBUG Rendering post: posts/my-resume/index.html
GET /posts/my-resume/ 200 52.976 ms - -
GET /css/fonts/fontawesome-webfont.woff?v= 200 3.979 ms - -
GET /fancybox/jquery.fancybox.css 200 409.097 ms - -
GET /js/script.js 200 407.932 ms - -
GET /css/style.css 200 409.227 ms - -
GET /js/jquery.js 200 410.157 ms - -
GET /fancybox/jquery.fancybox.pack.js 200 408.515 ms - -
GET /css/images/banner.jpg 200 410.924 ms - -
```

#### Environment Info

Node version(`node -v`): `v8.4.0`

Hexo and Plugin version(`npm ls --depth 0`): 

```
hexo-site@0.0.0 /Users/SunCH/MyHexo/psyblog
├── hexo@3.3.9
├── hexo-deployer-git@0.3.1
├── hexo-generator-archive@0.1.4
├── hexo-generator-category@0.1.3
├── hexo-generator-index@0.2.1
├── hexo-generator-tag@0.2.0
├── hexo-renderer-ejs@0.3.1
[...]

Issue #4087, opened 2020-01-19
Markdown files are not rendered as expected


hujinbo:
## Check List

Please check followings before submitting a new issue.

- [x] I have already read [Docs page](https://hexo.io/docs/) & [Troubleshooting page](https://hexo.io/docs/troubleshooting)
- [x] I have already searched existing issues and they are not help  to me
- [x] I examined error or warning messages and it's difficult to solve
- [x] Using [the latest](https://github.com/hexojs/hexo/releases) version of Hexo (run `hexo version` to check)
- [x] Node.js is higher than 8.6.0

## Expected behavior
When I used the hexo-theme-next video tag and other tags in the same markdown file, only the video tag rendered successfully and the other tags cannot be rendered. But if you remove the video tag, it will render properly.

I reported the bug to hexo-theme-next, but they said it was a hexo problem.
https://github.com/theme-next/hexo-theme-next/issues/1353#issuecomment-575844786

## How to reproduce?

You can clone the source code of this blog for reproduction: https://github.com/hujinbo/hexo-next-test

* add post `tagPluginTest.md`
```
---
title: Hexo标签测试
---

## 居中引用

    {% cq %}你所做的事情，也许暂时看不到成功，但不要灰心，你不是没有成长，而是在扎根。{% endcq %}


显示效果如下所示：  

{% cq %}你所做的事情，也许暂时看不到成功，但不要灰心，你不是没有成长，而是在扎根。{% endcq %}


## 插入视频

    {% video https://cdn.hujinbo.me/video/demo.mp4 %}

显示效果如下所示：

{% video https://cdn.hujinbo.me/video/demo.mp4 %}
```
* run `hexo server`
* open http://localhost:4000

## Environment & Settings

**Node.js & npm version**

```
v10.16.0
6.9.0
```

**Your site `_config.yml`** (Optional)

```
# Hexo Configuration
## Docs: https://hexo.io/docs/configuration.html
## Source: https://github.com/hexojs/hexo/

# Site
title: Hexo
subtitle: ''
description: ''
keywords:
author: John Doe
language: en
timezone: ''

# URL
## If your site is put in a subdirectory, set url as 'http://yoursite.com/child' and root as '/child/'
url: http://yoursite.com
root: /
permalink: :year/:month/:day/:title/
permalink_defaults:
pretty_urls:
  trailing_index: true # Set to false to remove trailing 'index.html' from permalinks
  trailing_html: true # Set to false to remove trailing '.html' from permalinks

[...]

Issue #3543, opened 2019-05-03
{% raw %} tags shown when mixed triple backtick and raw blocks are used with hljs


phanan:
On vuejs.org, we're encountering a [weird issue](https://github.com/vuejs/vuejs.org/issues/1992) where `{% raw %}` and `{% endraw %}` tags are visible with a very specific set of settings. 

## What does it do?

To reproduce the bug with minimal settings.

## How to test

```sh
git clone -b raw-bug-poc https://github.com/phanan-forks/hexo.git
cd hexo
npm install
npm test
```

You'll notice that with the provided stub (`content`) AND `hexo.config.highlight.hljs` set to `true`, hexo will leave the `{% raw %}` and `{% endraw %}` tags untouched, and the (only) test case will fail with this assertion error:

```diff
+ expected - actual

<figure class="highlight js"><table><tr><td class="gutter"><pre><span class="line">1</span><br></pre></td><td class="code"><pre><code class="hljs js">alert(<span class="hljs-string">"Foo"</span>)<br></code></pre></td></tr></table></figure>
-{% raw %}
+
<p>Foo</p>
-{% endraw %}
+
<figure class="highlight html"><table><tr><td class="gutter"><pre><span class="line">1</span><br></pre></td><td class="code"><pre><code class="hljs html"><span class="hljs-tag">&lt;<span class="hljs-name">div</span>&gt;</span><span class="hljs-tag">&lt;/<span class="hljs-name">div</span>&gt;</span><br></code></pre></td></tr></table></figure>
```

## Screenshots

None.

## Pull request tasks

- [x] Add test cases for the changes.
- [ ] Passed the CI test.

phanan:
Original (failing) PR: #3459

seaoak:
I found that `lib/extend/tag.js` causes this issue.

https://github.com/hexojs/hexo/blob/master/lib/extend/tag.js#L118

```js
  str = str.replace(/<pre><code.*>[\s\S]*?<\/code><\/pre>/gm, escapeContent);
```

This code eats all of the `raw` block between two triple backtick code blocks.

Why does the option `highlight.hljs` affect this issue?

When `highlight.hljs=true`, the output of `highlight()` includes the string `<pre><code class="...`.
When `highlight.hljs=false`, the output includes `<span>` element instead of `<code>` element (`<pre><span class="...`).
This is because this issue is affected by the option `highlight.hljs`.

I investigate how correct this code (maybe just change RegExp to "non-greedy").

[...]

Issue #3346, opened 2018-11-08
{% %} should not be rendered in a post


yifei325325:
## Environment Info

Node version(`node -v`):

Your site `_config.yml` (Optional):

Your theme `_config.yml` (Optional): 

Hexo and Plugin version(`npm ls --depth 0`): 

## For BUG

比如新建一个名为“test.md”的文档，在里面包含以下这句话，
测试测试测试`{% test %}`测试测试
在使用hexo g 命令时就会报错，提示Template render error: (unknown path) 
而如果把上面的文字改成
测试测试测试
```
{% test %}
```
测试
这样使用hexo g 命令就会通过。

而在markdown 编辑器中 ，测试测试测试`{% test %}`测试测试 这样的写法是支持的，预览也不会出任何问题。

gautamz07:
I suppose your not suppose to have ejs code inside your .md file , are to trying to showcase a code snippet ?

yifei325325:
Thanks for your reply, now i try to explain my issue in English.
when i have a markdown file like test.md, all the markdown syntax passed the "hexo g" command. but when i include the statement like this 
-----------------------------------------
codes for example `{% test code %}` code for example
------------------------------------------
ignores the dash please. the two " ` " means ,there is a code inside.

when i try to use "hexo g" command to compile the file to html file. it shows the error like "Template render error: (unknown path) ". But when i delete this statement, everything goes fine.
How odd!

yifei325325:
but when i write the code like this
![image](https://user-images.githubusercontent.com/9191218/48198185-05b13f80-e393-11e8-97e1-8b330d36a865.png)
use the "hexo g" command , every thing goes well.  @gautamz07

YoshinoriN:
Same problem reported at (#2515 )
But, it seems solved (#2461)

We need investigation.

YoshinoriN:
Yup, I know. I reproduce it v3.8.0
I think, this problem is a regression bug or not fixed in #2461

yifei325325:
> Yup, I know. I reproduce it v3.8.0
> I think, this problem is a regression bug or not fixed in #2461

Thanks, I got it. hope it will be fixed on next release version.

52Heartz:
我今天也遇到了这个问题，我对问题的理解以及解决方法如下：

[...]

PR #4171, opened 2020-03-04
refactor(post/tag): render tag before content

SukkaW:
## What does it do?

Continuation of https://github.com/hexojs/hexo/issues/3543#issuecomment-569865652 and https://github.com/hexojs/hexo/pull/4161

Currently post rendering process is something like this: https://github.com/hexojs/hexo/pull/4161#issuecomment-591802138. But basically, hexo will render markdown first, then the tag (using nunjucks). So hexo has to remove all swig tags before markdown rendering, and restore them after markdown rendering is finished.

In this PR, I make hexo renders tag first, then markdown, so that remove and restore swig tag are not necessary any more, resulted in 6% faster in generation speed. And the post rendering related test cases are passed as well.

I wonder why Hexo has to render tag after markdown at the beginning.

## How to test

```sh
git clone -b render-tag-before-md https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [ ] Passed the CI test.

Issue resolved: https://github.com/hexojs/hexo/issues/2821

### DIFF
--- a/lib/extend/tag.js
+++ b/lib/extend/tag.js
@@ -77,7 +77,7 @@ class NunjucksBlock extends NunjucksTag {
   }
 
   run(context, args, body) {
-    return this._run(context, args, trimBody(body));
+    return this._run(context, args, trimBody(body)) + '\n';
   }
 }
 
@@ -111,7 +111,7 @@ class NunjucksAsyncBlock extends NunjucksBlock {
       body = () => result || '';
 
       this._run(context, args, trimBody(body)).then(result => {
-        callback(err, result);
+        callback(err, result && `${result}\n`);
       });
     });
   }
--- a/lib/extend/tag.js
+++ b/lib/extend/tag.js
@@ -109,7 +109,7 @@ class NunjucksAsyncBlock extends NunjucksBlock {
       body = () => result || '';
 
       this._run(context, args, trimBody(body)).then(result => {
-        callback(err, result && `${result}\n`);
+        callback(err, `${result}\n`);
       });
     });
   }
--- a/lib/extend/tag.js
+++ b/lib/extend/tag.js
@@ -4,8 +4,6 @@ const stripIndent = require('strip-indent');
 const { cyan } = require('chalk');
 const nunjucks = require('nunjucks');
 const Promise = require('bluebird');
-const placeholder = '\uFFFC';
-const rPlaceholder = /(?:<|&lt;)!--\uFFFC(\d+)--(?:>|&gt;)/g;
 
 class NunjucksTag {
   constructor(name, fn) {
@@ -228,15 +226,8 @@ class Tag {
       options = {};
     }
 
-    const cache = [];
-
-    const escapeContent = str => `<!--${placeholder}${cache.push(str) - 1}-->`;
-
-    str = str.replace(/<pre><code.*>[\s\S]*?<\/code><\/pre>/gm, escapeContent);
-
     return Promise.fromCallback(cb => { this.env.renderString(str, options, cb); })
-      .catch(err => Promise.reject(formatNunjucksError(err, str)))
-      .then(result => result.replace(rPlaceholder, (_, index) => cache[index]));
+      .catch(err => Promise.reject(formatNunjucksError(err, str)));
   }
 }
 
--- a/lib/hexo/post.js
+++ b/lib/hexo/post.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const assert = require('assert');
 const moment = require('moment');
 const Promise = require('bluebird');
 const { join, extname } = require('path');
@@ -11,29 +10,6 @@ const { copyDir, exists, listDir, mkdirs, readFile, rmdir, unlink, writeFile } =
 const yfm = require('hexo-front-matter');
 
 const preservedKeys = ['title', 'slug', 'path', 'layout', 'date', 'content'];
-const rEscapeContent = /<escape(?:[^>]*)>([\s\S]*?)<\/escape>/g;
-const rPlaceholder = /(?:<|&lt;)!--hexoPostRenderEscape:(\d+)--(?:>|&gt;)/g;
-
-class PostRenderCache {
-  constructor() {
-    this.cache = [];
-  }
-
-  escapeContent(str) {
-    return str.replace(rEscapeContent, (_, content) => `<!--hexoPostRenderEscape:${this.cache.push(content) - 1}-->`);
-  }
-
-  loadContent(str) {
-    const restored = str.replace(rPlaceholder, (_, index) => {
-      assert(this.cache[index]);
-      const value = this.cache[index];
-      this.cache[index] = null;
-      return value;
-    });
-    if (restored === str) return restored;
-    return this.loadContent(restored); // self-recursive for nexted escaping
-  }
-}
 
 const prepareFrontMatter = data => {
   for (const [key, item] of Object.entries(data)) {
@@ -230,17 +206,12 @@ class Post {
     // disable Nunjucks when the renderer spcify that.
     const disableNunjucks = ext && ctx.render.renderer.get(ext) && !!ctx.render.renderer.get(ext).disableNunjucks;
 
-    const cacheObj = new PostRenderCache();
-
     return promise.then(content => {
       data.content = content;
 
       // Run "before_post_render" filters
       return ctx.execFilter('before_post_render', data, { context: ctx });
     }).then(() => {
-      // Escape content wrapped in <escape></escape>
-      data.content = cacheObj.escapeContent(data.content);
-
       if (isSwig) {
         // Render with Nunjucks if the post is a swig file
         return tag.render(data.content, data);
@@ -256,9 +227,8 @@ class Post {
         engine: data.engine,
         toString: true,
         onRenderEnd(content) {
-          // Replace cache data with real contents
-          data.content = cacheObj.loadContent(content);
-
+          data.content = content.replace(/<!--hexoPostRenderEscape:/g, '')
+            .replace(/:hexoPostRenderEscape-->/g, '');
           return data.content;
         }
       }, options);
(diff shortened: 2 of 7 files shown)


lib/hexo/post.js
```diff
@@ -11,24 +11,19 @@ const { copyDir, exists, listDir, mkdirs, readFile, rmdir, unlink, writeFile } =
 const yfm = require('hexo-front-matter');
 
 const preservedKeys = ['title', 'slug', 'path', 'layout', 'date', 'content'];
-
-const _escapeContent = (cache, str) => {
-  const placeholder = '\uFFFC';
-  return `<!--${placeholder}${cache.push(str) - 1}-->`;
-};
+const rEscapeContent = /<escape(?:[^>]*)>([\s\S]*?)<\/escape>/g;
+const rPlaceholder = /(?:<|&lt;)!--hexoPostRenderEscape:(\d+)--(?:>|&gt;)/g;
```
SukkaW: Since `\uFFFC` will be rendered by Nunjucks, I have to change it to `hexoPostRenderEscape:`

lib/hexo/post.js
```diff
@@ -254,46 +236,45 @@ class Post {
...
+      return ctx.execFilter('before_post_render', data, { context: ctx });
     }).then(() => {
+      // Escape content wrapped in <escape></escape>
       data.content = cacheObj.escapeContent(data.content);
 
       if (isSwig) {
-        // Render with Nunjucks if this is a swig file
+        // Render with Nunjucks if the post is a swig file
         return tag.render(data.content, data);
       }
```
stevenjoezhang: Pending #2903 ? `isSwig` will be removed in this pull request

SukkaW: That doesn't matter. If one of two PRs is merged first, then the other will merge and rebased.

SukkaW:
I am considering fixing #3346 in this PR as well since that issue is also related with post rendering process.

stevenjoezhang:
```md
{% pullquote warning %}
Text
{% endpullquote %}
# Title 0
{% pullquote warning %}
Text
{% endpullquote %}

{% pullquote warning %}
Text
{% endpullquote %}

# Title 1

{% pullquote warning %}
Text
{% endpullquote %}

{% pullquote warning %}Text{% endpullquote %}
# Title 2
{% pullquote warning %}Text{% endpullquote %}

{% pullquote warning %}Text{% endpullquote %}

# Title 3

{% pullquote warning %}Text{% endpullquote %}
```

截屏2020-03-04下午5 21 23


The bug reported in #4161 was reproduced. I am confirming if this is a bug (or feature) of Nunjucks
https://github.com/mozilla/nunjucks/blob/40dfdf0ea3e6ba43c3a180eca57d935c36d188c2/nunjucks/src/parser.js#L640-L649

lib/plugins/filter/before_post_render/backtick_code_block.js
```diff
@@ -63,7 +63,7 @@ function backtickCodeBlock(data) {
       .replace(/{/g, '&#123;')
       .replace(/}/g, '&#125;');
 
-    return `${start}<escape>${content}</escape>${end}`;
+    return `${start}<!--hexoBacktickCodeBlockEscape:${content}:hexoBacktickCodeBlockEscape-->${end}`;
```
SukkaW: Let me show you an example:

The post content is:

```markdown
# Title

[code block]
some content
```

Before this PR, those content will escaped to:

```markdown
# Title

<!--\uFFFc1-->
some content
```

So markdown renderer will rendered it into:

```html
<h1>Title</h1>

<!--\uFFFc1-->
<p>some content</p>
```

(Code block escaped as comment will be treated as a blank line, so next line will be wrapped in `<p>`).

But if we simply remove the escaping, there will be no new line before `some content`:

```markdown
# Title

[code block]
some content
```

So it will be rendered into:

```html
<h1>Title</h1>

<!--\uFFFc1-->
some content
```

No `<p>`, which caused unit test to fail.

That's why I bring up this workaround here.

stevenjoezhang: I see, it's different from #4161. This is the behavior of the markdown renderer, not Nunjucks.

SukkaW: @stevenjoezhang 

However, since https://github.com/hexojs/hexo/pull/4171#issuecomment-594412367 is still there, we have to find out a way to solve this.

seaoak:
Thank you for this great patch.

> I wonder why Hexo has to render tag after markdown at the beginning.

Me, too. :wink:

---

> However, since https://github.com/hexojs/hexo/pull/4171#issuecomment-594412367 is still there, we have to find out a way to solve this.

Generally speaking, in Markdown, any *block* content is separated by empty line(s) each other.

And custom tags in Nunjucks is not intended for Markdown block content only.
So Nunjucks does not generate additional newline characters by default.

That is,

```
{% pullquote warning %}
Text
{% endpullquote %}
# Title 0
{% pullquote warning %}
Text
{% endpullquote %}
```

should be written as folows:

```
{% pullquote warning %}
Text
{% endpullquote %}

# Title 0

{% pullquote warning %}
Text
{% endpullquote %}
```

If we want to add a newline character *automatically* for Nunjucks custom tags in Hexo,
the codes in `lib/extend/tags.js` should be modified as follows:

`````diff
diff --git a/lib/extend/tag.js b/lib/extend/tag.js
index 31f9f2aa..883fdb5f 100644
--- a/lib/extend/tag.js
+++ b/lib/extend/tag.js
@@ -77,7 +77,7 @@ class NunjucksBlock extends NunjucksTag {
   }

   run(context, args, body) {
-    return this._run(context, args, trimBody(body));
+    return this._run(context, args, trimBody(body)) + '\n';
   }
 }

@@ -111,7 +111,7 @@ class NunjucksAsyncBlock extends NunjucksBlock {
       body = () => result || '';

       this._run(context, args, trimBody(body)).then(result => {
-        callback(err, result);
+        callback(err, result && `${result}\n`);
       });
     });
   }
`````

I'm not sure whether this behavior is desirable in Hexo.
Even so, this modification should be independent of this patch.

**Comment by SukkaW** [comment]
> Even so, this modification should be independent of this patch.

@segayuu 

IMHO, it is fine to include this patch inside this PR since it is related with post rendering issue fix.

File: lib/extend/tag.js [hunk 4]
```diff
@@ -228,15 +226,8 @@ class Tag {
       options = {};
     }
 
-    const cache = [];
-
-    const escapeContent = str => `<!--${placeholder}${cache.push(str) - 1}-->`;
-
-    str = str.replace(/<pre><code.*>[\s\S]*?<\/code><\/pre>/gm, escapeContent);
```
SukkaW: Hexo only use `tag.render` in `post.js`, which is after `backtick_code_filter` is executed.

In `backtick_code_filter`, `{` & `}` are already escaped and it is fine to render with nunjucks directly.

seaoak:
I'm sorry I'll be too worried.

This line will be simpler because `result` must be a string.

```diff
diff --git a/lib/extend/tag.js b/lib/extend/tag.js
@@ -111,7 +111,7 @@ class NunjucksAsyncBlock extends NunjucksBlock {
       body = () => result || '';

       this._run(context, args, trimBody(body)).then(result => {
-        callback(err, result && `${result}\n`);
+        callback(err, `${result}\n`);
       });
     });
   }
```

SukkaW:
> This line will be simpler because `result` must be a string.

Updated.

lib/extend/tag.js
```diff
@@ -77,7 +75,7 @@ class NunjucksBlock extends NunjucksTag {
   }
 
   run(context, args, body) {
-    return this._run(context, args, trimBody(body));
+    return this._run(context, args, trimBody(body)) + '\n';
```
jiangtj: Why add `\n`?

Look at the following test cases, which one is expected?
```js
    const str = [
      'begin',
      '{% test foo bar %}',
      'test content',
      '{% endtest %}',
      'end'
    ].join('\n');
    const result = await tag.render(str);

    // there is only one `\n` before and after
    result.should.eql('begin\nfoo bar test content\nend');
    // Or should there be two `\n` at the end?
    result.should.eql('begin\nfoo bar test content\n\nend');
```

stevenjoezhang:
Another issue: https://github.com/hexojs/hexo/issues/4087
Indented code blocks will be parsed and rendered by Nunjucks.

SukkaW:
@stevenjoezhang 

For indented code blocks there are no better ways. For backtick code blocks we already have a filter.

seaoak:
@SukkaW @stevenjoezhang

To escape Nunjucks tags in indented code blocks properly,
we might need a new `before_post_render` filter (same as backtick code blocks)

I try to make a patch of such a filter:
seaoak/hexo:feature/escape_indented_code_block_before_render
https://github.com/seaoak/hexo/tree/feature/escape_indented_code_block_before_render

With merging it, this @SukkaW 's patch resolves the issue #4087.

Do you think about this new filter?
If it is useful, I continue to try to make it. :smile:

SukkaW:
> To escape Nunjucks tags in indented code blocks properly,
> we might need a new `before_post_render` filter (same as backtick code blocks)

I am rather not to do that.

Only escape `{` and `}` which is inside code block, and after code block is highlighted.

lib/extend/tag.js
seaoak: This extra `\n` is to separete a Nunjucks tag from immediate-after content.
So I expect *two* `\n` at the end.

Please refer the above comment https://github.com/hexojs/hexo/pull/4171#issuecomment-595027051

SukkaW:
@seaoak 

From #4179 we might consider it is necessary for hexo to render markdown first, right?

seaoak:
@SukkaW Yes.  I think so.

SukkaW:
@seaoak So we should then focus on how to make swig escaping quicker.

PR #4179, opened 2020-03-11
Escape indented code block before render

seaoak:
## What does it do?

Supplement of the PR #4171, for the issue #4087.

Since the PR #4171 changes rendering order,
Nunjucks tags in indented code blocks become targets of rendering by Nunjucks.
This is undesirable.

This situation can be observed in the issue #4087.

This patch escapes Nunjucks tags in indented code blocks before rendering by Nunjucks.

## How to test

```sh
git clone -b feature/escape_indented_code_block_before_render https://github.com/seaoak/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [ ] Passed the CI test.

### DIFF
--- a/lib/plugins/filter/before_post_render/indented_code_block.js
+++ b/lib/plugins/filter/before_post_render/indented_code_block.js
@@ -2,10 +2,10 @@
 
 const { escapeHTML, stripIndent } = require('hexo-util');
 
-const rIndentedCodeBlock = /(^|(?:^|\n)(?:(?: {0,3}>){1,3}[^\S\r\n])?[^\S\r\n]*\n)((?:((?:(?: {0,3}>){1,3}[^\S\r\n])?)( {4}|\t)([^\n]*?)(\n|$))+)/g;
+const rIndentedCodeBlock = /(^|(?:^|\n)(?:(?: {0,3}>){1,3}[^\S\r\n])?[^\S\r\n]*)((?:(?:^|\n)((?:(?: {0,3}>){1,3}[^\S\r\n])?)( {4}|\t)([^\n]*))+)(?=\n|$)/g;
 
 function indentedCodeBlock(data) {
-  data.content = data.content.replace(rIndentedCodeBlock, ($0, before, content, start, $4, $5, end) => {
+  data.content = data.content.replace(rIndentedCodeBlock, ($0, before, content, start, $4, $5) => {
 
     // PR #3765
     if (start.includes('>')) {
@@ -25,7 +25,7 @@ function indentedCodeBlock(data) {
 
     const markedContent = '<!--hexoPostRenderEscape:' + wrappedContent + ':hexoPostRenderEscape-->';
 
-    return `${before}${start}${markedContent}${end}`;
+    return `${before}${start}${markedContent}\n`;
   });
 }
 
--- a/lib/plugins/filter/before_post_render/indented_code_block.js
+++ b/lib/plugins/filter/before_post_render/indented_code_block.js
@@ -3,7 +3,7 @@
 const stripIndent = require('strip-indent');
 const { highlight } = require('hexo-util');
 
-const rIndentedCodeBlock = /^(?:((?:[^\S\r\n]{0,3}>){0,3}[^\S\r\n]*?)( {4}|\t)([^\n]*?)(\n|$))+/gm;
+const rIndentedCodeBlock = /^(?:((?: {0,3}>){0,3}[^\S\r\n])( {4}|\t)([^\n]*?)(\n|$))+/gm;
 
 function indentedCodeBlock(data) {
   const config = this.config.highlight || {};
--- a/lib/plugins/filter/before_post_render/indented_code_block.js
+++ b/lib/plugins/filter/before_post_render/indented_code_block.js
@@ -18,13 +18,14 @@ function indentedCodeBlock(data) {
       .replace(/{/g, '&#123;')
       .replace(/}/g, '&#125;');
 
-    const wrappedContent = `<pre><code>${escapedContent}</code></pre>`;
+    // hack to guard blank lines from incorrect rendering
+    const guardedContent = escapedContent.replace(/^/mg, start).replace(start, '');
 
-    const guardedContent = '<!--hexoPostRenderEscape:'
-        + wrappedContent.replace(/^/mg, start) // hack to guard blank lines from incorrect rendering
-        + ':hexoPostRenderEscape-->';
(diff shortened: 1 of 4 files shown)


seaoak:
Code highlighting might be unnecessary for indented code blocks.

SukkaW:
> Code highlighting might be unnecessary for indented code blocks.

Yeah, there's no way for indented code blocks to add language mark.
Simply wrapped it inside `<pre><code> </code></pre>` will be ok then.

lib/plugins/filter/before_post_render/indented_code_block.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+const stripIndent = require('strip-indent');
```
SukkaW: `strip-indent` has been removed.

```suggestion
const { stripIndent } = require('hexo-util');
```

seaoak: Thank you for your advice (just I was bothering!)
I updated.

lib/plugins/filter/before_post_render/indented_code_block.js
```diff
@@ -0,0 +1,28 @@
...
+      content = (content + paddingOnEnd).replace(regexp, '').replace(/\n$/, '');
+    }
+
+    content = escapeHTML(stripIndent(content))
+      .replace(/{/g, '&#123;')
+      .replace(/}/g, '&#125;');
+
+    content = `<pre><code>${content}</code></pre>`;
+
+    return `${start}<!--hexoPostRenderEscape:${content}:hexoPostRenderEscape-->${end}`;
```
SukkaW:
```suggestion
return start
  + '<!--hexoPostRenderEscape:<pre><code>'
  + escapeHTML(stripIndent(content))
    .replace(/{/g, '&#123;')
    .replace(/}/g, '&#125;')
  + '</code></pre>:hexoPostRenderEscape-->'
  + end;
```

SukkaW: Avoid unnecessary `content` reassignment without breaking code readability.

SukkaW:
```
test test test
  hello world
test test
  hello
  world
test test
```

```
- a
  - b
    - c
      - d
```

What about adding test cases above to see if it can handle the situation.

lib/plugins/filter/before_post_render/indented_code_block.js
seaoak: I agree your advice about avoiding reassignment.
But I think it is also important that the code shows the process step-by-step.
So I choice to use new variables instead of reassignment.

seaoak:
@SukkaW I updated.
Your test cases are imported into test codes.  Thank you!

SukkaW:
> Your test cases are imported into test codes. Thank you!

@seaoak I get a better idea.
Since it is hard for a regexp to do `marked.js`'s job (to identify which line is nested list, which line is indented code block). What about not to wrapped code block inside `<pre><code></code></pre>` (this should be done by markdown renderer), only escape `{` & `}` then.

stevenjoezhang:
Users may use tags in nested list
```
- a
  - b
    - c
      - {% post_link 'hello world' %}
```

seaoak:
As @stevenjoezhang 's commnet, @SukkaW 's idea can not be used.
But I tried to make many test cases, the regexp will be OK.

SukkaW:
> But I tried to make many test cases, the regexp will be OK.

They might be many other cases that should be covered:

```
- a
  - b
    * c
      1. a
    * b
```

```
> 1. a
>   - b
>     1. c
```

And if the regexp is over complicated, the performance might be affected.

seaoak:
The condition to recognize "indented code block" is not so complicated.
It is whether the preceding line is a blank line or not.

Only in blockquote, the condition is a bit more complicated.

I think it is acceptable.

lib/plugins/filter/before_post_render/indented_code_block.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
+
+const { escapeHTML, stripIndent } = require('hexo-util');
+
+const rIndentedCodeBlock = /(^|(?:^|\n)(?:(?: {0,3}>){1,3}[^\S\r\n])?[^\S\r\n]*\n)((?:((?:(?: {0,3}>){1,3}[^\S\r\n])?)( {4}|\t)([^\n]*?)(\n|$))+)/g;
```
SukkaW:
```suggestion
const rIndentedCodeBlock = /(^|(?:^|\n)(?:(?: {0,3}>){1,3}[^\S\r\n])?[^\S\r\n]*\n)((?:((?:(?: {0,3}>){1,3}[^\S\r\n])?)( {4}|\t)([^\n]*?)(\n|$))+)/gm;
```

Should `gm` to be used?

```
test test

    hello world

    hello world

test test
```

There is a case that second `hello world` won't be matched.

lib/plugins/filter/before_post_render/indented_code_block.js
```diff
@@ -0,0 +1,32 @@
...
+
+function indentedCodeBlock(data) {
+  data.content = data.content.replace(rIndentedCodeBlock, ($0, before, content, start, $4, $5, end) => {
```
SukkaW:
```suggestion
  # Return early if there is no 4 consecutive spaces
  if (!data.content.includes('    ')) return;
  data.content = data.content.replace(rIndentedCodeBlock, ($0, before, content, start, $4, $5, end) => {
```

lib/plugins/filter/before_post_render/indented_code_block.js
seaoak: With the option `m`, indented line becomes target even if its preceding line is not a blank line.

I adjusted regexp, and add test case.

lib/plugins/filter/before_post_render/indented_code_block.js
seaoak: To support single-tab-indented code block, this optimization can not be used.

seaoak:
I Updated.

seaoak:
Unfortunately, I awake to the danger that this filter spoils indented codes in code block tag.

For example,

```
{% codeblock lang:javascript %}
if (a) {
  if (b) {
    if (c) {
      doSomething1();

      doSomething2();
    }
  }
}
{% endcodeblock %}
```

this filter escapes the line 7-8 with `<pre><code>...</code></pre>`.
This is undesirable.

I have no idea to resolve this. :cry:

seaoak:
I'll discard this patch.

Issue #4181, opened 2020-03-11
`npm test` failed for toc helper


seaoak:
## Check List

Please check followings before submitting a new issue.

- [x] I have already read [Docs page](https://hexo.io/docs/) & [Troubleshooting page](https://hexo.io/docs/troubleshooting)
- [x] I have already searched existing issues and they are not help  to me
- [x] I examined error or warning messages and it's difficult to solve
- [x] Using [the latest](https://github.com/hexojs/hexo/releases) version of Hexo (run `hexo version` to check)
- [x] Node.js is higher than 8.6.0

## Expected behavior

Unit test of Hexo passes.

```bash
$ git clone https://github.com/hexo/hexo.git
$ cd hexo
$ npm install
$ npm test
```

## Actual behavior

`npm test` failed for two items.
Both of them are for toc helper.

![SS20200311a_Hexoのtestがエラーになる](https://user-images.githubusercontent.com/1160250/76408871-fe5bd500-63d0-11ea-9494-1c4867d8e4bb.png)

## How to reproduce?

always.

## Environment & Settings

hexo master 0fd57368484886c9cb43ba7e44af4c75e0a08988

hexo-util master/1.9.0 595aaab524d289d573e627005e651465e9f2a72e

node v12.16.1

PR #4125, opened 2020-02-13
perf(open_graph): avoid using htmlTag() and enhance cache

SukkaW:
## What does it do?

In 8bd3439421e2b50799d13e1ba3f804d599c94962 @curbengh utilize `open_graph()` helper with new `htmlTag()` util.

However, [`htmlTag()`](https://github.com/hexojs/hexo-util/blob/7e084de7244a8f11ff8571fae8a5782936ea65b0/lib/html_tag.js) from hexo-util is a little over complicated, thus I bring up this PR with avoiding usage of `htmlTag()`. 

## How to test

```sh
git clone -b optimize-opengraph-helper https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Screenshots

Benchmark Result after https://github.com/hexojs/hexo/pull/4125/commits/d13ab9a3c816b76ce092f83cd54ccbd96164fe68:

| | NodeJS 8 | NodeJS 10 | NodeJS 12 | NodeJS 13 |
|:---:|:---:|:---:|:---:|:---:|
| Current Master Branch | 17.87 | 14.83 | 15.91 | 14.77 |
| This PR | **16.18** | **14.32** | **15.23** | **14.6** |

There is about 3% - 5% performance improvements.

---

Benchmark Result after https://github.com/hexojs/hexo/pull/4125/commits/095d28e0a5012ddd17bc67ca8970ccbc1ff12809

| | NodeJS 8 | NodeJS 10 | NodeJS 12 | NodeJS 13 |
|:---:|:---:|:---:|:---:|:---:|
| Current Master Branch | 17.87 | 14.83 | 15.91 | 14.77 |
| This PR | **16.38** | **15.12** | **14.32** | **14.52** |

## Pull request tasks

- [x] Add test cases for the changes.
- [x] Passed the CI test.

PR #4182, opened 2020-03-11
Fix mismatching between TOC helper and `hexo-util.htmlTag`

seaoak:
## What does it do?

Fix the issue #4181.

Unit test of Hexo fails for two items of TOC helper function.
This is caused by mismatching between TOC helper and `hexo-util.htmlTag`.
(I guess it only by test case and source codes.)

## How to test

```sh
git clone -b bugfix/test_failed_for_toc_helper https://github.com/seaoak/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/helper/toc.js
+++ b/lib/plugins/helper/toc.js
@@ -25,7 +25,7 @@ function tocHelper(str, options = {}) {
   for (let i = 0, len = data.length; i < len; i++) {
     const el = data[i];
     const { level, id, text } = el;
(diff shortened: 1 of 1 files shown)


SukkaW:
What about avoid using `htmlTag`? As I noticed in PR #4125, by avoid using `htmlTag()` resulted in improved performance?

seaoak:
@SukkaW I agreed.

This PR is integrated into the PR #4183.

PR #4208, opened 2020-03-30
perf(meta_generator): avoid unnecessary check

SukkaW:
## What does it do?

This PR is mean to supersede #4070 and #4078.

With cached value `NEED_INJECT`, all the checks will be only executed once.

## How to test

```sh
git clone -b meta_generator_optimize https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [ ] Passed the CI test.


SukkaW:
According to the benchmark, the generation duration dropped by 3% after this PR.

Issue #4190, opened 2020-03-18
代码块渲染错误


fuyitaoqqq:
## Check List

Please check followings before submitting a new issue.

- [x] I have already read [Docs page](https://hexo.io/docs/) & [Troubleshooting page](https://hexo.io/docs/troubleshooting)
- [x] I have already searched existing issues and they are not help to me
- [x] I examined error or warning messages and it's difficult to solve
- [x] Using [the latest](https://github.com/hexojs/hexo/releases) version of Hexo (run `hexo version` to check)
- [x] Node.js is higher than 8.6.0

## Environment & Settings

**Node.js & npm version**

```
Node.js v12.16.1
npm 6.13.4
```

**Your site `_config.yml`** (Optional)

```
```

**Your theme `_config.yml`** (Optional)

```
```

**Hexo and Plugin version(`npm ls --depth 0`)**

```
hexo-site@0.0.0 /opt/apps/blog
├── hexo@4.2.0
├── hexo-generator-archive@1.0.0
├── hexo-generator-category@1.0.0
├── hexo-generator-feed@2.2.0
├── hexo-generator-index-pin-top@0.2.2
├── hexo-generator-searchdb@1.2.0
├── hexo-generator-tag@1.0.0
├── hexo-renderer-ejs@1.0.0
├── hexo-renderer-marked@2.0.0
├── hexo-renderer-stylus@1.1.0
└── hexo-server@1.0.0
```

**Your package.json `package.json`**

```
{
  "name": "hexo-site",
  "version": "0.0.0",
  "private": true,
  "scripts": {
    "build": "hexo generate",
    "clean": "hexo clean",
    "deploy": "hexo deploy",
    "server": "hexo server"
  },
  "hexo": {
    "version": "4.2.0"
  },
  "dependencies": {
    "hexo": "^4.0.0",
    "hexo-generator-archive": "^1.0.0",
    "hexo-generator-category": "^1.0.0",
    "hexo-generator-feed": "^2.2.0",
    "hexo-generator-index-pin-top": "^0.2.2",
    "hexo-generator-searchdb": "^1.2.0",
    "hexo-generator-tag": "^1.0.0",
    "hexo-renderer-ejs": "^1.0.0",
    "hexo-renderer-marked": "^2.0.0",
    "hexo-renderer-stylus": "^1.1.0",
    "hexo-server": "^1.0.0"
  }
}
```

## Others

md文件内容
[...]

Issue #4220, opened 2020-04-05
parsing code error in  backticks


xxxxnnxxxx:
```{{7*7}}``` ---> 49
`{{7*7}}` ---> 49

this like a template injection
![hexo_code](https://user-images.githubusercontent.com/12854812/78468138-3f80a400-7747-11ea-89f6-d6e2e3a11f7f.png)
![hexo_display](https://user-images.githubusercontent.com/12854812/78468140-44455800-7747-11ea-8131-a8b5b1d0def4.png)
![hexo-version](https://user-images.githubusercontent.com/12854812/78468172-a8681c00-7747-11ea-9630-fc5eb4ae500b.png)

SukkaW:
@stevenjoezhang @seaoak 

It seems that the swig tag inside backtick code filter is not properly escaped. Should we bring up a new filter to fix this?

stevenjoezhang:
I think it is necessary to discuss this issue. Here's another one about backtick code filter

https://github.com/hexojs/hexo/issues/4190

seaoak:
I'm trying to make a small patch for this issue (and the issue #4190).

https://github.com/hexojs/hexo/pull/4229

@SukkaW @stevenjoezhang 
Could you taste it?

PR #4229, opened 2020-04-08
Bugfix/parsing code error in backticks

seaoak:
## What does it do?

Fix the issue #4190 and #4220.

Current `backtick_code_block` filter does not consider a Swig template
which form is a pair of triple backticks in single line.

This patch makes the filter to ignore such lines.

And also the filter mistook triple-backtick as the end of targets
even if there are some extra characters at the head of line.

This patch fixes it too.

## How to test

```sh
git clone -b bugfix/parsing_code_error_in_backticks https://github.com/seaoak/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [ ] Passed the CI test.

### DIFF
--- a/lib/plugins/filter/before_post_render/backtick_code_block.js
+++ b/lib/plugins/filter/before_post_render/backtick_code_block.js
@@ -2,14 +2,15 @@
 
 const { highlight } = require('hexo-util');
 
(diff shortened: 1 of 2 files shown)


SukkaW:
What about adding #4190,#4220 related test cases?

seaoak:
@SukkaW 
I only rebase (squash) this branch.  Please review.

SukkaW:
LGTM then! I am going to merge this after I triggering a rebuild at travis.

Issue #3818, opened 2019-10-29
Drop NodeJs 8 support & Require NodeJs >= 10


tomap:
This issue is very similar to #3508 
We will need to drop NodeJs 8 Support as it reached EOL on December 31st
See https://github.com/nodejs/Release#release-schedule

It will also bring the benefit to be abe to use more native APIs.
Here is a list of issues/PR pending NodeJs 8 drop:
* https://github.com/hexojs/hexo-util/pull/116

Note that in most cases, it will imply a major version bump
We also will need to check which version to use (10.0, 10.1, ...?)

SukkaW:
https://github.com/hexojs/hexo-util/pull/114

After Node.js 8 is dropped. we can mitigate `decodeURL(str)` to newest whatwg url api, and remove `const { URL } = require('url');` from mitigated helper.

SukkaW:
Also, according to the performance comparison by https://blog.kuzzle.io/bluebird-vs-native-vs-async/await-state-of-promises-performances-in-2019 , async / await looks promising on Node.js 10 / 12. We might start to implement #3328 after Nodejs.8 is dropped.


![image](https://user-images.githubusercontent.com/40715044/67767422-862cec80-fa8b-11e9-885c-54ca1ad0e168.png)
![image](https://user-images.githubusercontent.com/40715044/67767429-888f4680-fa8b-11e9-8d6a-81ca2c41336f.png)

curbengh:
Much like our [previous](https://github.com/hexojs/hexo/issues/3508) dropping of Node 6, since we're developing solely on master branch (not version branch), we could only drop Node 8 (in the master branch, not yet in npm) when we're sure that there will be no more minor version bump (i.e. 4.x).

My idea is that we can work on anything that doesn't break Node 8 for now ([async/await](https://github.com/hexojs/hexo/issues/3328) as mentioned here is an example), even if it requires some minor workaround (e.g. https://github.com/hexojs/hexo-util/pull/117); those workarounds can be removed later once we drop Node 8.

As in timeline, if we're sure there'll be no more 4.x, let say in December, we can even start dropping Node 8 at that time (again, just in repo, not npm). But that's assuming we're not in :santa: mood...

Since Node 8 is quite modern, as in it even supports some [ES2019](https://node.green/#ES2019) features, I don't expect the transition Node 8 -> 10 to be as drastic as Node 6 -> 8, hopefully it's smoother this time.

[...]

Issue #4239, opened 2020-04-11
Setting timezone in config.yml doesn't work as expected


alexschneider:
## Check List

Please check followings before submitting a new issue.

- [x] I have already read [Docs page](https://hexo.io/docs/) & [Troubleshooting page](https://hexo.io/docs/troubleshooting)
- [x] I have already searched existing issues and they are not help  to me
- [n/a] I examined error or warning messages and it's difficult to solve
- [x] Using [the latest](https://github.com/hexojs/hexo/releases) version of Hexo (run `hexo version` to check)
- [x] Node.js is higher than 8.6.0

## Expected behavior
Timezone gets set to UTC

## Actual behavior
Timezone remains local time

## How to reproduce?

* Set timezone to `UTC` in _config.yml
* Change date_format to `date_format: YYYY-MM-DDTHH:mm:ss.SSSZ`
* Write a post using a timezone
* When I load it up, the timezone is `-0700` for me as I am currently in PDT, expect `Z` or `+0000`

## Environment & Settings

**Node.js & npm version**

```
> hexo version
hexo: 4.2.0
hexo-cli: 3.1.0
os: Windows_NT 10.0.18362 win32 x64
node: 13.7.0
v8: 7.9.317.25-node.28
uv: 1.34.1
zlib: 1.2.11
brotli: 1.0.7
ares: 1.15.0
modules: 79
nghttp2: 1.40.0
napi: 5
llhttp: 2.0.1
openssl: 1.1.1d
cldr: 36.0
icu: 65.1
tz: 2019c
unicode: 12.1
```

**Your site `_config.yml`** (Optional)

```
```

**Your theme `_config.yml`** (Optional)

```
```

**Hexo and Plugin version(`npm ls --depth 0`)**

```
+-- hexo@4.2.0
+-- hexo-deployer-git@2.1.0
+-- hexo-generator-alias@0.1.3
+-- hexo-generator-archive@1.0.0
+-- hexo-generator-category@1.0.0
+-- hexo-generator-tag@1.0.0
+-- hexo-renderer-ejs@1.0.0
+-- hexo-renderer-marked@2.0.0
+-- hexo-renderer-stylus@1.1.0
`-- hexo-server@1.0.0
```

**Your package.json `package.json`**

```
{
  "name": "hexo-site",
  "version": "0.0.0",
  "private": true,
  "scripts": {
    "build": "hexo generate",
    "clean": "hexo clean",
    "deploy": "hexo deploy",
[...]

PR #4255, opened 2020-04-21
chore/ci: drop Node.js 8 and add Node.js 14

SukkaW:
## What does it do?

Drop Node 8 (#3818) . It has reached EOL since 2019-12-31 and Node 10 is now in Long Term LTS.

Node.js 10.18.1+ will be used after this PR.

## How to test

```sh
git clone node-10.0 https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.


package.json
```diff
@@ -80,7 +80,7 @@
     "sinon": "^8.0.1"
   },
   "engines": {
-    "node": ">=8.10.0"
+    "node": ">=10.18.1"
```
JLHwung, line 83: I am wondering why `10.18.1` is chosen as the minimum support version? Are we depending on the API provided since `10.18.1`?

AFAIK all `node >= 10.9 < 11` shares the same V8 major and `node >= 10.12` has recursive `fs.mkDir` support which we may benefit from.

curbengh: [`10.18.0`](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V10.md#2019-12-17-version-10180-dubnium-lts-mylesborins) includes a critical security fix.

[`10.18.1`](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V10.md#2020-01-09-version-10181-dubnium-lts-bethgriggs) has timezone fix, maybe related to https://github.com/hexojs/hexo/issues/4239?

JLHwung: I don't see a strong reason limit to `>=10.18.1` because

1) Security fix will always be patched as long as node.js 10 is in active LTS but apparently we will not bump `engines` to latest patches

2) On the timezone fix, did you refer to the `tzdata` updates? It seems to me that this is a routine thing but I may be wrong.

The `engines` requirements should be specified as a minimal versions to run `hexo` successfully. So typically it is bound to specific V8 versions / Node.js API, which means `hexo` can not even run on Node.js versions lower than `10.x.y`.

That said, it does not mean that we recommend users to _stay_ at the lowest version. The best practice is to use LTS and to update to the latest minors.

SukkaW: @JLHwung @curbengh 

Well, what about Node.js 10.13.0 then? It is the first LTS version of Node.js 10 and could be used as the minimum required version then.

SukkaW: FYI, https://github.com/hexojs/hexo-fs/pull/55 required Node.js version to be higher than 10.12.0.

PR #4258, opened 2020-04-22
fix(#4236): don't create "<page>/index" directories when post_asset_folder is true

### LINKED ISSUE #4236: Images in Page "index" directories don't work like normal post_asset_folder's
## Expected behavior
post_asset_folder: true creates asset folders for pages that work

## Actual behavior
post_asset_folder: true creates a folder named index in the root directory when you run the $hexo new page <pagename> command (/<pagename>/index) but this folder does not work like the other asset folders.

## How to reproduce?
1. Create a page using the command
2. Put assets in the folder that's created along side of index.md
3. Images are not referenced via markdown ("") or plugin ("{%asset_img%}") as per usual asset_folder behavior

## Environment & Settings

**Node.js & npm version**
Windows 10 64
Node v12.10
NPM 6.10.3

jiangtj:
## What does it do?

fix #4236

## How to test

```sh
git clone -b fix-post-asset-folder https://github.com/jiangtj/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [ ] Passed the CI test.


SukkaW:
Why shouldn't page has asset_folder then?

Considering the case:

```
├─source
│  ├─example-page.md
│  ├─example-page
│  │  ├─some-pic.png
```

jiangtj:
In fact, the `hexo new` command only has these variables and cannot set `permalink`

https://github.com/hexojs/hexo/blob/b60f815b975178ee1c907a92e5a1403830e56f37/lib/plugins/console/index.js#L49-L60

So when we execute the `hexo new page` (`layout` is `page`), we will get the following result

```
├─source
│  ├─example-page
│  │  ├─index.md
│  │  ├─index(The current master will create this directory, but for `index.md`, this `example-page` is the resource directory, not `example-page/index`)
```

I tried using `hexo new page example-page -p example-page` and got the same path as in your example, I may need to modify the source code to filter out the path suffix `/index.md`, not all pages

jiangtj:
I encountered a serious problem, the test case is perfect to run locally, but the error is reported on ci. . .

PR #4274, opened 2020-04-26
perf(open_graph): replace striptags with regexp

SukkaW:
## What does it do?

hexo-util's `stripHTML()` is using `striptags`, which is designed to be able to handle any invalid html made by XSS attacker.

However, the stripHTML here only needs to handle `page.content`, which is definitely valid HTML as it is generated by renderer (marked, Nunjucks, etc.), a simple regexp will do the job well then.

## How to test

```sh
cd hexo
npm install
npm test
```

## Screenshots

The performance of generation improved by 7% according to the benchmark.

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/helper/open_graph.js
+++ b/lib/plugins/helper/open_graph.js
@@ -2,7 +2,8 @@
 
 const { parse, resolve } = require('url');
 const { isMoment, isDate } = require('moment');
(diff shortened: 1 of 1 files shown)


SukkaW:
Closed due to security reason.

PR #4278, opened 2020-04-27
feat: bring up config.updated_option

SukkaW:
## What does it do?

Continuation of #3235, specifically #3235 (comment). https://github.com/hexojs/hexo/pull/3887#issuecomment-567723586

Quote from #3887

> User may prefer to have full control on value of post.updated and not use file modification time. Currently for my blog, I have to use a new variable lastUpdated as a workaround; since that variable is not recognized by many plugins, I have to fork many plugins.
>
> With this PR, config.optional_updated can be enabled (disabled by default), to leave post.updated to be empty if updated: is not set in the front-matter.

This PR introduced a new option `updated_option` which accept:

- `mtime`: This is the default behavior of hexo, use file's `mtime` as `data.updated` when not set in front-matter.
- `date`: Use `data.date` as `data.updated` when not set in front-matter.
- `empty`: Completely remove `data.updated` when not set in front-matter. This shouldn't be used by common users as it *will* break many themes (like NexT) and plugins (like hexo-generator-feed)

The `use_date_for_updated` introduced in hexo v4 is still retained, and still has higher priority than `updated_option` (That's mean when `use_date_for_updated` is set to true, `updated_option` will be ignored).

### Breaking changes

See #3887 for more details.

## How to test

```sh
cd hexo
npm install
npm test
```

## Pull request tasks

- [ ] Passed the CI test.

### DIFF
--- a/lib/hexo/default_config.js
+++ b/lib/hexo/default_config.js
@@ -55,8 +55,10 @@ module.exports = {
   // Date / Time format
   date_format: 'YYYY-MM-DD',
   time_format: 'HH:mm:ss',
-  // Use post date for updated date instead of file modification date when front-matter provides no updated date
-  use_date_for_updated: false,
+  updated_option: 'mtime',
+  // * mtime: file modification date (default)
+  // * date: use_date_for_updated
+  // * empty: no more update
   // Pagination
   per_page: 10,
   pagination_dir: 'page',
--- a/lib/models/page.js
+++ b/lib/models/page.js
@@ -17,7 +17,6 @@ module.exports = ctx => {
     },
     updated: {
       type: Moment,
-      default: moment,
       language: ctx.config.languages,
       timezone: ctx.config.timezone
     },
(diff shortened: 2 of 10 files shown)


SukkaW:
Without `default: moment` defined in `Post` model, the `page.updated` will no longer exist by default, causing open_graph test case (which is rely on `page.updated`) to fail.

curbengh:
> Without `default: moment` defined in `Post` model, the `page.updated` will no longer exist by default, causing open_graph test case (which is rely on `page.updated`) to fail.

Comment out [that line](https://github.com/hexojs/hexo/blob/70ad7e6b21ca0e4b27ee3f9c2c799f2259fd2ec9/test/scripts/helpers/open_graph.js#L52) and mention this PR.

SukkaW:
@curbengh Done.

The issue is now ready for reivew.

curbengh:
I re-label this as breaking change due to https://github.com/hexojs/hexo/pull/4278/files#diff-a17a4a3f711955a771f0f61a61525fd9L30, `post.updated` no longer has a default value and it could be a breaking change for dev that uses Hexo API.

PR #4299, opened 2020-05-10
Reduce array#reduce

segayuu:
## How to test

```sh
cd hexo
npm install
npm test
```

## Pull request tasks

- [ ] Passed the CI test.

### DIFF
--- a/lib/box/index.js
+++ b/lib/box/index.js
@@ -78,30 +78,13 @@ class Box extends EventEmitter {
       process: fn
     });
   }
-  _readDir(base, fn, prefix = '') {
-    const { ignore } = this;
 
-    if (base && ignore) {
-      if (ignore.length && isMatch(base, ignore)) {
-        return Promise.resolve([]);
-      }
-    }
-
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -78,7 +78,7 @@
     "sinon": "^9.0.2"
   },
   "engines": {
-    "node": ">=10.13.0"
```
SukkaW, line 81: Noticed that the minimum required Node.js version is bumped to 11.14.0.

stevenjoezhang: Node.js 11 is end of life. If possible, the minimum required Node.js version should be bumped to 12.x

segayuu: https://github.com/hexojs/warehouse/issues/83#issuecomment-629762394
This can be thought of as Hexo is not ready for a PR that depends on node @ 12 and above.
I'm wondering if this PR merge should be pending so much, as there will be a lot of commits while waiting for a major Hexo update.

SukkaW: Is it possible to make this PR compatible with Node.js 12?

segayuu: Yes. We need to add one more nest of then().

PR #4333, opened 2020-05-26
Reduce stream

segayuu:
## How to test

```sh
cd hexo
npm install
npm test
```

## Pull request tasks

- [ ] Passed the CI test.

### DIFF
--- a/lib/box/index.js
+++ b/lib/box/index.js
@@ -216,20 +216,17 @@ function escapeBackslash(path) {
 }
 
 function getHash(path) {
-  return new Promise((resolve, reject) => {
-    const src = createReadStream(path);
-    const hasher = createSha1Hash();
-    let hash_value;
+  const src = createReadStream(path);
+  const hasher = createSha1Hash();
 
-    src.pipe(hasher)
-      .on('data', chunk => {
-        hash_value = chunk.toString('hex');
-      })
-      .on('end', () => {
-        resolve(hash_value);
-      })
-      .on('error', reject);
+  const finishedPromise = new Promise((resolve, reject) => {
+    src.once('error', reject);
+    src.once('end', resolve);
   });
+
+  src.on('data', chunk => { hasher.update(chunk); });
+
+  return finishedPromise.then(() => hasher.digest('hex'));
 }
 
 function toRegExp(ctx, arg) {
--- a/lib/plugins/console/generate.js
+++ b/lib/plugins/console/generate.js
@@ -7,7 +7,7 @@ const prettyHrtime = require('pretty-hrtime');
 const { cyan, magenta } = require('chalk');
 const tildify = require('tildify');
 const { PassThrough } = require('stream');
-const { CacheStream, createSha1Hash } = require('hexo-util');
+const { createSha1Hash } = require('hexo-util');
 
 function generateConsole(args = {}) {
   const force = args.f || args.force;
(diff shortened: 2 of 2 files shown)


SukkaW:
Whoa, nearly all test failed!

lib/plugins/helper/toc.js
```diff
@@ -3,12 +3,13 @@
 const { tocObj, escapeHTML, encodeURL } = require('hexo-util');
 
 function tocHelper(str, options = {}) {
-  options = Object.assign({
+  options = {
```
SukkaW, line 6: Only when first parameter of `Object.assign` is an empty Object will syntax be faster than `Object.assign`. Otherwise I can't see why should we use spread syntax to shallow merge Object.

lib/extend/console.js
```diff
@@ -5,17 +5,21 @@ const abbrev = require('abbrev');
...
+    return this.store.get(this.alias[name]);
   }
 
   list() {
-    return this.store;
+    const result = {};
+    for (const [key, value] of this.store) {
+      result[key] = value;
+    }
+    return result;
```
SukkaW, line 22:
```suggestion
    // FIXME: Could use Object.fromEntries after Node.js 10 is dropped.
    const result = {};
    for (const [key, value] of this.store) {
      result[key] = value;
    }
    return result;
```

Could add a `FIXME` message.

lib/extend/filter.js
```diff
@@ -9,12 +9,18 @@ const typeAlias = {
...
-    if (!type) return this.store;
-    return this.store[type] || [];
+    if (!type) {
+      const result = {};
+      for (const [key, value] of this.store) {
+        result[key] = value;
+      }
+      return result;
+    }
+    return this.store.get(type) || [];
```
SukkaW, line 23: Same here, a `FIXME` message for future development.

lib/extend/generator.js
```diff
@@ -5,15 +5,19 @@ const Promise = require('bluebird');
...
+    this.store = new Map();
   }
 
   list() {
-    return this.store;
+    const result = {};
+    for (const [key, value] of this.store) {
+      result[key] = value;
+    }
+    return result;
```
SukkaW, line 16: Same here, a FIXME message for future development.

lib/extend/helper.js
```diff
@@ -2,22 +2,26 @@
...
+    this.store = new Map();
   }
 
   list() {
-    return this.store;
+    const result = {};
+    for (const [key, value] of this.store) {
+      result[key] = value;
+    }
+    return result;
```
SukkaW, line 13: Same here, a FIXME message for future development.

lib/extend/migrator.js
```diff
@@ -4,15 +4,19 @@ const Promise = require('bluebird');
...
+    this.store = new Map();
   }
 
   list() {
-    return this.store;
+    const result = {};
+    for (const [key, value] of this.store) {
+      result[key] = value;
+    }
+    return result;
```
SukkaW, line 15: Same here, a FIXME message for future development.

lib/models/types/moment.js
```diff
@@ -35,7 +31,7 @@ class SchemaTypeMoment extends SchemaType {
   }
 
   match(value, query, data) {
-    return value ? value.valueOf() === query.valueOf() : false;
+    return !!value && value.valueOf() === query.valueOf();
```
SukkaW, line 34:
```suggestion
    return Boolean(value) && value.valueOf() === query.valueOf();
```

Personally I prefer using `Boolean()` approach. But since it is not included in `eslint-config-hexo`, use `!!` will be fine as well.

lib/hexo/locals.js
```diff
@@ -48,10 +48,8 @@ class Locals {
 
   toObject() {
     const result = {};
-    const keys = Object.keys(this.getters);
 
-    for (let i = 0, len = keys.length; i < len; i++) {
-      const key = keys[i];
+    for (const key of this.getters.keys()) {
```
SukkaW, line 52: Should we use `for..of` to replace traditional `for` loop? Last time I checked the `for..of` is nearly 20% slower than traditional `for` loop.

lib/plugins/processor/post.js
```diff
@@ -60,9 +59,8 @@ module.exports = ctx => {
         data.published = false;
       }
 
-      for (let i = 0, len = keys.length; i < len; i++) {
-        const key = keys[i];
-        if (!preservedKeys[key]) data[key] = info[key];
+      for (const key of Object.keys(info)) {
```
SukkaW, line 62: Same here. `for..of` is considered to be not so fast as traditional `for` loop.

SukkaW:
Also, the generating performance regression should be aware as well (Total generation time has increased by 23% during the benchmark).

lib/hexo/locals.js
segayuu: The return value of `Map#keys()` is Iterable, not Array. This is different than `Object.keys()`.
It's interesting how fast I get if I wrap it in `Array.from()` and do it in a for loop :)

SukkaW:
The Map is only faster for frequent adding & removals.

For extend APIs (which only have a few registers), it seems better to retain using Object.

lib/box/index.js
```diff
@@ -216,20 +216,17 @@ function escapeBackslash(path) {
...
+  const hasher = createSha1Hash();
+
+  const finishedPromise = new Promise((resolve, reject) => {
+    src.once('error', reject);
+    src.once('end', resolve);
   });
+
+  src.on('data', chunk => { hasher.update(chunk); });
+
+  return finishedPromise.then(() => hasher.digest('hex'));
```
SukkaW, line 229: Would you mind revert the changes here? It seems that the changes caused the performance regression. Let's see if any improvements could be made by reverting changes here

SukkaW:
Also, would you mind splitting into more smaller PR? Too many commits cover too many fields make it very hard to review and locate the performance issue.

segayuu:
Thank you for the PR split. This PR will be reduced according to the title.

PR #4339, opened 2020-05-30
Fix not to pass callback to hexo-fs

segayuu:
## What does it do?
The PR of #4336 is a provisional measure.
hexo-fs@3.1.0 needs to be modified to not pass the callback directly to hexo-fs.

## How to test

```sh
cd hexo
npm install
npm test
```

## Pull request tasks

- [ ] Passed the CI test.

### DIFF
--- a/lib/box/file.js
+++ b/lib/box/file.js
@@ -11,6 +11,10 @@ class File {
   }
 
   read(options, callback) {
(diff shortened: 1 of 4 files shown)


SukkaW:
Noticed that Windows test failed on delete operation (`ENOTEMPTY`). Could dig into it later.

Issue #4228, opened 2020-04-07
Does not display the title of the article


scillidan:
nihao
When I use `hexo 3.8.0` and `icarus 2.3.0` , I can use `title: #` to make the article on the **homepage(important!)** not display the title, like this
![](https://raw.githubusercontent.com/scillidan/scillidan.driver.github/master/icarus_issues_569.png)
But when I use `hexo 4.2.0`, when I do this, the title will be displayed as a `.md` file name.
I don't know if this is a bug or a request. I also submitted an issue in [this place](https://github.com/ppoffice/hexo-theme-icarus/issues/663)
I hope to have this option to can hide the title.

stevenjoezhang:
This issue is caused by https://github.com/hexojs/hexo/pull/3672

See also

https://github.com/hexojs/hexo-theme-unit-test#post
https://github.com/hexojs/hexo-theme-unit-test/blob/master/source/_posts/no-title.md

In addition, `title: #` is not standard usage. Just leave it blank.

scillidan:
thanks, I go watching it.

PR #3672, opened 2019-08-16
if post.title is empty, set it to post.slug

rupertqin:
## How to test

```sh
git clone -b BRANCH https://github.com/USER/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Add test cases for the changes.
- [x] Passed the CI test.

PR #4344, opened 2020-06-02
Posts without title should be accessible

stevenjoezhang:
## What does it do?

Issue resolved: https://github.com/hexojs/hexo/issues/4228
See also https://github.com/hexojs/hexo/pull/3672

## How to test

```sh
git clone -b revert-3672 https://github.com/hexojs/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [ ] Passed the CI test.

### DIFF
--- a/lib/plugins/processor/post.js
+++ b/lib/plugins/processor/post.js
@@ -120,10 +120,6 @@ module.exports = ctx => {
         data.title = data.link.replace(/^https?:\/\/|\/$/g, '');
       }
 
-      if (!data.title) {
-        data.title = data.slug;
-      }
-
       if (data.permalink) {
         data.slug = data.permalink;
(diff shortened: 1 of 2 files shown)


SukkaW:
We should retain empty title behavior, as `slug` might not be needed for every one.

LuchaoQi:
> We should retain empty title behavior, as `slug` might not be needed for every one.

agreed

stevenjoezhang:
@LuchaoQi Are you using hexo-theme-icarus? Please open an issue here: https://github.com/ppoffice/hexo-theme-icarus/issues

"Posts without title should be accessible." is required by [hexo-theme-unit-test](https://github.com/hexojs/hexo-theme-unit-test#post)

PR #4359, opened 2020-06-16
fix(#3464): override permalink use the front-matter

### LINKED ISSUE #1158: Overriding permalink in front-matter
I'm trying to override the permalink for individual posts in the "front-matter" but I'm not having much luck. Is the feature actually supported, and if so what is the proper configuration to have it working? Any pointers?

### LINKED ISSUE #1813: Permalink issues
This is an umbrella issue pertaining permalink in Hexo.
We should document how permalink (or link in general) is resolved from `config.permalink`, `post.permalink` file name, slug, title.
These cause quite a bit of confusion for the users and escaping of special characters.
#1812
#1608
#1110
#881 / #1258

https://github.com/hexojs/hexo-generator-sitemap/issues/1 Use URL safe characters (slug should be serving this purpose)
https://github.com/hexojs/hexo-generator-sitemap/issues/13 Use permalink as canonical URL (in HTML header and sitemap)
#1822 `.` in post file name

### LINKED ISSUE #3464: The front-matter's permalink can not override config.permalink
from #3442 

## Environment Info

Your site `_config.yml` (Optional):

```yaml
new_post_name: :year-:month-:day-:title.md
```

and frontmatter is 

```md
---
title: "This is a test"
date: 2019-01-31 20:44:31
permalink: first-blog
---
```

## Expected behavior

URL is `http://localhost:3000/first-blog`

## Actual behaviot

URL is `http://localhost:3000/2019/01/first-blog`

SukkaW:
## What does it do?

The PR closes #1158, closes #1813, closes #3464

https://github.com/hexojs/hexo/issues/3464#issuecomment-644238739

https://github.com/hexojs/hexo/blob/edef5c2a9c1217aaab02f64397e7e94d2f98e4fa/lib/plugins/processor/post.js#L123-L126

Originally, during the processing of the post, the `permalink` defined in the front-matter which is `data.permalink`, will override `data.slug` and `data.permalink` will be deleted.

And, in `post_permalink` filter, `data.slug` will be used to get `name` and `title` property:

https://github.com/hexojs/hexo/blob/edef5c2a9c1217aaab02f64397e7e94d2f98e4fa/lib/plugins/filter/post_permalink.js#L15-L16

So if I got a post: `source/_posts/i-love-hexo/hello.md`, I will get:

- title: `i-love-hexo/hello`
- name: `hello`

If I then defines `permalink` as `permalink-test/hi.md`, I will get:

- title: `permalink-test/hi`
- name: `hi`.

I don't know why it is designed like this. Maybe using the `permalink` to override the relative path under `source/_posts`. But it doesn't make any sense - why I have to override the relative path? I could just create required directories under `source/_posts` and it will be fine.

Also, the behavior of `Post` is different from `Page`, which `permalink` in the front-matter is truly able to override the `hexo.config.permalink`. And that behavior is what user really want.

So in this PR, I made a few changes toward post processor and `post_permalink` filter.

**As the behavior of the `permalink` in the front-matter has changed, thus the PR is still considered as a Breaking Change**

## How to test

```sh
git clone -b fix-issue-3464 https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/filter/post_permalink.js
+++ b/lib/plugins/filter/post_permalink.js
@@ -6,7 +6,13 @@ let permalink;
 
 function postPermalinkFilter(data) {
   const { config } = this;
-  const { id, _id, slug, title, date } = data;
+  const { id, _id, slug, title, date, __permalink } = data;
+
+  if (__permalink) {
+    if (!__permalink.startsWith('/')) return `/${__permalink}`;
+    return __permalink;
+  }
+
   const hash = slug && date
     ? createSha1Hash().update(slug + date.unix().toString()).digest('hex').slice(0, 12)
     : null;
@@ -57,7 +63,6 @@ function postPermalinkFilter(data) {
       meta[key] = config.permalink_defaults[key];
     }
   }
-
 
   return permalink.stringify(meta);
 }
--- a/lib/plugins/processor/post.js
+++ b/lib/plugins/processor/post.js
@@ -121,7 +121,7 @@ module.exports = ctx => {
       }
 
       if (data.permalink) {
-        data.slug = data.permalink;
+        data.__permalink = data.permalink;
         delete data.permalink;
       }
 
--- a/test/scripts/filters/post_permalink.js
+++ b/test/scripts/filters/post_permalink.js
@@ -148,4 +148,27 @@ describe('post_permalink', () => {
 
     await Promise.all(posts.map(post => Post.removeById(post._id)));
   });
+
+  it('permalink - should override everything', async () => {
+    hexo.config.permalink = ':year/:month/:day/:title/';
+
+    const posts = await Post.insert([{
+      source: 'my-new-post.md',
+      slug: 'hexo/permalink-test',
+      __permalink: 'hexo/permalink-test',
+      title: 'Permalink Test',
+      date: moment('2014-01-02')
+    }, {
+      source: 'another-new-post.md',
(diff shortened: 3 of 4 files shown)


lib/plugins/processor/post.js
```diff
@@ -121,7 +121,7 @@ module.exports = ctx => {
       }
 
       if (data.permalink) {
-        data.slug = data.permalink;
+        data.__permalink = data.permalink;
```
SukkaW, line 124: BREAKING CHANGES.

`data.slug` won't be override by `data.permalink`. That means `data.slug` will still come from the original path relative to `source/_posts`.

Also,  a new `__permalink` is created to avoid overriding `Post.virtual('permalink').get.

lib/plugins/filter/post_permalink.js
```diff
@@ -6,7 +6,13 @@ let permalink;
 
 function postPermalinkFilter(data) {
   const { config } = this;
-  const { id, _id, slug, title, date } = data;
+  const { id, _id, slug, title, date, __permalink } = data;
+
+  if (__permalink) {
+    if (!__permalink.startsWith('/')) return `/${__permalink}`;
+    return __permalink;
+  }
```
SukkaW, line 14: When a `__permalink` is detected (which is added by post processor from `permalink` inside the front-matter), it will be return directly.

PR #4360, opened 2020-06-16
refactor(hexo): merge theme_config before generation

### LINKED ISSUE #3890: The proposal of installing theme through npm

SukkaW:
## What does it do?

#4120, https://github.com/hexojs/hexo/pull/4120#issuecomment-619574347, https://github.com/hexojs/hexo/issues/3890#issuecomment-644603994

This PR closes #3890.

## How to test

```sh
git clone -b theme_config_ctx https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/hexo/index.js
+++ b/lib/hexo/index.js
@@ -311,6 +311,12 @@ class Hexo extends EventEmitter {
         this.source.process(),
         this.theme.process()
       ]);
+    }).then(() => {
+      // Merge hexo.config.theme_config into hexo.theme.config before post rendering & generating
+      // config.theme_config has "_config.[theme].yml" merged in load_theme_config.js
+      if (this.config.theme_config) {
+        this.theme.config = deepMerge(this.theme.config, this.config.theme_config);
+      }
     }).then(() => this._generate({cache: false})).asCallback(callback);
   }
 
@@ -358,11 +364,6 @@ class Hexo extends EventEmitter {
     const { config, env, theme, theme_dir } = this;
     const ctx = { config: { url: this.config.url } };
     const localsObj = this.locals.toObject();
-
-    // config.theme_config has "_config.[theme].yml" merged in load_theme_config.js
-    if (config.theme_config) {
-      theme.config = deepMerge(theme.config, config.theme_config);
-    }
 
     class Locals {
       constructor(path, locals) {
--- a/lib/hexo/index.js
+++ b/lib/hexo/index.js
@@ -35,6 +35,14 @@ const routeCache = new WeakMap();
 
 const castArray = obj => { return Array.isArray(obj) ? obj : [obj]; };
 
+const mergeCtxThemeConfig = ctx => {
+  // Merge hexo.config.theme_config into hexo.theme.config before post rendering & generating
+  // config.theme_config has "_config.[theme].yml" merged in load_theme_config.js
+  if (ctx.config.theme_config) {
+    ctx.theme.config = deepMerge(ctx.theme.config, ctx.config.theme_config);
+  }
+};
+
 const createLoadThemeRoute = function(generatorResult, locals, ctx) {
   const { log, theme } = ctx;
   const { path, cache: useCache } = locals;
@@ -312,11 +320,7 @@ class Hexo extends EventEmitter {
         this.theme.process()
       ]);
     }).then(() => {
-      // Merge hexo.config.theme_config into hexo.theme.config before post rendering & generating
-      // config.theme_config has "_config.[theme].yml" merged in load_theme_config.js
-      if (this.config.theme_config) {
-        this.theme.config = deepMerge(this.theme.config, this.config.theme_config);
-      }
+      mergeCtxThemeConfig(this);
     }).then(() => this._generate({cache: false})).asCallback(callback);
   }
 
@@ -341,8 +345,13 @@ class Hexo extends EventEmitter {
         this.theme.watch()
       ]);
     }).then(() => {
+      mergeCtxThemeConfig(this);
+    }).then(() => {
       this.source.on('processAfter', this._watchBox);
-      this.theme.on('processAfter', this._watchBox);
+      this.theme.on('processAfter', () => {
+        this._watchBox();
+        mergeCtxThemeConfig(this);
+      });
 
       return this._generate({cache: useCache});
     }).asCallback(callback);
--- a/test/scripts/hexo/hexo.js
+++ b/test/scripts/hexo/hexo.js
@@ -155,7 +155,7 @@ describe('Hexo', () => {
 
   // Issue #3964
   it('load() - merge theme config - deep clone', async () => {
-    const hexo = new Hexo(__dirname);
+    const hexo = new Hexo(__dirname, { silent: true });
     hexo.theme.config = { a: { b: 1, c: 2 } };
     hexo.config.theme_config = { a: { b: 3 } };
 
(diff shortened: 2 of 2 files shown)


tomap:
Hi,

I'm not sure why you are moving the code? Is it done sooner with your code?
Or later?

Why?

SukkaW:
@tomap Technically, sooner. `this._genterateLocals` build `Locals` in `this._generate`, while `this._generate` is called after `this.load`.

More details see https://github.com/hexojs/hexo/pull/4120#issuecomment-619574347 . @stevenjoezhang  want the plugin being able to read `config.theme_config` from `theme.config`.

stevenjoezhang:
Yes, this is very important for theme developers. Otherwise additional code is needed to ensure that the configuration in `theme_config` is loaded correctly.
See also
https://github.com/theme-next/hexo-theme-next/commit/1b0b53a355184db65b2f19349c7b1372f1825174
https://github.com/theme-next/hexo-theme-next/blob/a7a948a08f48dbe48879e277703d75706aec18e5/scripts/events/lib/config.js#L34

stevenjoezhang:
https://github.com/hexojs/hexo/blob/d1178192dbfb56469d2aa604c2ec67c9c957bfdc/lib/hexo/index.js#L304-L340

I confirm that this patch works fine with `hexo g`. However, in `hexo s` the `_generate` method is usually called by `watch` (line 325, 338), instead of `load` (line 312)
Is it possible to put the relevant code directly into the `_generate` method?

SukkaW:
> Is it possible to put the relevant code directly into the `_generate` method?

Then any filter related with post rendering (like `before_post_render`) will be affected.

curbengh:
> Merge theme config during watch() is now supported. Reloading the theme config is supported as well (see the test case added)

My test config:

_config.yml
``` yml
theme_config:
  foo:
    bar: 'bax'
```

theme's _config.yml

``` yml
foo:
  baz: 'bos'
```

article.ejs (to display theme config)
```
<%- JSON.stringify(theme.foo, null, 2) %>
```

---

In `hexo s`, after I change the value of `theme.foo.baz` in theme's _config.yml and reload the page, it crash with undefined property. No effect when changing `theme.foo.bar` in _config.yml, reloading the page doesn't show the updated value.

SukkaW:
> Reloading the page doesn't show the updated value.

@curbengh It is expected behavior. Because site's `_config.yml` won't reload. Only theme's `_config.yml` will.

SukkaW:
@curbengh Could you please re-verify reloading theme's config?

```bash
$ npm i sukkaw/hexo#theme_config_ctx
```

Then

```yaml
# _config.yml
theme_config:
  foo:
    bar: 'bax'
```

```yaml
# theme's _config.yml
foo:
  baz: 'bos'
```

Add following line to the theme template:

```ejs
<% console.log(theme.foo) %>
```

`hexo s` should output:

```
{ baz: 'bos', bar: 'bax' }
```

Update theme's config:

```yaml
# theme's _config.yml
foo:
  baz: 'box'
```

Then reload the page should output:

```
{ baz: 'box', bar: 'bax' }
```

curbengh:
This time I tried showing everything inside `theme` object,

```
<% console.log(JSON.stringify(theme, null, 2)) %>
```

After I changed the `theme.foo.bar` in the theme's config and reload the page, the `theme` object is left with `theme.foo.bar` only.

SukkaW:
@curbengh So it appears that it is not the issue if the PR though. We could dig into the watcher later.

Have you run unit test? Is it passed?

curbengh:
I can confirm it's my machine's issue, not even downgrading to v3.9 helps. I recently migrate Ubuntu - Solus and I don't recall having this watch issue in Ubuntu. Anyway, I'll debug on my own, don't want to distract this PR.

Unit test passed on Node 12 & 14 in my machine. I still prefer to have a theme dev @stevenjoezhang to verify this PR.

SukkaW:
@curbengh @stevenjoezhang 

I have added a test case, using `before_post_render`, `after_post_render`, `before_generate` filters to check if `theme.config` has been merged.

jerryc127:
主題配置文件已經默認了一些配置
例如
item:
  - qq
  - wechat

我的_config.xxxxx.yml
更改配置
item:
  - twitter

結果 是生成 qq wechat twitter 
而不是我修改的 twitter

按道理 如果_config.xxxxx.yml配置已有 應該按照_config.xxxxx.yml的内容為准，因爲它優先性高.
這是本來設計就是這樣麽

curbengh:
Priority is not relevant if a key doesn't have duplicate subkeys, they will be merged, not replaced.

``` yml
# themes/xxxx/_config.yml
item:
  lorem: 'a'
  ipsum: 'b'
```

``` yml
# _config.xxxx.yml
item:
  dolor: 'c'
```

``` js
console.log(theme.config)
item: {
  lorem: 'a',
  ipsum: 'b',
  dolor: 'c'
}
```

---

A subkey will only be replaced if there is duplicate:

``` yml
# themes/xxxx/_config.yml
item:
  lorem: 'a'
  ipsum: 'b'
```

``` yml
# _config.xxxx.yml
item:
  lorem: 'c'
```

``` js
console.log(theme.config)
item: {
  lorem: 'c',
  ipsum: 'b'
}
```

jerryc127:
你这个举例 我明白
问题是我那例子属于subkey么
```
item:
  - qq
  - wechat
```

qq wechat 这个配置都是属于 item这个key的。 qq和wechat都是item 的值，就好比c是lorem的值，b是ipsum的值，
那这样 我的这个例子不是应该替换么。。

难道以 -xxx -xxx -xxx 这种形式赋值 都会被识别为几个subkey?

curbengh:
In your example, `item` has array value. array value is merged.

``` js
const { deepMerge } = require('hexo-util')

const config = {
  theme_config: {
    item: ['a', 'b']
  }
}

const theme = {
  config: {
    item: ['c']
  }
}

theme.config = deepMerge(theme.config, config.theme_config);

console.log(theme.config)
// { item: [ 'c', 'a', 'b' ] }
```

---

If you want to replace the original `item`, you have to remove it.

YunYouJun:
Why not [overwrite array](https://www.npmjs.com/package/deepmerge#arraymerge-example-overwrite-target-array) instead of [combine arrays](https://www.npmjs.com/package/deepmerge#arraymerge-example-combine-arrays)?

Is there any special reason? I think for themes, overwriting array is more suitable to providing some example configurations without disturbing the user.

Or just provide a configuration to decide whether to overwrite the array or merge the array?

PR #4352, opened 2020-06-10
fix(post): swig tag inside post

SukkaW:
## What does it do?

The PR fixes #3259, #3346 & https://github.com/hexojs/hexo/issues/3346#issuecomment-595497849

## How to test

```sh
git clone -b test-post-render-swig https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Add test cases for the changes.
- [x] Passed the CI test.

PR #4362, opened 2020-06-18
fix(post): avoid swig in code being double escaped

SukkaW:
## What does it do?

Continuing #4352

https://github.com/hexojs/hexo/pull/4352/files#r441953318

Marked.js will escape code by default. If the swig inside ` ``  ` has been escaped, it will be escaped again by marked.js.

This PR avoid it.

cc @stevenjoezhang

## How to test

```sh
git clone -b avoid-swig-double-escape https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [ ] Passed the CI test.

### DIFF
--- a/lib/hexo/post.js
+++ b/lib/hexo/post.js
@@ -10,6 +10,9 @@ const { slugize, escapeRegExp } = require('hexo-util');
 const { copyDir, exists, listDir, mkdirs, readFile, rmdir, unlink, writeFile } = require('hexo-fs');
 const yfm = require('hexo-front-matter');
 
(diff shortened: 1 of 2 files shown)


SukkaW:
It seems that the PR has huge impact on performance. I don't know if there is a better way to do that.

SukkaW:
@stevenjoezhang 

Refactored & rebased, with performance improvements.

SukkaW:
@curbengh Would you mind reviewing this again? I have rebased to solve conflict.

PR #4370, opened 2020-06-22
feat(tagcloud): new option class & level

stevenjoezhang:
## What does it do?

The color of the tag in the tag-cloud is defined by the inline style, e.g.
```html
<a href="/tags/bcd/" style="font-size: 20px; color: #4682b4">bcd</a>
```
However, for themes that support the dark mode, it's necessary to display different tag colors for different background colors. Class name will be a better choice.
See also https://github.com/next-theme/hexo-theme-next/commit/cbcd3f04abfbc15c82f1cf430eca15591015fadc

Usage:
```njk
{{ tagcloud({ class: 'tag-cloud' }) }}
```
```styl
for $tag-cloud in (0 .. 10) {
  $tag-cloud-color = mix($tag-cloud-end, $tag-cloud-start, $tag-cloud * 10);
  .tag-cloud-{$tag-cloud} {
    color: $tag-cloud-color;
  }
}

@media (prefers-color-scheme: dark) {
  for $tag-cloud in (0 .. 10) {
    $tag-cloud-color = mix($tag-cloud-end-dark, $tag-cloud-start-dark, $tag-cloud * 10);
    .tag-cloud-{$tag-cloud} {
      color: $tag-cloud-color;
    }
  }
}
```

## How to test

```sh
git clone -b tagcloud https://github.com/hexojs/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/helper/tagcloud.js
+++ b/lib/plugins/helper/tagcloud.js
@@ -18,6 +18,7 @@ function tagcloudHelper(tags, options) {
   const unit = options.unit || 'px';
   const color = options.color;
   const className = options.class;
+  const level = options.level || 10;
   const { transform } = options;
   const separator = options.separator || ' ';
   const result = [];
@@ -58,7 +59,7 @@ function tagcloudHelper(tags, options) {
     const ratio = length ? sizes.indexOf(tag.length) / length : 0;
     const size = min + ((max - min) * ratio);
     let style = `font-size: ${parseFloat(size.toFixed(2))}${unit};`;
-    const attr = className ? ` class="${className}-${Math.round(ratio * 10)}"` : '';
+    const attr = className ? ` class="${className}-${Math.round(ratio * level)}"` : '';
 
     if (color) {
       const midColor = startColor.mix(endColor, ratio);
--- a/lib/plugins/helper/tagcloud.js
+++ b/lib/plugins/helper/tagcloud.js
@@ -66,7 +66,7 @@ function tagcloudHelper(tags, options) {
     }
 
     result.push(
-      `<a href="${this.url_for(tag.path)}" style="${style}"${name}>${transform ? transform(tag.name) : tag.name}</a>`
+      `<a href="${this.url_for(tag.path)}" style="${style}"${attr}>${transform ? transform(tag.name) : tag.name}</a>`
     );
   });
 
--- a/lib/plugins/helper/tagcloud.js
+++ b/lib/plugins/helper/tagcloud.js
@@ -58,7 +58,7 @@ function tagcloudHelper(tags, options) {
     const ratio = length ? sizes.indexOf(tag.length) / length : 0;
     const size = min + ((max - min) * ratio);
     let style = `font-size: ${parseFloat(size.toFixed(2))}${unit};`;
-    const name = className ? ` class="${className}-${Math.round(ratio * 10)}"` : '';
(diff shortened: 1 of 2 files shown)


lib/plugins/helper/tagcloud.js
```diff
@@ -57,14 +58,15 @@ function tagcloudHelper(tags, options) {
     const ratio = length ? sizes.indexOf(tag.length) / length : 0;
     const size = min + ((max - min) * ratio);
     let style = `font-size: ${parseFloat(size.toFixed(2))}${unit};`;
+    const name = className ? ` class="${className}-${Math.round(ratio * 10)}"` : '';
```
jiangtj, line 61: I perfer `ceil`

if use `round`
| Level | 0 | 1 | ... | 9 | 10 |
| --- | --- | --- | --- | --- | --- |
| Probability | 5% | 10% | --- | 10% | 5% |

if use `ceil`
| Level | 1 | 2 | ... | 9 | 10 |
| --- | --- | --- | --- | --- | --- |
| Probability | 10% | 10% | --- | 10% | 10% |

stevenjoezhang: But how about level 0? `Math.ceil(0) == 0`, but there is no level 0 in your `Math.ceil` table.
Your theory seems to describe the case where tag.length approaches infinity
| Level | 0 | 1 | 2 | ... | 9 | 10 |
| --- | --- | --- | --- | --- | --- | --- |
| Probability | 0% | 10% | 10% | ... | 10% | 10% |

However, the actual situation is that tag.length is finite (and ratio is discrete), so the probability of level 0 must be greater than 0.

stevenjoezhang:
```suggestion
    let level = Math.ceil(ratio * 10);
    if (level === 0) level ++;
    const name = className ? ` class="${className}-${level}"` : '';
```

stevenjoezhang: In fact the asymmetry of level 0 and level 10 comes from this line
```js
const ratio = length ? sizes.indexOf(tag.length) / length : 0;
```

```js
const ratio = length ? (sizes.indexOf(tag.length) + 0.5) / length : 0;
```
Of course, I don't think this change is necessary. We are designing a blog system instead of some sort of scientific computing. Such an error (0.5 / length) is acceptable.

SukkaW: `name` could be replaced with `attr`.

lib/plugins/helper/tagcloud.js
```diff
@@ -57,14 +58,15 @@ function tagcloudHelper(tags, options) {
...
 
     if (color) {
       const midColor = startColor.mix(endColor, ratio);
       style += ` color: ${midColor.toString()}`;
     }
 
     result.push(
-      `<a href="${this.url_for(tag.path)}" style="${style}">${transform ? transform(tag.name) : tag.name}</a>`
+      `<a href="${this.url_for(tag.path)}" style="${style}"${name}>${transform ? transform(tag.name) : tag.name}</a>`
```
SukkaW, line 69:
```suggestion
      `<a href="${this.url_for(tag.path)}" style="${style}"${attr}>${transform ? transform(tag.name) : tag.name}</a>`
```

Same goes here.

PR #4376, opened 2020-06-23
refactor: deprecate config.external_link as boolean

SukkaW:
## What does it do?

`config.external_link` with boolean value is deprecated since Hexo 4.0.0. It is now removed.

## How to test

```sh
git clone -b deprecate-external-link-filter-boolean https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/filter/after_post_render/external_link.js
+++ b/lib/plugins/filter/after_post_render/external_link.js
@@ -10,15 +10,16 @@ function externalLinkFilter(data) {
   const { config } = this;
 
   if (typeof EXTERNAL_LINK_POST_CONFIG === 'undefined') {
-    if (typeof config.external_link === 'undefined' || typeof config.external_link === 'object'
-      // Deprecated: config.external_link boolean option will be removed in future
-      || config.external_link === true) {
-      EXTERNAL_LINK_POST_CONFIG = Object.assign({
-        enable: true,
-        field: 'site',
-        exclude: []
-      }, config.external_link);
+    // Deprecated: config.external_link no longer support boolean
+    if (typeof config.external_link === 'boolean') {
+      throw new TypeError('config.external_link no longer supports Boolean value, changelog: https://github.com/hexojs/hexo/releases/');
     }
+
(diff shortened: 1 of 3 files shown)


lib/plugins/filter/after_post_render/external_link.js
```diff
@@ -10,15 +10,16 @@ function externalLinkFilter(data) {
...
-      // Deprecated: config.external_link boolean option will be removed in future
-      || config.external_link === true) {
-      EXTERNAL_LINK_POST_CONFIG = Object.assign({
-        enable: true,
-        field: 'site',
-        exclude: []
-      }, config.external_link);
+    // Deprecated: config.external_link no longer support boolean
+    if (typeof config.external_link === 'boolean') {
+      throw new TypeError('config.external_link no longer supports Boolean value, changelog: https://github.com/hexojs/hexo/releases/');
```
curbengh, line 15: I think it would better to retain support for now (deprecate but not dropping it yet) and display a warning.
Since it will deprecated in v5, it's clearer to link directly to https://github.com/hexojs/hexo/releases/tag/5.0.0
Usage of `config.use_date_for_updated` should also display warning.

SukkaW: We already deprecate it in Hexo 4.0.0. So it is fine when we remove it in Hexo 5.0.0.

I will open another PR to add a warning for `use_date_for_updated`.

curbengh: > We already deprecate it in Hexo 4.0.0.

That was the initial intention, however we forgot to mention it in the v4 [release note](https://github.com/hexojs/hexo/releases/tag/4.0.0). Can we _soft_-deprecate it this time?

SukkaW: @curbengh IMHO we could bring up a config validation.

SukkaW:
#4381

PR #4381, opened 2020-06-26
feat(config): validate config after load_config

SukkaW:
## What does it do?

Continuation fo #4371 and supersedes #4376. 

## How to test

```sh
git clone -b config-validation https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/hexo/validate_config.js
+++ b/lib/hexo/validate_config.js
@@ -1,0 +1,33 @@
+'use strict';
+
+module.exports = ctx => {
+  const { config, log } = ctx;
+
+  log.info('Validating config');
+
+  // Validation for config.url && config.root
+  if (typeof config.url !== 'string') {
+    throw new TypeError('_config.yml validation failed, "url" should be string!');
+  }
+  if (config.url.trim().length <= 0) {
+    throw new TypeError('_config.yml validation failed, "url" should not be empty!');
+  }
+
+  if (typeof config.root !== 'string') {
+    throw new TypeError('_config.yml validation failed, "root" should be string!');
+  }
+  if (config.root.trim().length <= 0) {
+    throw new TypeError('_config.yml validation failed, "root" should not be empty!');
+  }
+
+  // Soft deprecate use_date_for_updated
+  if (typeof config.use_date_for_updated === 'boolean') {
+    log.warn('_config.yml validation: "use_date_for_updated" is deprecated, please use "updated_option" instead. See https://hexo.io/docs/configuration for more details.');
+  }
+
+  // Soft deprecate external_link Boolean
+  if (typeof config.external_link === 'boolean') {
+    log.warn('_config.yml validation: "external_link" with a Boolean value is deprecated. See https://hexo.io/docs/configuration for more details.');
+  }
+};
+
--- a/lib/hexo/validate_config.js
+++ b/lib/hexo/validate_config.js
@@ -7,14 +7,14 @@ module.exports = ctx => {
 
   // Validation for config.url && config.root
   if (typeof config.url !== 'string') {
-    throw new TypeError('Invalid config detected: "url" should be string!');
+    throw new TypeError(`Invalid config detected: "url" should be string, not ${typeof config.url}!`);
   }
   if (config.url.trim().length <= 0) {
     throw new TypeError('Invalid config detected: "url" should not be empty!');
   }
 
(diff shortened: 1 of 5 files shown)


lib/hexo/validate_config.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+
+module.exports = ctx => {
+  const { config, log } = ctx;
+
+  log.info('Validating config');
+
+  // Validation for config.url && config.root
+  if (typeof config.url !== 'string') {
+    throw new TypeError('_config.yml validation failed, "url" should be string!');
```
curbengh, line 10: I think `Invalid config detected. "url"...` is more suitable since `custom.yml` and `_config.json` are also supported.

SukkaW: LGTM. I will update to that.

curbengh:
Is it possible to run it in `generate` & `server` only? Currently this is also executed in `clean`.

jiangtj:
How about use some schema schema validator? e.g. [ajv](https://github.com/ajv-validator/ajv) [joi](https://github.com/hapijs/joi) etc.

SukkaW:
@jiangtj It can be a part of a roadmap. But so far we just don't need it.

lib/hexo/validate_config.js
```diff
@@ -0,0 +1,33 @@
...
+
+  log.info('Validating config');
+
+  // Validation for config.url && config.root
+  if (typeof config.url !== 'string') {
+    throw new TypeError('Invalid config detected: "url" should be string!');
+  }
+  if (config.url.trim().length <= 0) {
+    throw new TypeError('Invalid config detected: "url" should not be empty!');
+  }
```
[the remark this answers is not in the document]
curbengh: config validation can only be executed between yml/json loading and `String.replace()`.
Perhaps validate inside `load_config.js`?

curbengh, line 14: suggest to include `undefined` checking here as well, need to run before 'string' checking.

test/scripts/hexo/validate_config.js
```diff
@@ -0,0 +1,102 @@
...
+    hexo.log.warn = logSpy;
+    hexo.log.info = spy();
+  });
+
+  it('config.url - undefined', () => {
+    delete hexo.config.url;
+
+    try {
+      validateConfig(hexo);
+    } catch (e) {
```
curbengh, line 24: for more accurate test, suggest to have `should.fail()` right after `validateConfig()`.

lib/hexo/validate_config.js
SukkaW: @curbengh Added in L9:

```js
  if (typeof config.url !== 'string') {
```

`undefined` just won't pass it.

test/scripts/hexo/validate_config.js
SukkaW: Added in https://github.com/hexojs/hexo/pull/4381/commits/c2d34280969fc774e62c3c8894732be195f5912b

lib/hexo/validate_config.js
curbengh: That check is inaccurate for `undefined` since it's an empty value.

SukkaW: @curbengh I have already added an undefined test cases. It seems working properly.

test/scripts/hexo/validate_config.js
```diff
@@ -0,0 +1,108 @@
...
+
+  it('config.url - undefined', () => {
+    delete hexo.config.url;
+
+    try {
+      validateConfig(hexo);
+      should.fail();
+    } catch (e) {
+      e.name.should.eql('TypeError');
+      e.message.should.eql('Invalid config detected: "url" should be string!');
```
curbengh, line 27: `undefined` is practically an _empty_ value. If a user accidentally removes the value of `url:` or even the key itself, it's clearer to say `"url" should not be empty`.

SukkaW: Although for YAML, `undefined` is usually caused by empty value. But Hexo also supports config in `json` format (Although most people just won't using json).

Anyway, I will update the error message then.

SukkaW:
@curbengh Would you mind have a look at the PR again?

PR #4398, opened 2020-07-05
fix: remove unused type check

Himself65:
`args` must be an array


SukkaW:
Welcome @Himself65 to join @hexojs/core team!

PR #4401, opened 2020-07-06
test: replace sinon#assert

SukkaW:
## What does it do?

Replace `sinon#assert` with `Chai.Assertion.LanguageChain`.

## How to test

```sh
git clone -b test-replace-sinonAssert https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.


segayuu:
This PR is effectively a Revert of #4164 that was there before.
The reason I put that PR out is to show a detailed message if the assert fails.
Did you see a downside to the reliance on sinon#assert over the benefit of getting a detailed message?

SukkaW:
@segayuu 

Emmmm, I see. I think I should revert the PR then.

PR #4413, opened 2020-07-11
Replace(benckmark): from travis to github actions

segayuu:
## How to test

```sh
cd hexo
npm install
npm test
```

## Pull request tasks

- [ ] Passed the CI test.


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,54 @@
...
+          repository: hexojs/hexo-theme-landscape
+          path: .tmp-hexo-theme-unit-test/themes/landscape
+      - name: Install Dependencies
+        run: npm install --silent
+        working-directory: .tmp-hexo-theme-unit-test
+      - name: Replace node_modules/hexo
+        run: |
+          rm -rf node_modules/hexo
+          ln -sf $GITHUB_WORKSPACE node_modules/hexo
+        working-directory: .tmp-hexo-theme-unit-test
```
SukkaW, line 46: IMHO I prefer running those setup inside `benchmark.js`, so that we could run benchmark on our local machine as well (simply `node benchmark.js` instead of manually setup the benchmark environment).

segayuu: OK:) If you can use git command in github actions runner, that is better. I'll give it a try.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,35 @@
...
+          restore-keys: |
+            ${{ runner.OS }}-npm-cache
+      - name: Running benchmark
+        run: |
+          echo "============== Hexo Benchmark ==============" > benchmark-result-${{ matrix.node-version }}.txt
+          node test/benchmark.js >> benchmark-result-${{ matrix.node-version }}.txt
+      - uses: actions/upload-artifact@v2
+        with:
+          name: benchmark-results
+          path: benchmark-result-${{ matrix.node-version }}.txt
```
SukkaW, line 35: We could post a comment at PR to show the benchmark result.

https://github.com/unsplash/comment-on-pr

@hexobot could be utilized.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,36 @@
+name: Benchmark
+
+on:
+  push:
+    branchs: [ master ]
```
curbengh, line 5: `branches`

---

I wonder if it's possible to:

```
on:
  push:
    branches: [ master ]
    paths:
      - 'libs/**'
```

both `branches` and `paths` must match (equivalent to AND).

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,36 @@
...
+          restore-keys: |
+            ${{ runner.OS }}-npm-cache
+      - name: Running benchmark
+        run: |
+          echo "============== Hexo Benchmark ==============" > benchmark-result-${{ matrix.node-version }}.txt
+          node test/benchmark.js >> benchmark-result-${{ runner.OS }}-${{ matrix.node-version }}.txt
+      - uses: actions/upload-artifact@v2
+        with:
+          name: benchmark-results
+          path: benchmark-result-${{ runner.OS }}-${{ matrix.node-version }}.txt
```
SukkaW, line 36: IMHO, we shouldn't redirect the result into a text file and upload to `artifact`. Personally I hate download a file just to see the benchmark result.

.github/workflows/profiling.yml
```diff
@@ -0,0 +1,59 @@
...
+          rm -rf node_modules/hexo
+          ln -sf $GITHUB_WORKSPACE node_modules/hexo
+        working-directory: hexo-theme-unit-test
+      - name: Start test run
+        run: npx --no-install hexo clean
+        working-directory: hexo-theme-unit-test
+      - name: Generating flamegraph
+        run: |
+          npm install 0x --silent
+          node ./node_modules/.bin/0x --name ${{ matrix.node-version }} --output-dir "${GITHUB_WORKSPACE}/0x" -- node ./node_modules/.bin/hexo g;
```
SukkaW, line 54: If we run `0x` directly in the GitHub Action, we could utilized `npx`:

```
npx 0x
```

.github/workflows/benchmark.yml
curbengh: This benchmark result is the same as the current one?

```
Cold processing
...
Hot processing
...
Another Cold processing
...
```

If that's the case, I agree it's better to just stdout.

SukkaW: @curbengh Yes. still `console.table`.

.github/workflows/profiling.yml
```diff
@@ -0,0 +1,59 @@
...
+        working-directory: hexo-theme-unit-test
+      - uses: actions/upload-artifact@v2
+        with:
+          name: profiling-results
+          path: ${{ github.workspace }}/0x/${{ matrix.node-version }}.html
```
SukkaW, line 59: Without `surge.sh` we could use GitHub Pages instead.

PR #4418, opened 2020-07-15
perf(tag): rendering optimization

SukkaW:
## What does it do?

Escape swig tag inside code block only once.

## How to test

```sh
git clone -b post-tag-render-optimize https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/extend/tag.js
+++ b/lib/extend/tag.js
@@ -5,6 +5,7 @@ const { cyan, magenta, red } = require('chalk');
 const { Environment } = require('nunjucks');
 const Promise = require('bluebird');
 const rSwigRawFullBlock = /{% *raw *%}/;
+const rCodeTag = /<code[^<>]*>[\s\S]+?<\/code>/g;
 const escapeSwigTag = str => str.replace(/{/g, '&#123;').replace(/}/g, '&#125;');
 
 class NunjucksTag {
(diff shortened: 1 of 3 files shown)


lib/extend/tag.js
```diff
@@ -5,7 +5,8 @@ const { cyan } = require('chalk');
 const { Environment } = require('nunjucks');
 const Promise = require('bluebird');
 const placeholder = '\uFFFC';
-const rPlaceholder = /(?:<|&lt;)!--\uFFFC(\d+)--(?:>|&gt;)/g;
+const rPlaceholder = /<!--\uFFFC(\d+)-->/g;
```
curbengh, line 8: I wonder why the previous regex matches `&lt;` `&gt;`? I thought `<>` is only escaped in [code block](https://github.com/hexojs/hexo-util/blob/cd6dcfda34f39cbf992c1e20bb1d4d5570ec7b3c/lib/highlight.js#L104)?

SukkaW: @curbengh IMHO, the regexp here might be copied from `post.js` in which the `<>` will be escaped by post renderer (like marked.js).

But inside tag rendering process, there will be no escape at all.

SukkaW:
@seaoak @curbengh @stevenjoezhang The PR is ready for review now.

SukkaW:
> WFM

@curbengh A lot of test cases has been added during pre-hexo-5.0.0 development and I am sure most of the case has been covered in unit test now.

Issue #4109, opened 2020-01-29
More detailed information for "Template render error: (unknown path)"


TingxunShi:
## Check List

Please check followings before submitting a new feature request.

- [ ] I have already read [Docs page](https://hexo.io/docs/)
- [ ] I have already searched existing issues

## Feature Request

Currently I am stucking with a very common issue: hexo failed to generate pages, gave the following information 
```
[myname@myname-pc blog_src]$ hexo g
INFO  Start processing
FATAL Something's wrong. Maybe you can find the solution here: https://hexo.io/docs/troubleshooting.html
Template render error: (unknown path)
  Error: expected end of comment, got end of file
    at Object._prettifyError (/home/myname/Projects/blog_src/node_modules/nunjucks/src/lib.js:36:11)
    at Template.render (/home/myname/Projects/blog_src/node_modules/nunjucks/src/environment.js:542:21)
    at Environment.renderString (/home/myname/Projects/blog_src/node_modules/nunjucks/src/environment.js:380:17)
    at /home/myname/Projects/blog_src/node_modules/hexo/lib/extend/tag.js:120:48
    at tryCatcher (/home/myname/Projects/blog_src/node_modules/bluebird/js/release/util.js:16:23)
    at Function.Promise.fromNode.Promise.fromCallback (/home/myname/Projects/blog_src/node_modules/bluebird/js/release/promise.js:209:30)
    at Tag.render (/home/myname/Projects/blog_src/node_modules/hexo/lib/extend/tag.js:120:18)
    at Object.onRenderEnd (/home/myname/Projects/blog_src/node_modules/hexo/lib/hexo/post.js:288:20)
    at /home/myname/Projects/blog_src/node_modules/hexo/lib/hexo/render.js:64:19
    at tryCatcher (/home/myname/Projects/blog_src/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/home/myname/Projects/blog_src/node_modules/bluebird/js/release/promise.js:547:31)
    at Promise._settlePromise (/home/myname/Projects/blog_src/node_modules/bluebird/js/release/promise.js:604:18)
    at Promise._settlePromise0 (/home/myname/Projects/blog_src/node_modules/bluebird/js/release/promise.js:649:10)
    at Promise._settlePromises (/home/myname/Projects/blog_src/node_modules/bluebird/js/release/promise.js:729:18)
[...]

Issue #3859, opened 2019-11-13
Misleading Nunjucks error message


TingxunShi:
## Check List

Please check followings before submitting a new issue.

- [x] I have already read [Docs page](https://hexo.io/docs/) & [Troubleshooting page](https://hexo.io/docs/troubleshooting)
- [x] I have already searched existing issues and they are not help to me
- [x] I examined error or warning messages and it's difficult to solve
- [x] Using [the latest](https://github.com/hexojs/hexo/releases) version of Hexo (run `hexo version` to check)
- [x] Node.js is higher than 8.6.0

## Question

Error message below is shown (**I'm sure that I have filled in `url: ` field in my hexo `_config.yml`**)

```
[myname@myname-pc blog_src]$ hexo g
INFO  Start processing
INFO  Files loaded in 378 ms
ERROR Render HTML failed: about/index.html
TypeError: Cannot read property 'replace' of null
    at Hexo.externalLinkFilter (/home/myname/Projects/blog_src/node_modules/hexo/lib/plugins/filter/after_render/external_link.js:45:15)                                                                                                                               
    at Hexo.tryCatcher (/home/myname/Projects/blog_src/node_modules/bluebird/js/release/util.js:16:23)
    at Hexo.<anonymous> (/home/myname/Projects/blog_src/node_modules/bluebird/js/release/method.js:15:34)
    at /home/myname/Projects/blog_src/node_modules/hexo/lib/extend/filter.js:60:50
    at tryCatcher (/home/myname/Projects/blog_src/node_modules/bluebird/js/release/util.js:16:23)
    at Object.gotValue (/home/myname/Projects/blog_src/node_modules/bluebird/js/release/reduce.js:166:18)
    at Object.gotAccum (/home/myname/Projects/blog_src/node_modules/bluebird/js/release/reduce.js:155:25)
    at Object.tryCatcher (/home/myname/Projects/blog_src/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/home/myname/Projects/blog_src/node_modules/bluebird/js/release/promise.js:547:31)
    at Promise._settlePromise (/home/myname/Projects/blog_src/node_modules/bluebird/js/release/promise.js:604:18)
    at Promise._settlePromiseCtx (/home/myname/Projects/blog_src/node_modules/bluebird/js/release/promise.js:641:10)
[...]

Issue #1771, opened 2016-02-23
Better error message


leesei:
It helps the user identify and fix errors if we 
- throw sensible errors
- print warnings more verbosely.

Do comment if you found some area that needs improving.

---

**EDIT** I have a WIP branch here:
https://github.com/leesei/hexo/commits/better-error-message

To use this branch:

```
npm i leesei/hexo#better-error-message
```

leesei:
#### slug not found

Hexo fails silently upon this:
https://github.com/hexojs/hexo/issues/1562
https://github.com/hexojs/hexo/issues/1594

leesei:
#### `_config.yaml`

YAML is quite strict in its format and Hexo does not report YAML parsing error properly.
We should catch parsing error early, log and abort the operation.

https://github.com/hexojs/hexo-deployer-git/issues/15
https://github.com/hexojs/hexo/issues/1263
https://github.com/hexojs/hexo/issues/11

add resource like [Online YAML Parser](http://yaml-online-parser.appspot.com/)

leesei:
#### Missing theme

#1658

leesei:
#### Better trace of invalid tag

Currently the stack track (in grey, which is hard to read) is logged
However the file causing the error and the erroneous tag are not mentioned

Try adding this to any file in `_post`:
```
{% codeblock %}
alert('Hello World!');
{% codeblock %}
```

tomByrer:
Thank you for looking into these.

modkzs:
The error message that hexo s or g print is very bad. It doesn't mention which file, row and columns that cause generating failure. So every time I have to spend long time finding what error is

leesei:
I have a WIP branch here:
https://github.com/leesei/hexo/commits/better-error-message

leesei:
#### Stack trace color

Currently the stack track is logged in grey, which is hard to read

> Done, see https://github.com/hexojs/hexo-log/pull/1

nickdesaulniers:
did this ever get merged?  Seeing an unknown tag error, and while I can grep for the tag, a readable error message rather than a stack trace that listed the source file would be :100:

leesei:
ping @tommy351, please merge the PR for hexo-log first.

NoahDragon:
Both changes are merged, thank you. 
https://github.com/hexojs/hexo-log/pull/1
https://github.com/hexojs/hexo-cli/pull/13

PR #4420, opened 2020-07-15
feat(tag): show source of the error & beautify

SukkaW:
## What does it do?

#3859 #4109 #1771

- Show the source file that contains error
  `source` is available as a property of `options` that passed to `tag.render`.
- Beautify the Error
  - The source of error will be colored in magenta while the error type will be in red.

## How to test

```sh
cd hexo
npm install
npm test
```

## Screenshots

image

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/extend/tag.js
+++ b/lib/extend/tag.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const { stripIndent } = require('hexo-util');
-const { cyan } = require('chalk');
+const { cyan, magenta, red } = require('chalk');
 const { Environment } = require('nunjucks');
 const Promise = require('bluebird');
 const placeholder = '\uFFFC';
@@ -129,7 +129,7 @@ const LINES_OF_CONTEXT = 5;
 
 const getContext = (lines, errLine, location, type) => {
   const message = [
-    location + ' ' + type,
+    location + ' ' + red(type),
     cyan('    =====               Context Dump               ====='),
     cyan('    === (line number probably different from source) ===')
   ];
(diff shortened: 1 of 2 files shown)


lib/extend/tag.js
```diff
@@ -158,14 +158,17 @@ const getContext = (lines, errLine, location, type) => {
...
   if (!match) return err;
   const errLine = parseInt(match[1], 10);
   if (isNaN(errLine)) return err;
 
   // trim useless info from Nunjucks Error
-  const splited = err.message.replace('(unknown path)', '').split('\n');
+  const splited = err.message.replace('(unknown path)', () => {
+    if (source) return magenta(source);
+    return '';
+  }).split('\n');
```
dailyrandomphoto, line 171: Instead of a function, static string will be better.
```js
const formatedSource = source ? magenta(source) : '';
const splited = err.message.replace('(unknown path)', formatedSource).split('\n');
```

lib/extend/tag.js
```diff
@@ -222,20 +225,23 @@ class Tag {
...
 
-  render(str, options, callback) {
+  render(str, options = {}, callback) {
     if (!callback && typeof options === 'function') {
       callback = options;
       options = {};
     }
 
+    // Get path of post from source
+    const source = options.source ? options.source : '';
```
dailyrandomphoto, line 235:
```diff
- const source = options.source ? options.source : '';
+ const source = options.source || '';
```

Issue #4433, opened 2020-07-22
Private posts with hexo


jehy:
## Check List

Please check followings before submitting a new feature request.

- [X] I have already read [Docs page](https://hexo.io/docs/)
- [X] I have already searched existing issues

## Feature Request

There is a plugin for hiding private posts - `hexo-hide-posts` - but it does't hide post if it is accessed directly, post is left in repo data and can be easily found and also I think that it should be a part of base hexo itself. It should be really simple - hexo should just skip reading this post if it has a private flag.

Of cause, it means that post will only be readable via source file, but I think that it's exactly what we need for private posts.

We would need to add a notice that private posts need private hexo repos, but this is also fine and clear.

stevenjoezhang:
You can use drafts: https://hexo.io/docs/writing#Drafts

jehy:
@stevenjoezhang Using drafts is pretty much indirect, messy and dangerous way. Real drafts will be in the same place as private posts, without any difference, and it is possible to accidentially publish them.

jehy:
Added a simple PR which implements this feature.

curbengh:
> Private posts are not drafts, that's completely another feature.

but https://github.com/hexojs/hexo/pull/4435 is functionally similar to drafts? My understanding of a private post is that the post is still rendered, but a reader (pre-approved by the blogger) needs to login to view it.

---

btw, a post/page is skipped whenever it has an underscore prefix `_foo.md`, this is another way of setting an article as draft; except for `_posts`, any files inside a folder with an underscore prefix are skipped.

jiangtj:
Hexo won't read outside the source folder file, you may be able to create a `secret` folder outside. 

> a post/page is skipped whenever it has an underscore prefix

Or underlined folders, the articles inside will also be ignored, e.g. `_secret`

jiangtj:
I still don't understand why the function of hiding is needed.

If you don’t need the article to be rendered and displayed, you can create a new private repo for storage.

curbengh:
> https://github.com/hexojs/hexo/issues/4433#issuecomment-678865598
> If you don’t need the article to be rendered and displayed, you can create a new private repo for storage.

I believe that's what OP meant

> https://github.com/hexojs/hexo/issues/4433#issue-663563013
> We would need to add a notice that private posts need private hexo repos

---

One idea is to utilise to have a front-matter like:

```
---
title: a private post
[...]

Issue #3328, opened 2018-11-05
[feature request] Consideration of using async / await


segayuu:
## For feature request

[The native promise after node v10 is fast enough](https://github.com/petkaantonov/bluebird#note
).
Since `async` / `await` is supported in node v7 or later, it can be used when support of node v6 LTS (maintenance) is turned off.
Using `async` / `await` tends to make the code simple, and sequential asynchronous is faster than `bluebird` on [the `bluebird` side benchmark](https://github.com/petkaantonov/bluebird/tree/v3.5.2/benchmark) (in the case of parallel, now `bluebird` is faster).

NoahDragon:
Agree. Performance is always the key improvement for Hexo. IMHO, this is not a feature, more like an enhancement. Anyway, thanks for bringing this up.

YoshinoriN:
I agree. Same as other hexo repositories.
IMO we can drop node v6 after 2019 Apr.

curbengh:
Just started working on this. A few observations:

- [`Bluebird.asCallback`](http://bluebirdjs.com/docs/api/ascallback.html) is to support callback style. Most Hexo API supports callback, so this can't be dropped for a foreseeable future.
- [`Promise.finally`](https://node.green/#ES2018-features-Promise-prototype-finally) is only supported in Node 10.3+, so need to wait until Node 8 is dropped.

SukkaW:
https://blog.kuzzle.io/bluebird-vs-native-vs-async/await-state-of-promises-performances-in-2019

> **TL;DR:**

> In Node.js 10, it is impossible to do without Bluebird in critical code areas given the performance of its promises compared to native promises and async/await. 
> In Node.js 12, the performance of async/await is comparable to that of Bluebird, so we can use async/await throughout Kuzzle core.

And according to the hexo's benchmark, currently I see no performance gained through replacing `bluebird` with async / await.

curbengh:
The benchmark linked by OP only shows async/await being slightly faster than bluebird (in sequential), the difference looks to be within margin of error.

I still believe bluebird is faster currently and for a foreseeable future; even our benchmark doesn't show it's faster in Node 13. Perhaps there may be bottleneck somewhere that either approach (bluebird/native) doesn't make any difference.

At the same time, async/await is significantly more readable, so if the perf is not affected, let's use it. especially for unit tests, where perf doesn't matter much.

dailyrandomphoto:
[...]

PR #4435, opened 2020-07-22
#4433: implement private pages and posts

jehy:
## What does it do?

It solves issue https://github.com/hexojs/hexo/issues/4433 and adds `private:true` option to posts. Private posts are excluded from generation.

PR may look a little frightening, but it's just an early return on `private:true` flag. Had to use async/await for it, otherwise it would be ugly to break from promise chain. So indent was removed and diff is a little messy.

## How to test

```sh
cd hexo
npm install
npm test
```
Then add `private:true` to any post/page, remove db.json and launch blog. Post/Page will disappear from everywhere.

## Pull request tasks

- [X] Passed the CI test.


lib/plugins/processor/asset.js
```diff
@@ -7,7 +7,7 @@ const { extname, relative } = require('path');
 const { Pattern } = require('hexo-util');
 
 module.exports = ctx => {
-  function processPage(file) {
+  async function processPage(file) {
```
SukkaW, line 10: Avoid using `async` here. The generation performance has been dropped by 30% after the PR.

lib/plugins/processor/asset.js
```diff
@@ -28,65 +28,64 @@ module.exports = ctx => {
...
+    if (!data.layout && output !== 'html' && output !== 'htm') {
+      data.layout = false;
+    }
 
-      if (doc) {
-        return doc.replace(data);
-      }
+    // FIXME: Data may be inserted when reading files. Load it again to prevent
+    // race condition. We have to solve this in warehouse.
+    const doc2 = Page.findOne({source: path});
```
SukkaW, line 82: There is no need to rename `doc` to `doc2` here.

lib/plugins/processor/post.js
```diff
@@ -22,7 +22,7 @@ const preservedKeys = {
 };
 
 module.exports = ctx => {
-  function processPost(file) {
+  async function processPost(file) {
```
SukkaW, line 25: Avoid using `async` here. The generation performance has been dropped by 30% after the PR.

jehy: That's strange, how  did you check it? There should not be suck an impact - at least, not at node 12+.

jehy:
@SukkaW I reimplemented those changes without async and added test cases.

lib/plugins/processor/post.js
```diff
@@ -131,18 +134,19 @@ module.exports = ctx => {
...
+          if (doc) {
+            return doc.replace(data);
+          }
+          return Post.insert(data);
+        }).then(doc => Promise.all([
+          doc.setCategories(categories),
+          doc.setTags(tags),
+          scanAssetDir(doc)
+        ]));
+    });
```
SukkaW, line 149: I see no necessary of changes here.

jehy: @SukkaW I had to make those changes to break out of promise chain. Otherwise, after 

```js
      if (data.private) {
        return;
      }
```
we would go to
```js
.then(doc => Promise.all([
      doc.setCategories(categories),
      doc.setTags(tags),
      scanAssetDir(doc)
    ]));
```
doc would be undefined and everything would break apart.

SukkaW: @jehy I see. It makes sense.
```suggestion
      return Promise.resolve(Post.findOne({source: file.path}))
        .then(doc => {
          if (doc) return doc.replace(data);

          return Post.insert(data);
        }).then(doc => Promise.all([
          doc.setCategories(categories),
          doc.setTags(tags),
          scanAssetDir(doc)
        ]));
    });
```

It might be possible to reduce a tick during `Promise` creation.

jehy: I suppose that won't make impact since the number of ticks doesn't change. But it may confuse future developers - they may think that `Post.findOne` was put into `Promise.resolve` because that's an async function.

jehy:
> Will the current draft feature enough for "private post" (instead of introducing a new private property)

Yup, creating additional property may seem an overhead, but using drafts is somehow messy and dangerous, as I told in related issue before.

> Whether we should exclude the post during process or even before process

I suppose we can't exclude a post before we read the file. But maybe I'm mistaken somewhere in the hexo flow.

May we should call someone for additional opinion?

jehy:
Anyone here? :)

lib/plugins/processor/post.js
curbengh: https://github.com/hexojs/hexo/issues/3328#issuecomment-666896317

Unit tests have been (mostly) [refactored](https://github.com/hexojs/hexo/pulls?q=is%3Apr+async%2Fawait+is%3Aclosed+) to async/await, refactoring hexo core has been considered before (https://github.com/hexojs/hexo/pull/3952, https://github.com/hexojs/hexo/pull/3970, https://github.com/hexojs/hexo/pull/4019), but we've decided to retain Bluebird promise for a foreseeable time due to its unmatched performance.

jehy: Perfomance isn't so unmatched in 12-14 versions of node. And promises themselves are just a tiny portion of performance which is often only visible in synthetic benchmarks. Using async await could let you write cleaner code and really raise performance. Also, it seems like soon there will be no developers who know about promises :)

lib/plugins/processor/asset.js
```diff
@@ -33,6 +33,9 @@ module.exports = ctx => {
       file.read()
     ]).spread((stats, content) => {
       const data = yfm(content);
+      if (data.private) {
```
curbengh, line 36: `data.publish === false` as we discussed in https://github.com/hexojs/hexo/issues/4433#issuecomment-679838640.

PR #4436, opened 2020-07-24
perf(external_link): faster regexp & condition shorthand

SukkaW:
## What does it do?

- Improve regexp: require less step for matching
- Adjust condition for shorthand optimization

## How to test

```sh
git clone -b perf-external-link https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/filter/after_post_render/external_link.js
+++ b/lib/plugins/filter/after_post_render/external_link.js
@@ -13,11 +13,11 @@ function externalLinkFilter(data) {
     return;
   }
 
-  data.content = data.content.replace(/<a\s+(?:[^<>]+\s)?href=["']([^<>"']+)["'][^<>]*>/gi, (str, href) => {
-    if (/target=/gi.test(str) || !isExternalLink(href, url, external_link.exclude)) return str;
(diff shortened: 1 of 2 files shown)


lib/plugins/filter/after_post_render/external_link.js
```diff
@@ -13,11 +13,11 @@ function externalLinkFilter(data) {
     return;
   }
 
-  data.content = data.content.replace(/<a\s+(?:[^<>]+\s)?href=["']([^<>"']+)["'][^<>]*>/gi, (str, href) => {
-    if (/target=/gi.test(str) || !isExternalLink(href, url, external_link.exclude)) return str;
+  data.content = data.content.replace(/<a(?:\s+?|\s+?[^<>]+)?href=["']([^<>"']+)["'][^<>]*>/gi, (str, href) => {
+    if (!isExternalLink(href, url, external_link.exclude) || /target=/i.test(str)) return str;
```
SukkaW, line 17: Most renderer plugin won't add `target=`. By moving `isExternalLink` forward unnecessary regex matching could be avoided.

Issue #3677, opened 2019-08-18
Replace cheerio with native API


curbengh:
## Feature Request

From the [benchmark](https://github.com/hexojs/hexo/issues/3663) (performed by @SukkaW) we can see that [cheerio](https://github.com/cheeriojs/cheerio) can cause significant performance hit.

While cheerio is not necessary for the core functions of Hexo, it is currently utilized by several default plugins:

- [x] ~~[meta_generator.js](https://github.com/hexojs/hexo/blob/master/lib/plugins/filter/meta_generator.js)~~
- [x] ~~[external_link.js](https://github.com/hexojs/hexo/blob/master/lib/plugins/filter/after_post_render/external_link.js)~~
- [ ] [toc.js](https://github.com/hexojs/hexo/blob/master/lib/plugins/helper/toc.js)
- [x] ~~[open_graph](https://github.com/hexojs/hexo/blob/master/lib/plugins/helper/open_graph.js)~~

We are hoping to re-implement them without depending on cheerio, if possible.

Related PRs:
- meta_generator: ~~https://github.com/hexojs/hexo/pull/3669~~ ~~https://github.com/hexojs/hexo/pull/3671~~
- open_graph: ~~https://github.com/hexojs/hexo/pull/3670~~ ~~https://github.com/hexojs/hexo/pull/3679~~ ~~https://github.com/hexojs/hexo/pull/3680~~
- external_link: ~~https://github.com/hexojs/hexo/pull/3685~~
- toc: https://github.com/hexojs/hexo/pull/3850

Related issue:
- https://github.com/hexojs/hexo/pull/550

curbengh:
Can this be part of [the roadmap](https://github.com/hexojs/hexo/issues/2492)?

SukkaW:
Cheerio use `perse5` now and use `htmlparser2` before.

There is a [benchmark](https://github.com/AndreasMadsen/htmlparser-benchmark) between NodeJS side dom parser: https://travis-ci.org/AndreasMadsen/htmlparser-benchmark

![image](https://user-images.githubusercontent.com/40715044/63223519-62c30700-c1e9-11e9-9d60-02c311e4cb68.png)

curbengh:
That means cheerio@1.x is slower than cheerio@0.x.

---

After a glance through the [htmlparser2](https://github.com/fb55/htmlparser2), I wonder if we can use it directly, instead of via cheerio; not sure if that can bring perf benefit.

SukkaW:
[...]

PR #4442, opened 2020-07-25
refactor(meta_generator): no longer ignore empty <head>

SukkaW:
## What does it do?

Ignoring empty `<head>` tag is firstly introduced in https://github.com/hexojs/hexo/pull/3315, in which `cheerio@1` will automatically append `<html><head><body>`. After #3677, cheerio has been dropped from `dependencies`. And cheerio in `devDependencies` has been pinned at `0.22.0` as well. So I just assume there will be no empty `<head>` tag any more.

The PR removes empty `<head>` detection which results in less generation time.

## How to test

```sh
git clone -b meta-generator-empty https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/test/scripts/filters/meta_generator.js
+++ b/test/scripts/filters/meta_generator.js
@@ -37,32 +37,6 @@ describe('Meta Generator', () => {
     should.not.exist(metaGenerator('<head><link><meta content="foo" name="generator"></head>'));
   });
 
-  it('ignore empty head tag', () => {
-    const content = '<head></head><head><link></head><head></head>';
-    hexo.config.meta_generator = true;
-    const result = metaGenerator(content);
-
-    const $ = cheerio.load(result);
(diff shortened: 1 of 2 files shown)


test/scripts/filters/meta_generator.js
```diff
@@ -37,32 +37,6 @@ describe('Meta Generator', () => {
...
-    const result = metaGenerator(content);
-
-    const $ = cheerio.load(result);
-    $('meta[name="generator"]').should.have.lengthOf(1);
-
-    const expected = '<head></head><head><link><meta name="generator" content="Hexo '
-      + hexo.version + '"></head><head><link></head>';
-    result.should.eql(expected);
-  });
-
```
curbengh, line 65: suggest to keep them, to be safe.

PR #4477, opened 2020-08-13
refactor(backtick_code): remove "~~~" support

SukkaW:
## What does it do?

From `rBacktick` it appears that hexo supports 2 syntax:

````
``` & ~~~
````

However, `~~~` syntax is never documented. And by removing it, the unit test could still be passed.

I am wonder `~~~` is introduced for nested code block highlight:

````
~~~
```
Look! This is an example of nested code block.
```
~~~
````

But Hexo already has a `codeblock` tag plugin, nested code block shouldn't be a problem.

The performance has been improved by 2%.

## How to test

```sh
git clone -b perf-faster-backtick-code-filter https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/filter/before_post_render/backtick_code_block.js
+++ b/lib/plugins/filter/before_post_render/backtick_code_block.js
@@ -2,7 +2,7 @@
 
 let highlight, prismHighlight;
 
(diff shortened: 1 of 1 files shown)


stevenjoezhang:
> A code fence is a sequence of at least three consecutive backtick characters (`) or tildes (~). (Tildes and backticks cannot be mixed.) A fenced code block begins with a code fence, indented no more than three spaces.

Ref https://spec.commonmark.org/0.29/#fenced-code-blocks

PR #4512, opened 2020-09-01
chore: use example.com for example domain

YoshinoriN:
## What does it do?

We should stop to use `yoursite.com`.
`yoursite.com` is a real company's domain.

We can use `example.com` instead of `yoursite.com`.

> [Reserved Example Second Level Domain Names](https://tools.ietf.org/html/rfc2606#section-3)

## How to test

```sh
cd hexo
npm install
npm test
```

## Screenshots

NONE

## Pull request tasks

- [ ] Passed the CI test.

### DIFF
--- a/lib/hexo/default_config.js
+++ b/lib/hexo/default_config.js
@@ -9,7 +9,7 @@ module.exports = {
   language: 'en',
   timezone: '',
   // URL
-  url: 'http://yoursite.com',
+  url: 'http://example.com',
   root: '/',
   permalink: ':year/:month/:day/:title/',
   permalink_defaults: {},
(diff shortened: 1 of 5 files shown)


SukkaW:
Agree. `example.com` & `example.org` is managed by IANA and is defined in RFC.

SukkaW:
Is the docs being updated?

SukkaW (approved):

YoshinoriN:
@SukkaW 

Yes :)

I have already submit...

* https://github.com/hexojs/site/pull/1541
* https://github.com/hexojs/hexo-starter/pull/46

Not yet submit https://github.com/hexojs/hexo-generator-alias.
I will submit after merge or reject https://github.com/hexojs/hexo-generator-alias/pull/55

Issue #4437, opened 2020-07-24
[Proposal] Migrate benchmark to GitHub Action


SukkaW:
- [x] Add `benchmark.js` to GitHub Action and remove it from Travis CI
- [x] Rewrite `profiling.sh` into `profiling.js` (And profiling will stay on Travis CI)
- [x] Find a new place to deploy the flamegraph (Maybe the JSON format of performance data as well)
- [x] Add `profiling.js` to GitHub Action along with new deployments, then remove the profiling from Travis CI
- [ ] Use @hexobot to post comment about the performance & flamegraph url under every PR & commits (just like coverall)

curbengh:
https://github.com/hexojs/hexo/pull/4413

PR #4514, opened 2020-09-02
chore/ci: move benchmark to gh action

SukkaW:
## What does it do?

Closes #4413.

The PR aims to gradually migrate from Travis CI to GitHub Actions. It is the first part of #4437.

## How to test

Just see if GitHub Action passed.

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -7,6 +7,8 @@ on:
   pull_request:
     branches:
       - master
+    paths-ignore:
+      - 'test/scripts/**'
 
 jobs:
   benchmark:
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,33 @@
+name: Benchmark
+
+on:
+  push:
+    paths:
+      - 'lib/**'
+  pull_request:
+    branches:
+      - master
+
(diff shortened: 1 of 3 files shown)


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,33 @@
+name: Benchmark
+
+on:
+  push:
+    paths:
+      - 'lib/**'
+  pull_request:
+    branches:
+      - master
```
curbengh, line 9: is `paths: ['lib/**']` going to be applied to pull_request too?

SukkaW: No.

And it is my desired behavior: PR like this one (which modifies the CI config and `benchmark.js`) should trigger benchmark as well.

curbengh: what about `paths-ignore: ['test/**']`? Some PRs only update the unit test, so it shouldn't be necessary to run the benchmark. The [doc](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-ignoring-paths) mentions the workflow will not run if only matches 'test/**', it will still run if there is any change to other folder.

curbengh (approved):

Issue #4556, opened 2020-10-14
Draft assets not rendered


darekkay:
## Check List

Please check followings before submitting a new issue.

- [x] I have already read [Docs page](https://hexo.io/docs/) & [Troubleshooting page](https://hexo.io/docs/troubleshooting)
- [x] I have already searched existing issues and they are not help  to me
- [x] I examined error or warning messages and it's difficult to solve
- [x] Using [the latest](https://github.com/hexojs/hexo/releases) version of Hexo (run `hexo version` to check)
- [x] Node.js is higher than [minimum required version](https://hexo.io/docs/#Minimum-required-Node-js-version)

## Expected behavior

Assets should be displayed for draft posts.

## Actual behavior

Assets are not displayed for draft posts.

This is caused by https://github.com/hexojs/hexo/pull/3489
I've got it working by reverting this change locally. While this PR removes draft assets from the generated folder, it also removes them from a draft build, which is not the expected behavior.

## How to reproduce?

1. Create a draft post
2. Add an asset
3. Include the asset: `{% asset_img image.png %}`
4. Run `hexo server --draft`

## Is the problem still there under "Safe mode"?

N/A (posts do not seem to be processed in safe mode)

## Environment & Settings

**Node.js & npm version(`node -v && npm -v`)**

```
v12.16.3
6.14.4
```

**Hexo and Plugin version(`npm ls --depth 0`)**

```
+-- del@6.0.0
+-- ejs@3.1.5
+-- fs-extra@9.0.1
+-- hexo@5.2.0
+-- hexo-auto-canonical@0.1.1
+-- hexo-browsersync@0.3.0
+-- hexo-generator-category@1.0.0
+-- hexo-generator-feed@3.0.0
+-- hexo-generator-index@2.0.0
+-- hexo-generator-sitemap@2.1.0
+-- hexo-generator-tag@1.0.0
+-- hexo-hide-posts@0.1.1
+-- hexo-html-minifier@1.0.0
+-- hexo-renderer-ejs@1.0.0
+-- hexo-renderer-marked@3.3.0
+-- hexo-renderer-stylus@2.0.1
+-- hexo-server@2.0.0
+-- highlight.js@10.2.1
+-- lodash@4.17.20
+-- prettier@1.19.1
`-- recursive-copy@2.0.11
```

darekkay:
This issue is not stale. The PR is open and ready to be reviewed.

PR #3489, opened 2019-03-15
skip assets of unpublished posts and delete them if exist

DaemondShu:
## What does it do?
Fixed issue #3478. 

## How to test

```sh
git clone -b BRANCH https://github.com/USER/hexo.git
cd hexo
npm install
npm test
```

## Screenshots
- the result before adding the unittest
image
The result is the same as the original master branch on my machine.

- the result after adding the unittest
image

- code coverage summary
image

## Pull request tasks

- [x] Add test cases for the changes.
- [x] Passed the CI test.

PR #4563, opened 2020-10-23
Generate draft assets in draft mode

darekkay:
## What does it do?

Since #3489, no draft assets are generated. This is the desired behavior when not rendering draft posts. But when working on drafts, the according assets are never rendered. This change checks if draft posts should be rendered, and if this is the case, the assets will not be deleted.

## How to test

I validated this fix by

- Running `hexo clean && rm -rf public/ && hexo generate` - Draft assets are not generated (as before)
- Running `hexo clean && rm -rf public/ && hexo generate --draft` - Draft assets are generated
- Running `hexo server --draft` - Draft assets are generated
- Adjusting the existing unit test

Fixes: #4556

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/processor/post.js
+++ b/lib/plugins/processor/post.js
@@ -145,6 +145,21 @@ module.exports = ctx => {
     ]));
   }
 
+  function shouldSkipAsset(post, asset) {
+    if (!ctx._showDrafts()) {
+      if (post.published === false && asset) {
+        // delete existing draft assets if draft posts are hidden
+        asset.remove();
+      }
+      if (post.published === false) {
+        // skip draft assets if draft posts are hidden
+        return true;
+      }
+    }
+
+    return asset !== undefined; // skip already existing assets
+  }
+
   function scanAssetDir(post) {
     if (!ctx.config.post_asset_folder) return;
 
@@ -164,8 +179,7 @@ module.exports = ctx => {
       const id = join(assetDir, item).substring(baseDirLength).replace(/\\/g, '/');
       const asset = PostAsset.findById(id);
 
-      if (asset) return post.published === false ? asset.remove() : undefined; // delete if already exist
-      else if (post.published === false) return undefined; // skip assets for unpulished posts and
+      if (shouldSkipAsset(post, asset)) return undefined;
 
       return PostAsset.save({
         _id: id,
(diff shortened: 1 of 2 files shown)


darekkay:
The failing jobs do not seem to be related to my changes

botreview (approved):
LGTM

jiangtj:
image

Can you merge the latest master? It fix some Action issues

darekkay:
I've rebased my branch onto the latest master. But now I can't see any GitHub actions being executed anymore.

lib/plugins/processor/post.js
```diff
@@ -164,8 +164,10 @@ module.exports = ctx => {
       const id = join(assetDir, item).substring(baseDirLength).replace(/\\/g, '/');
       const asset = PostAsset.findById(id);
 
-      if (asset) return post.published === false ? asset.remove() : undefined; // delete if already exist
-      else if (post.published === false) return undefined; // skip assets for unpulished posts and
+      if (!ctx._showDrafts()) {
```
jiangtj, line 167: Please make sure to return undefined when the asset exists.
```
if (asset) return undefined;
```

During the running of hexo s, if the source file is modified, this code will still be executed. At this time, the result of PostAsset.findById(id) is not undefined

jiangtj (changes requested):

jiangtj:
> I've rebased my branch onto the latest master. But now I can't see any GitHub actions being executed anymore.

I also don't understand why 😂

lib/plugins/processor/post.js
darekkay: I thought that `asset.remove()` doesn't return anything (= undefined), but I have rewritten the code to return undefined explicitly.

darekkay:
I'm happy if someone wants to take over the 0.05% drop in test coverage.

lib/plugins/processor/post.js
jiangtj: Maybe what I said is not clear 😂 

```js
const asset = PostAsset.findById(id);
```

The result of this line of code has two cases, one with a value, and one with `undefined`. When there is a value, we need to return `undefined` in order to skip the following `PostAsset.save()`

But in your modification, if `ctx._showDrafts()` is true, then no matter what the result of `PostAsset.findById(id)` is,  `PostAsset.save()` is executed and not skipped.

jiangtj: > I thought that `asset.remove()` doesn't return anything (= undefined), but I have rewritten the code to return undefined explicitly.

Yesh, I also think the code here should be rewritten, GJ

darekkay: @jiangtj Thanks, I think I finally got it 😄

I've refactored the code and commented all the code routes.

jiangtj (approved):

ylbeethoven:
I am surprised this fix is not included in 5.4.0. Can anyone please add this fix to the new version?

PR #4578, opened 2020-11-14
list_tags: span element & custom class for label

noraj:
## What does it do?

When `style` is not `list` with `show_count` the pseudo HTML is

```html
<a class="a">
    raw label
    <span class="count">count</span>
</a>
```

I implemented a way to customize the label too to something like

```html
<a class="a">
    <span class="label">raw label</span>
    <span class="count">count</span>
</a>
```

This is 100% backward compatible, for style = list there is not change, and if `option.class.label` is not set there is no span to not break the actual behavior. The new customization only appear when  `option.class.label` is set.

## How to test

```sh
git clone -b list_tags/label https://github.com/noraj/hexo.git
cd hexo
npm install
npm test
```

Put this in a template:

```pug
list_tags({show_count: true, class: {a: 'tempor', count: 'dolor', label: 'lorem'}, style: false, separator: ''})
```

## Screenshots

My use case was for Bulma CSS [addons tags](https://bulma.io/documentation/elements/tag/#list-of-tags) that is not possible without 2 spans.

I use this

```pug
list_tags({orderby: 'name', order: 1, show_count: true, class: {a: '', count: 'tag is-primary', label: 'tag is-dark'}, style: false, separator: ''})
```

image

## Pull request tasks

- [x] Passed the CI test. (yes error are not due to my code introduction)
- [x] documentation https://github.com/hexojs/site/pull/1564

### DIFF
--- a/lib/plugins/helper/list_tags.js
+++ b/lib/plugins/helper/list_tags.js
@@ -14,7 +14,7 @@ function listTagsHelper(tags, options) {
   const { style = 'list', transform, separator = ', ', suffix = '' } = options;
   const showCount = Object.prototype.hasOwnProperty.call(options, 'show_count') ? options.show_count : true;
   const classStyle = typeof style === 'string' ? `-${style}` : '';
-  let className, ulClass, liClass, aClass, countClass;
+  let className, ulClass, liClass, aClass, labelClass, countClass, labelSpan;
   if (typeof options.class !== 'undefined') {
     if (typeof options.class === 'string') {
       className = options.class;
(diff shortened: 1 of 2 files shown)


lib/plugins/helper/list_tags.js
```diff
@@ -25,13 +25,19 @@ function listTagsHelper(tags, options) {
     ulClass = options.class.ul || `${className}${classStyle}`;
     liClass = options.class.li || `${className}${classStyle}-item`;
     aClass = options.class.a || `${className}${classStyle}-link`;
+    labelClass = options.class.label || `${className}${classStyle}-label`;
     countClass = options.class.count || `${className}${classStyle}-count`;
+
+    labelSpan = !!Object.prototype.hasOwnProperty.call(options.class, 'label');
```
SukkaW, line 31:
```suggestion
    labelSpan = Object.prototype.hasOwnProperty.call(options.class, 'label');
```

IMHO `!!` is not necessary.

noraj: I course there are not and I never put them, that's again a post-commit hook. Those are dangerous and it's not the first time of of those hook mess with the code.

SukkaW (approved):

PR #4581, opened 2020-11-27
expose `escape_html` helper method for string manipulation to templates

awwong1:
## What does it do?

Expose the `escape_html` utility function from [https://github.com/hexojs/hexo-util/blob/master/lib/escape_html.js](https://github.com/hexojs/hexo-util/blob/master/lib/escape_html.js) so it can be called in templates.

## How to test

Testing may not be necessary? Covered in [https://github.com/hexojs/hexo-util/blob/master/test/escape_html.spec.js](https://github.com/hexojs/hexo-util/blob/master/test/escape_html.spec.js) already.

Other `hexo-util` functions, like `strip_html`, `trim`, `titlecase`, `word_wrap`, `truncate` are untested.

## Pull request tasks

- [ ] Passed the CI test.

### DIFF
--- a/lib/plugins/helper/format.js
+++ b/lib/plugins/helper/format.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const { stripHTML, wordWrap, truncate } = require('hexo-util');
+const { stripHTML, wordWrap, truncate, escapeHTML } = require('hexo-util');
 const titlecase = require('titlecase');
 
 exports.strip_html = stripHTML;
@@ -14,3 +14,6 @@ exports.word_wrap = wordWrap;
 exports.wordWrap = wordWrap;
 
 exports.truncate = truncate;
+
+exports.escape_html = escapeHTML;
+exports.escapeHTML = escapeHTML;
--- a/lib/plugins/helper/index.js
+++ b/lib/plugins/helper/index.js
@@ -15,13 +15,14 @@ module.exports = ctx => {
 
   helper.register('search_form', require('./search_form'));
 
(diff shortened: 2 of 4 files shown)


SukkaW (approved):
LGTM!

SukkaW (changes requested):
Well, would you mind adding unit tests?

awwong1:
> Well, would you mind adding unit tests?

@SukkaW Sure, I can add them if you feel it is necessary.
The unit tests will basically be ports of these for the relevant exposed functions: https://github.com/hexojs/hexo-util/tree/master/test

SukkaW:
> Sure, I can add them if you feel it is necessary.

@awwong1 

`escapeHTML` has its own test cases in hexo-util, thus a simple test case in `hexo` will be fine (it is used to improve the test coverage, which is calculated by how many lines of the code being executed during the test).

SukkaW (approved):

Android-KitKat:
Nice! `escape_html` is very useful in templates.

SukkaW:
> eg: Generate HTML Attribute.

@Android-KitKat 

That is very useful. But for now (since Hexo won't release a new version very soon), you can implement the helper by your self through Hexo API:

```js
// escape_html_helper.js
const { escapeHTML } = require('hexo-util');
hexo.extend.helper.register('escape_html', (str) =>  escapeHTML(str));
```

And put the file under your theme's `scripts` directory. This actually registers a helper with the name `escapa_html`. With the file, you can use `escape_html` in your theme right now (without waiting for a newer version of Hexo).

PR #4616, opened 2021-01-20
feat: handle config.root is not exist

jiangtj:
## What does it do?

See this description about `url` and `root`, https://github.com/hexojs/hexo-starter/blob/3d337683d271aa28e5bebd31dd35ba12a7a2be82/_config.yml#L15

We can know that  `url` and `root` are closely related, so why do we configure it twice? `root` can be generated by `url`

```yml
# You just need to configure it
url: https://hexo.io/foo
# Not required, optional
# root: /foo/
```

## How to test

```sh
cd hexo
npm install
npm test
```

## Pull request tasks

- [ ] Passed the CI test.

### DIFF
--- a/lib/hexo/load_config.js
+++ b/lib/hexo/load_config.js
@@ -26,6 +26,10 @@ module.exports = async ctx => {
   ctx.log.debug('Config loaded: %s', magenta(tildify(configPath)));
 
   ctx.config = deepMerge(ctx.config, config);
+  // If root is not exist, create it by config.url
+  if (!config.root) {
+    ctx.config.root = new URL(ctx.config.url).pathname + '/';
+  }
   config = ctx.config;
 
   validateConfig(ctx);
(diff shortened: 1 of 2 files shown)


lib/hexo/load_config.js
```diff
@@ -26,6 +26,10 @@ module.exports = async ctx => {
   ctx.log.debug('Config loaded: %s', magenta(tildify(configPath)));
 
   ctx.config = deepMerge(ctx.config, config);
+  // If root is not exist, create it by config.url
+  if (!config.root) {
+    ctx.config.root = new URL(ctx.config.url).pathname + '/';
```
stevenjoezhang, line 31: How about
```js
let { pathname } = new URL(ctx.config.url);
if (!pathname.endsWith('/')) pathname += '/';
ctx.config.root = pathname;
```

stevenjoezhang (approved):

Issue #4626, opened 2021-01-28
[excerpt.js] Use span instead of anchor element for better SEO performance


uuznn:
## Check List

Please check followings before submitting a new feature request.

- [x] I have already read [Docs page](https://hexo.io/docs/)
- [x] I have already searched existing issues

## Feature Request

When hexo converts  `<!-- more -->` to anchor, using `<a id="more"></a>`  element here may not be SEO friendly

https://github.com/hexojs/hexo/blob/30f34f59e475d5a75a171506c055a8096939408c/lib/plugins/filter/after_post_render/excerpt.js#L15

Try change to:
```html
<span id="more"></span>
```

Reference to https://developers.google.com/search/docs/advanced/guidelines/links-crawlable

Lighthouse Report: (or https://web.dev/measure/)

![屏幕截图 2021-01-28 120513](https://user-images.githubusercontent.com/76462613/106105818-909ed000-617f-11eb-8d21-0afb327492e1.png)

## Others

I have do this in my [fork branch](https://github.com/uuznn/hexo/tree/post-excerpt), but all test failed! XD

PR #4627, opened 2021-01-29
fix(excerpt): use span instead of anchor element for better SEO performance

stevenjoezhang:
## What does it do?

Issue resolved: https://github.com/hexojs/hexo/issues/4626

## How to test

```sh
git clone -b more https://github.com/hexojs/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.


SukkaW (approved):

SukkaW:
Notice that might be a Breaking Changes. Although no need for bumping a major version, it is still necessary to mention it in Change Log.

PR #4687, opened 2021-04-24
Prefix anchors by post path

### LINKED ISSUE #1302: Anchors in individual blog entries aren't made unique in the main page
A blog is composed of several entries. When I add footnotes to an entry, the anchors are numbered `#fn1`, `#fn2` etc. Hexo joins the entries in a single page but does not change the anchors. Thus, when I click on the 2nd footnote of an entry other than the top one, the browser navigates to the topmost (and hence, wrong) `#fn2` anchor.

PaperStrike:
## What does it do?
Resolve #1302 by using post path and `url_for` on post anchors.

## How to test
Test cases will be added later. Code below fails currently.

```sh
git clone -b prefix-anchor https://github.com/PaperStrike/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [ ] Passed the CI test.

### DIFF
--- a/lib/plugins/filter/after_post_render/anchor.js
+++ b/lib/plugins/filter/after_post_render/anchor.js
@@ -1,0 +1,11 @@
+'use strict';
+
+const { url_for } = require('hexo-util');
+const rAnchorHref = /(?<=<[^>]* href=")#[^"]+(?="[^>]*>)/g;
+
+function prefixAnchor(data) {
+  if (!data.path) return;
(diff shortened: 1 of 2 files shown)


stevenjoezhang:
I realize that this does not solve the problem, as the link on the homepage will still point to inside the article. When readers click the anchor on the homepage, it will jump (rather than scroll to the corresponding position on the homepage). We can make it unique by adding the post's hash to the anchor name when generating the `post.excerpt`.

PR #4714, opened 2021-06-07
feat(console-new): support default title from path

xu-song:
## What does it do?

Before this commit, we can create a post by
```bash
hexo new --path hexo/hexo-tutorial hexo-tutorial 
```
Hexo create a file `source/_posts/hexo/hexo-tutorial.md` with title `title: hexo-tutorial `

In many cases, the title has the same value of `basename of path`, this commit simplify above script
```bash
hexo new --path hexo/hexo-tutorial                # title=basename(path)  by default
```

## How to test

```sh
git clone -b patch-1 https://github.com/xu-song/hexo.git
```


stevenjoezhang (approved):
👍

PR #4723, opened 2021-06-26
Allow overriding landscape theme sidebar config

rayfoss:
Landscape provides a fully loaded config, there is no clean way to get a minimal config without **hard** forking the theme.

This allows a cleaner way over overriding theme configs.

#3890
#4154
https://github.com/hexojs/hexo-theme-landscape/issues/102

In most inheritance models, class methods are fully superseded by the extending class, pyYaml supports this model.

Alternatively we could use a database migration model, where the config has a theme version and schema version and merges are done based on that.

This PR just works with the present situation.

## How to test

```sh
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.
- [x] was written on a phone.

### DIFF
--- a/lib/hexo/load_theme_config.js
+++ b/lib/hexo/load_theme_config.js
@@ -27,8 +27,9 @@ module.exports = ctx => {
     // ctx.config.theme_config should have highest priority
     // If ctx.config.theme_config exists, then merge it with _config.[theme].yml
     // If ctx.config.theme_config doesn't exist, set it to _config.[theme].yml
+    const override = config.override || {} // Allow overriding upstream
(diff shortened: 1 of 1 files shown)


rayfoss:
It's clear this isn't enough. What's needed is a major update full of hooks and a two tier theme yml. One one internal and override-discouraged, one external and fully override able. As well as hooks for late-cascade scss styles. Same for a JavaScript module hook or 10.

Forking is always worse than override-able thing we can provide.

stevenjoezhang:
We need to update the default `_config.yml` of landscape and comment out some configuration options @hexojs/core

Issue #4430, opened 2020-07-19
Post Processor: about race condition


SukkaW:
## Check List

Please check followings before submitting a new issue.

- [x] I have already read [Docs page](https://hexo.io/docs/) & [Troubleshooting page](https://hexo.io/docs/troubleshooting)
- [x] I have already searched existing issues and they are not help to me
- [x] I examined error or warning messages and it's difficult to solve
- [x] Using [the latest](https://github.com/hexojs/hexo/releases) version of Hexo (run `hexo version` to check)
- [x] Node.js is higher than 8.6.0

## Question

According to https://github.com/hexojs/hexo/commit/9ba40f4dd5894c20905a5380020833da8275015a

> Parallel processing and parallel generating may cause race conditions. We have to solve this in warehouse.

And in https://github.com/hexojs/hexo/commit/9ba40f4dd5894c20905a5380020833da8275015a#diff-18267fb4b13330715876832e13d9697fR83 it says:

> FIXME: Data may be inserted when reading files. Load it again to prevent race condition. We have to solve this in warehouse.

@tommy351 Would you mind providing information about "race condition" and how to trigger it? I have no idea about how to reproduce it and don't know which warehouse function is the culprit.

## Others

By removing the line `const doc = Page.findOne({source: path});` the generation speed is about 20% faster.

tommy351:
I don't remember the detail at this moment. It may be related to file watch I guessed? I will try to look into the source code later.

SukkaW:
@tommy351 Thanks very much!

SukkaW:
unstale

PR #4743, opened 2021-07-21
Escape HTML  by default in list_tag

tomap:
escape the HTML in Tag values. It should be text, not HTML

This is a breaking change as some people might have been relying on this use case (having html in tags) 
and documentation should be updated to explain that when using function transform, value should be escaped if needed

### DIFF
--- a/test/scripts/helpers/list_tags.js
+++ b/test/scripts/helpers/list_tags.js
@@ -205,3 +205,41 @@ describe('list_tags', () => {
     ].join(''));
   });
 });
+
+describe('list_tags transform', () => {
+  const Hexo = require('../../../lib/hexo');
+  const hexo = new Hexo(__dirname);
+  const Post = hexo.model('Post');
+
+  const ctx = {
+    config: hexo.config
+  };
+
+  const listTags = require('../../../lib/plugins/helper/list_tags').bind(ctx);
+
+  before(async () => {
+    await hexo.init();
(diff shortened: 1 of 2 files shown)


tomap:
I don't know the plans for Hexo 6, but if we follow semver, we should publish this fix in Hexo 6

test/scripts/helpers/list_tags.js
```diff
@@ -205,3 +205,41 @@ describe('list_tags', () => {
...
+
+  const listTags = require('../../../lib/plugins/helper/list_tags').bind(ctx);
+
+  before(async () => {
+    await hexo.init();
+    const posts = await Post.insert([
+      {source: 'foo', slug: 'foo'}
+    ]);
+
+    // TODO: Warehouse needs to add a mutex lock when writing data to avoid data sync problem
```
SukkaW, line 226: See also https://github.com/hexojs/hexo/issues/4430

SukkaW (approved):

test/scripts/helpers/list_tags.js
tomap: I just copied this part from the tests above. Not sure how to fix it

SukkaW: I just add that comment to add a link back to the already opened issue (for later fix). The current workaround already looks good for this PR.

PR #4744, opened 2021-07-22
Feature: When execute filter's callback, add new parameters.

tcatche:
## What does it do?

When a callback to a registered filter is executed, the parameters passed to the filter are those that have been processed by other filters.  The filter executed later may be missing related information. 

For example, the internal hexo filter(`lib\plugins\filter\post_permalink.js`) get `Post Object`  as parameter. 
After this filter run, return a `permalink` and dismisses other `Post` properties.
And then, other filter after this will lose `Post` info and only get a  `permalink` parameter without any `Post` info:

```js
// before
hexo.extend.filter.register('post_permalink', function(data){
  // only `2021/07/22/hello` without other info about the post.
  console.log(data); 
  return data;
});
```

The code try to solve this, give a  new params to filter callback, The filter can get the origin info before other filter process:

```js
// after
hexo.extend.filter.register('post_permalink', function(data, post){
// get more params.
  console.log(data, post); 
  return data;
});
```

## How to test

```sh
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/extend/filter.js
+++ b/lib/extend/filter.js
@@ -62,6 +62,7 @@ class Filter {
     const ctx = options.context;
     const args = options.args || [];
 
+    args.push(data);
     args.unshift(data);
 
     return Promise.each(filters, filter => Reflect.apply(Promise.method(filter), ctx, args).then(result => {
@@ -78,6 +79,7 @@ class Filter {
     const ctx = options.context;
     const args = options.args || [];
 
+    args.push(data);
     args.unshift(data);
 
     for (let i = 0, len = filtersLen; i < len; i++) {
--- a/test/scripts/extend/filter.js
+++ b/test/scripts/extend/filter.js
@@ -94,13 +94,15 @@ describe('Filter', () => {
   it('exec()', async () => {
     const f = new Filter();
 
-    const filter1 = spy(data => {
+    const filter1 = spy((data, origin) => {
       data.should.eql('');
+      origin.should.eql('');
       return data + 'foo';
     });
 
-    const filter2 = spy(data => {
+    const filter2 = spy((data, origin) => {
       data.should.eql('foo');
+      origin.should.eql('');
       return data + 'bar';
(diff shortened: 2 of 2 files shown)


lib/extend/filter.js
```diff
@@ -62,6 +62,7 @@ class Filter {
     const ctx = options.context;
     const args = options.args || [];
 
+    args.push(data);
     args.unshift(data);
```
SukkaW, line 66:
```suggestion
    args.push(data);
    args.unshift(data);
```

If we already `unshift` the data as the first element of the `args`, why `push` it again?

SukkaW (changes requested):

lib/extend/filter.js
tcatche: Almost all filters cb will change the first elements of args, that means maybe one filter's callback may change an object element to a string, and the next callback can only get the string as "Function parameters", and lose other object info.

For example:  when use `ctx.execFilterSync('post_permalink', this, {context: ctx});`  trigger `post_permalink` cb(`this` point to a `Post Object`). The first filter callback register as `hexo.extend.filter.register('post_permalink', function(data) { return data.title });`  

And then the second filter callback can only get `post title` as the only "Function parameters", and it lost other info about the `Post Object`.

```js
// example1
hexo.extend.filter.register('post_permalink', function(data){
  return data;
});
```

In example1, the `filter1` parameter data is  `post title` and lose `Post Object` other properties.


**All of args content will be passed to filter's next callback as arguments.**


When push data to `args` again, This will as filter's callback in last parameter, and the second can still get the `Post Object` .

```js
// example2
hexo.extend.filter.register('post_permalink', function(data, post){
// do sth
});
```
In example2, the `post` parameter is from  `args.push(data);` 

So i use `args.push(data)` hope give other filter's callback more infomation to do sth.

PR #4771, opened 2021-09-07
feat: load hexo plugin in the theme's package.json

stevenjoezhang:
## What does it do?

Resolve https://github.com/hexojs/hexo-renderer-marked/issues/205#issuecomment-912917230
See also https://github.com/hexojs/hexo/pull/4112 https://github.com/hexojs/hexo/pull/2558

## How to test

```sh
git clone -b theme-pluginhttps://github.com/hexojs/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [ ] Passed the CI test.

### DIFF
--- a/lib/hexo/load_plugins.js
+++ b/lib/hexo/load_plugins.js
@@ -47,7 +47,7 @@ function loadModuleList(ctx, basedir) {
 function loadModules(ctx) {
   return Promise.map([ctx.base_dir, ctx.theme_dir], basedir => loadModuleList(ctx, basedir))
     .then(([hexoModuleList, themeModuleList]) => {
-      return Object.entries(Object.assign(hexoModuleList, themeModuleList));
+      return Object.entries(Object.assign(themeModuleList, hexoModuleList));
     })
     .map(([name, path]) => {
       // Load plugins
--- a/lib/hexo/load_plugins.js
+++ b/lib/hexo/load_plugins.js
@@ -45,7 +45,7 @@ function loadModuleList(ctx, basedir) {
 }
 
 function loadModules(ctx) {
-  return Promise.all([ctx.base_dir, ctx.theme_dir].map(basedir => loadModuleList(ctx, basedir)))
+  return Promise.map([ctx.base_dir, ctx.theme_dir], basedir => loadModuleList(ctx, basedir))
     .then(([hexoModuleList, themeModuleList]) => {
       return Object.entries(Object.assign(hexoModuleList, themeModuleList));
     })
--- a/lib/hexo/load_plugins.js
+++ b/lib/hexo/load_plugins.js
@@ -24,7 +24,7 @@ function loadModuleList(ctx, basedir) {
       const deps = Object.keys(json.dependencies || {});
       const devDeps = Object.keys(json.devDependencies || {});
 
-      return deps.concat(devDeps);
+      return basedir === ctx.base_dir ? deps.concat(devDeps) : deps;
     });
   }).filter(name => {
     // Ignore plugins whose name is not started with "hexo-"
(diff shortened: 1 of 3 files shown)


stevenjoezhang:
Both the dependencies and devDependencies in the Hexo `package.json` will be loaded as hexo plugins
Only the dependencies in the theme `package.json` will be loaded (if installed)

If you have any other suggestions, please reply here.

jiangtj:
devDependencies are used in the current project for test or build, I think they should be ignored.

I am more concerned about peerdependencies. In npm, it should be used to declare the host environment dependency, which is suitable for our plug-ins. However, the actual performance of peerdependencies is different in npm6 (node ​​14) and npm7. Npm6 will not automatically install peerdependencies dependencies.

https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependencies

I think we should consider support for peerdependencies in future versions

stevenjoezhang:
I do agree to ignore devDependencies, this will revert https://github.com/hexojs/hexo/pull/2558

SukkaW (commented):
Any unit test can be added?

SukkaW:
Also, maybe we need a separated Node.js 12 PR before this can be merged.

stevenjoezhang:
> Also, maybe we need a separated Node.js 12 PR before this can be merged.

PR opened: https://github.com/hexojs/hexo/pull/4779

SukkaW:
@stevenjoezhang Would you mind rebasing the PR with the latest `master` branch? Otherwise, GitHub Action's Lint task will never pass.

SukkaW (approved):

test/scripts/hexo/load_plugins.js
```diff
@@ -67,6 +64,10 @@ describe('Load plugins', () => {
 
   after(() => fs.rmdir(hexo.base_dir));
 
+  afterEach(() => {
+    createPackageFile('hexo-another-plugin');
```
stevenjoezhang, line 68: This line will reset Hexo package.json, to make sure in the following unit tests, `hexo-plugin-test` plugin is loaded from theme's package.json, not Hexo package.json

lib/hexo/load_plugins.js
```diff
@@ -37,22 +37,26 @@ function loadModuleList(ctx) {
...
-    const path = ctx.resolvePlugin(name);
-
-    // Load plugins
-    return ctx.loadPlugin(path).then(() => {
-      ctx.log.debug('Plugin loaded: %s', magenta(name));
-    }).catch(err => {
-      ctx.log.error({err}, 'Plugin load failed: %s', magenta(name));
+  return Promise.map([ctx.base_dir, ctx.theme_dir], basedir => loadModuleList(ctx, basedir))
+    .then(([hexoModuleList, themeModuleList]) => {
+      return Object.entries(Object.assign(hexoModuleList, themeModuleList));
```
jiangtj, line 50: It seems that the plugin defined in the theme will override the one under hexo. The priority in the theme is higher, but the plugin under hexo is added by the user. I think it should have a higher priority.

stevenjoezhang: There are two different considerations:
- The priority in the Hexo `package.json` is higher: If the theme is not maintained, the users will be able to override the outdated plugins defined in theme `package.json`
- The priority in the theme is higher: The plugin in Hexo `package.json` may not be compatible with the theme, and the theme could provide a proper version for the users

jiangtj: > There are two different considerations:

Yesh, but I prefer the first one. After all, if you use the plugin defined by the theme, you only need to remove the relevant plugins in your package.json. In the second option, I will not be able to cover the plugins inside the theme.

Or take a poll? 😂

SukkaW (approved):

jiangtj (approved):
:+1:

PR #4780, opened 2021-09-20
refactor/perf(post): use state machine to escape swig tag

SukkaW:
## What does it do?

Replace regexp with the state machine.

## How to test

```sh
git clone -b post-render-escape-swig https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [ ] Passed the CI test.

### DIFF
--- a/lib/hexo/post.js
+++ b/lib/hexo/post.js
@@ -12,9 +12,6 @@ const { parse: yfmParse, split: yfmSplit, stringify: yfmStringify } = require('h
 const preservedKeys = ['title', 'slug', 'path', 'layout', 'date', 'content'];
 
 const rHexoPostRenderEscape = /<hexoPostRenderCodeBlock>([\s\S]+?)<\/hexoPostRenderCodeBlock>/g;
-const rSwigVarAndComment = /{[{#][\s\S]+?[}#]}/g;
-const rSwigFullBlock = /{% *(\S+?)(?: *| +.+?)%}[\s\S]+?{% *end\1 *%}/g;
-const rSwigBlock = /{%[\s\S]+?%}/g;
 
 const rSwigPlaceHolder = /(?:<|&lt;)!--swig\uFFFC(\d+)--(?:>|&gt;)/g;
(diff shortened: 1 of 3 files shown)


lib/hexo/post.js
```diff
@@ -28,30 +25,153 @@ const _restoreContent = cache => (_, index) => {
...
-    this.cache = [];
+    this.stored = [];
   }
 
   restoreAllSwigTags(str) {
-    const restored = str.replace(rSwigPlaceHolder, _restoreContent(this.cache));
-    if (restored === str) return restored;
-    return this.restoreAllSwigTags(restored); // self-recursive for nexted escaping
+    const restored = str.replace(rSwigPlaceHolder, _restoreContent(this.stored));
+    return restored;
```
SukkaW, line 48: The self-recursive escaping is firstly introduced by @seaoak in #2321 trying to mitigate a `back_tick_code` inside a `swig tag` issue when the year is 2016. Since then, Hexo's post rendering mechanics has changed dramatically, and when @2321 being merged in 2019, there is no longer required self-recursive as nested escaping will never happen. After this change, the unit test case still passed.

stevenjoezhang (approved):

stevenjoezhang:
The rendering of left curly brackets has been changed in this pull request, which will affect the display of mathematical formulas (`\begin{equation}` becomes `\beginequation}`)

Unit tests: https://github.com/hexojs/hexo/tree/post-render

SukkaW:
> The rendering of left curly brackets has been changed in this pull request, which will affect the display of mathematical formulas (`\begin{equation}` becomes `\beginequation}`)
> 
> Unit tests: https://github.com/hexojs/hexo/tree/post-render

Ouch. I thought I have solved the issue. I will look into it now.

stevenjoezhang:
There is another issue, the following markdown document will cause a rendering error
```
Disable rendering of Nunjucks tag `{{ }}`/`{% %}` and [tag plugins](/docs/tag-plugins) when enabled
```

See https://github.com/hexojs/site/runs/3680904140

PR #4781, opened 2021-09-21
fix(#1490): if `post_asset_folder` is set, restrict renderable files to default file extension

### LINKED ISSUE #1490: asset files were rendered when post_asset_folder set to true
This is the source structure:

``` bash
$ tree source/_posts/

source/_posts/
|-- hello-world.md
|-- test-post
|   |-- test_css.css
|   `-- test_js.js
`-- test-post.md
```

but when I executed `hexo generate`, the post in public folder looks like this:

``` bash
$ tree public/2015/09/08/test-post/

public/2015/09/08/test-post/
|-- index.html
|-- test_css
|   `-- index.html
|-- test_css.css
|-- test_js
|   `-- index.html
`-- test_js.js
```

It seems the css files and javascript files also were rendered.

kristofzerbe:
## What does it do?

It alleviates the problem on the rendering asset files as posts, if `post_asset_folder` is set to `true`, by restricting the rendered file types to the default new post type (`new_post_name`). 

Caveat: Doesn't work with MD files as assets, when your default file type is MD. 

It is not a final solution, as it would be better to identify to subfolder as asset folder in this stage of code execution.

## How to test

```sh
cd hexo
npm install
npm test
```

1. Set `post_asset_folder` to `true` in `_config.yml`
2. Create new post
3. Create subfolder (without leading underscore!) in newly created asset folder for the post
4. Copy a HTML, JS or CSS file into the subfolder
5. run `hexo generate`

None of the subfolders files should be rendered as post, as it will be in the current version. They will be treated as real assets.

### DIFF
--- a/lib/plugins/processor/post.js
+++ b/lib/plugins/processor/post.js
@@ -241,7 +241,14 @@ module.exports = ctx => {
 
       if (!result || isHiddenFile(result.path)) return;
 
+      //checks only if there is a renderer for the file type or if is included in skip_render
       result.renderable = ctx.render.isRenderable(path) && !isMatch(path, ctx.config.skip_render);
+
+      //if post_asset_folder is set, restrict renderable files to default file extension 
+      if (result.renderable && ctx.config.post_asset_folder) {
+        result.renderable = (extname(ctx.config.new_post_name) === extname(path));
+      }
+
       return result;
     }),
 
--- a/lib/plugins/processor/post.js
+++ b/lib/plugins/processor/post.js
@@ -241,10 +241,10 @@ module.exports = ctx => {
 
       if (!result || isHiddenFile(result.path)) return;
 
-      //checks only if there is a renderer for the file type or if is included in skip_render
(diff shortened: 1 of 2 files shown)


SukkaW (commented):
LGTM!

Would you mind adding unit test cases for the changes as well?

tomap:
@SukkaW maybe we could add the unit tests ourselves (@hexojs/core ) and fix the linting issue to move forward faster?

SukkaW:
> @SukkaW maybe we could add the unit tests ourselves (@hexojs/core ) and fix the linting issue to move forward faster?

Yeah, that would be fine.

yoshinorin:
@kristofzerbe 
Thank you :)

@hexojs/core 
I wrote the test code & fix lint.
Would you please review again?

Thank you :)

kristofzerbe:
Thank you for the test and lint... Happy New Year

tomap (approved):

PR #4791, opened 2021-10-02
fix(processor): remove race condition failsafe

### LINKED ISSUE #4430: Post Processor: about race condition

SukkaW:
## What does it do?

This closes #4430

## How to test

```sh
git clone -b fix-race-condition https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/processor/post.js
+++ b/lib/plugins/processor/post.js
@@ -163,10 +163,6 @@ function processPost(ctx, file) {
       data.permalink = undefined;
     }
 
-    // FIXME: Data may be inserted when reading files. Load it again to prevent
-    // race condition. We have to solve this in warehouse.
-    const doc = Post.findOne({source: file.path});
-
     if (doc) {
       return doc.replace(data);
     }
--- a/lib/plugins/processor/asset.js
+++ b/lib/plugins/processor/asset.js
@@ -96,10 +96,6 @@ function processPage(ctx, file) {
       data.layout = false;
     }
 
-    // FIXME: Data may be inserted when reading files. Load it again to prevent
-    // race condition. We have to solve this in warehouse.
-    const doc = Page.findOne({source: path});
-
     if (doc) {
       return doc.replace(data);
     }


lib/plugins/processor/post.js
```diff
@@ -163,11 +164,10 @@ function processPost(ctx, file) {
...
-    // FIXME: Data may be inserted when reading files. Load it again to prevent
-    // race condition. We have to solve this in warehouse.
-    const doc = Post.findOne({source: file.path});
-
     if (doc) {
+      if (file.type !== 'update') {
+        ctx.log.warn(`Trying to "create" ${magenta(file.path)}, but the file already exists!`);
+      }
       return doc.replace(data);
     }
```
SukkaW, line 172: There are only three reasons that a `doc` will already exist while the processor is trying to `create` a post:

- A race condition inside the warehouse (which seems to be fixed in warehouse@2.1.0 back in 2015)
- The `file.type` is `update`. in this case, there is no need to load the `doc` again.
- In some rare situation where two `file` object is created, with the same `id` and the same `create` type (E.g. inside our unit test when we trying to `create` 2 `file`s under the same `file.path`. This is not likely to happen in the real-world production). But Hexo can still deal with the issue by replacing `doc` directly. And I add a warning message for the case, and hopefully, someone who encounters the issue would report it to us.

test/scripts/processors/post.js
```diff
@@ -311,8 +311,10 @@ describe('post', () => {
     post.slug.should.eql('foo');
     post.published.should.be.true;
 
-    post.remove();
-    unlink(file.source);
+    return Promise.all([
+      post.remove(),
+      unlink(file.source)
+    ]);
```
SukkaW, line 317: Both `_Document.remove` method and `fs.unlink` method returns `Promise`.

stevenjoezhang (approved):

PR #4863, opened 2022-01-09
fix: plugin loading conflict with @vercel/nft

SukkaW:
## What does it do?

Hexo uses browserify's `resolve` to resolve a path of a plugin by default, which is incompatible with [`@vercel/nft`](https://github.com/vercel/nft). See also: https://github.com/vercel/nft/issues/257

So we use Node.js's built-in `require.resolve` by default (which is compatible with `@vercel/nft`), fallback to `resolve.sync` only when the module is not found (or other errors).

Since there are no other significant behavior differences (from hexo's usage) between `resolve.sync` and `require.resolve`, we should drop `resolve` from dependency in the future.

## How to test

```sh
git clone -b fix-plugin-resolve https://github.com/sukkaw/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/hexo/index.js
+++ b/lib/hexo/index.js
@@ -23,8 +23,8 @@ const Locals = require('./locals');
 const defaultConfig = require('./default_config');
 const loadDatabase = require('./load_database');
 const multiConfigPath = require('./multi_config_path');
-const { sync } = require('resolve');
 const { deepMerge, full_url_for } = require('hexo-util');
+let resolveSync; // = require('resolve');
 
 const libDir = dirname(__dirname);
 const dbVersion = 1;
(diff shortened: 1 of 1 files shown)


stevenjoezhang (approved):

tomap (approved):
I approve, but unable to fully understand this change and it's potential impacts

SukkaW:
> I approve, but unable to fully understand this change and its potential impacts

The API `require.resolve` is introduced in Node.js since 0.3.0, however it doesn't support limit scope resolving (only search dependency in the given directories' `node_modules`, in hexo's case is hexo's folder) back then. That's probably why package `resolve` (made by browserify) is used (it provides a `base_dir` option).

And since Node.js 8.9.0, `require.resolve` accepted the second argument, which enables to only search under specific given `paths`.

It shouldn't have any impact at all. And in case there is any, I still use the old `resolve` package as a fallback, trying to avoid a breaking change.

PR #4869, opened 2022-01-13
Fix js-yaml tags for v4.0.0+

marcofranssen:
Repair backwards compatibility https://github.com/nodeca/js-yaml/blob/master/migrate_v3_to_v4.md
with a bunch of plugins that use these yaml tags.

e.g.: https://github.com/lavas-project/hexo-pwa#options uses these yaml tags in the routes `!!js/regexp`

## What does it do?

Fixes backwards compatibility for some plugins that make use of js-yaml tags.

This broke since the hexo 5.4.0 release that bumps `js-yaml` to v4.0.0.

See: 
https://github.com/nodeca/js-yaml/blob/master/CHANGELOG.md#400---2021-01-03
and
https://github.com/nodeca/js-yaml/blob/master/migrate_v3_to_v4.md

for details.

> ⚠️ This fix is required for both the 5.x release as well the 6.x release.

### DIFF
--- a/lib/plugins/renderer/yaml.js
+++ b/lib/plugins/renderer/yaml.js
@@ -2,9 +2,10 @@
 
 const yaml = require('js-yaml');
 const { escape } = require('hexo-front-matter');
+const schema = yaml.DEFAULT_SCHEMA.extend(require('js-yaml-js-types').all);
 
 function yamlHelper(data) {
-  return yaml.load(escape(data.text));
+  return yaml.load(escape(data.text), { schema });
 }
 
 module.exports = yamlHelper;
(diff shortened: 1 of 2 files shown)


SukkaW (approved):
That's a good catch!

jerryc127:
cool

yoshinorin (approved):

yoshinorin:
@marcofranssen 
Thank you :)

marcofranssen:
@yoshinorin a pleasure. When can we expect a patch for the 6.x and 5.4.x versions?

yoshinorin:
@marcofranssen 
I'm going to publish `6.1.0` and `5.4.1` after merge #4877 and  #4876 :)

yoshinorin:
I just published `5.4.1`.

> https://www.npmjs.com/package/hexo

Regarding to `6.0.0`, please wait a while.

PR #4902, opened 2022-02-24
chore/test: use chai#should

SukkaW:
## What does it do?

This deprecates the old `require('chai/register-should')` which causes the test to fail.

## Pull request tasks

- [x] Passed the CI test.


yoshinorin (approved):
LGTM 👍

stevenjoezhang (approved):

Issue #4917, opened 2022-03-19
YAMLException: Specified list of YAML types (or a single Type object) contains a non-Type object.


noraj:
I upgraded my hexo dependencies:

```                                                                                                                                                                                                                                        
 hexo                       ^6.0.0  →  ^6.1.0                                                                                                                                                                                                
 hexo-renderer-markdown-it  ^5.0.0  →  ^6.0.1                                                                                                                                                                                                
 hexo-server                ^2.0.0  →  ^3.0.0
```

## Check List

- [x] I have already read [Docs page](https://hexo.io/docs/) & [Troubleshooting page](https://hexo.io/docs/troubleshooting)
- [x] I have already searched existing issues and they are not help  to me
- [x] I examined error or warning messages and it's difficult to solve
- [x] Using [the latest](https://github.com/hexojs/hexo/releases) version of Hexo (run `hexo version` to check)
- [x] Node.js is higher than [minimum required version](https://hexo.io/docs/#Minimum-required-Node-js-version)

## Expected behavior

Normal generation with `hexo g`.

## Actual behavior

```
$ hexo g
YAMLException: Specified list of YAML types (or a single Type object) contains a non-Type object.
    at /home/noraj/Dev/hackdoc/node_modules/hexo/node_modules/js-yaml/lib/schema.js:104:13
    at Array.forEach (<anonymous>)
    at Schema.extend (/home/noraj/Dev/hackdoc/node_modules/hexo/node_modules/js-yaml/lib/schema.js:102:12)
    at Object.<anonymous> (/home/noraj/Dev/hackdoc/node_modules/hexo/lib/plugins/renderer/yaml.js:5:36)
    at Module._compile (node:internal/modules/cjs/loader:1097:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1151:10)
    at Module.load (node:internal/modules/cjs/loader:975:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:999:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at module.exports (/home/noraj/Dev/hackdoc/node_modules/hexo/lib/plugins/renderer/index.js:15:16)
    at Hexo.init (/home/noraj/Dev/hackdoc/node_modules/hexo/lib/hexo/index.js:235:35)
[...]

PR #4929, opened 2022-04-04
fix(#4917): suppress YAMLException when load js-yaml (backport: #4927)

yoshinorin:
This PR is backport from #4927 
I'm going to release a patch version (`5.4.2`) after merge this PR.

Thank you :)

## Pull request tasks

- [ ] Passed the CI test.


SukkaW (approved):

SukkaW (changes requested):
It seems that js-yaml-types just use `String.prototype.matchAll` which is not yet supported until Node.js 16.

What about downgrade js-yaml version for Hexo 5.4.2?

PR #4931, opened 2022-04-05
chore(ISSUE_TEMPLATE): delete `Your theme _config.yml` section

yoshinorin:
## What does it do?

Honestly, we can't understand anything even if we confirm a theme config they are using.

## Pull request tasks

- [x] Passed the CI test.

## Others

IMHO, **unofficial** plugins setting is also.  (But, we can not be able anything to it soon.)


stevenjoezhang:
Yes, it will be much more helpful to let the user provide a GitHub repo containing the necessary files to reproduced the bug. It's hard to debug with only config files or plugins list.

SukkaW (approved):

PR #4973, opened 2022-05-18
let post_link use original post title as title attribute

ppwwyyxx:
## What does it do?

This PR modifies the `post_link` tag so that the "title" attribute will contain the actual post title.

I think this new behavior is better: 
* Same as the old behavior, the text will still show the title that users specified as arguments in `post_link` tag
* But once hovered, the tooltip will show full title of the actual post

In the old behavior:
* Once hovered, the tooltip will have the same content as the existing text, therefore the tooltip is not useful.

## Screenshots
The screenshot shows the new behavior, where tooltip text is the actual post title

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/tag/post_link.js
+++ b/lib/plugins/tag/post_link.js
@@ -27,6 +27,7 @@ module.exports = ctx => {
     if (!post) return error;
 
     let title = args.length ? args.join(' ') : post.title;
+    // Let attribute be the true post title so it appears in tooltip.
     const attrTitle = escapeHTML(post.title);
     if (escape === 'true') title = escapeHTML(title);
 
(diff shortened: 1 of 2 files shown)


ppwwyyxx:
The failures in github actions are:
```
Run marocchino/sticky-pull-request-comment@v2
Error: Request failed due to following response errors:
 - Resource not accessible by integration
```
this seems unrelaed to this PR

ppwwyyxx:
Can maintainers take a look at this simple PR? thanks

ppwwyyxx:
Can maintainers take a look? Thanks!

stevenjoezhang (approved):

renbaoshuo (approved):

PR #4998, opened 2022-06-17
Let post support multiple tags

JingeXC:
## What does it do?

When I call list_tags to display tags.Posts often display the same tags repeatedly.
eg, post A has tags 'html, css' and post B has tags 'html, javascript'.Then I call list_tags, it will return ['html,css', 'html,javascript'].

I think it returns ['html','css','javascript'] will be clearer and more reasonable.

So I did the following:
1.Cut the tags string from front-matter into an array.
2.Remove the same tags between posts in list_tags.

## Screenshots

before:
image
after:
image

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/helper/list_tags.js
+++ b/lib/plugins/helper/list_tags.js
@@ -53,6 +53,15 @@ function listTagsHelper(tags, options) {
   // Limit the number of tags
   if (options.amount) tags = tags.limit(options.amount);
 
+  // Remove the same tags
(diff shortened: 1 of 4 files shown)


stevenjoezhang:
But multiple tags are already supported.

```yml
tags:
  - html
  - javascript
```

Don't use comma to split them. Use YAML array.

JingeXC:
Thx!

PR #5053, opened 2022-09-07
Don't use data-uri for og:image

### LINKED ISSUE #5052: Data URI scheme for `og:image`
## Check List

## Feature Request

`open_graph` helper should not generate `og:image` of data-uri.

It's basically correct to output all image URLs in the article as `og:image`.

By the way, the "URL" in OGP must be `https://`, but `open_graph()` also handles Data URI schemes. As a result, many OGP crawlers seem to treat it as an error.

*NOTE: The Open Graph protocol / Types / URL*
https://ogp.me/#url

> All valid URLs that utilize the https:// or https:// protocols

This error should be prevented.

## Others

This issue is now reproduced on my blog.
https://no-hack-no.life/post/2022-08-01-want-ERD-to-become-more-popular/

The images used in the above URLs are all data-uri as follows:
1 my-post-entry

When this URL is loaded into Meta OGP Share Debugger, the following error occurs.
https://developers.facebook.com/tools/debug/?q=https%3A%2F%2Fno-hack-no.life%2Fpost%2F2022-08-01-want-ERD-to-become-more-popular%2F

2 result-of-share-debugger

All settings are [here](https://github.com/KentarouTakeda/blog), including package.json and _config.json.

KentarouTakeda:
## What does it do?

fix #5052 

## Pull request tasks

- [ ] Passed the CI test.

### DIFF
--- a/lib/plugins/helper/open_graph.js
+++ b/lib/plugins/helper/open_graph.js
@@ -115,7 +115,8 @@ function openGraphHelper(options = {}) {
     result += og('og:locale', localeToTerritory(language), false);
   }
 
-  images = images.map(path => new URL(path, url || config.url).toString());
+  images = images.map(path => new URL(path, url || config.url).toString())
+    .filter(url => !url.startsWith('data:'));
 
   images.forEach(path => {
     result += og('og:image', path, false);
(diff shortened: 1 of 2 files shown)


KentarouTakeda:
Some CI processes seem to be failing, but I don't know why.
(It seems to be an authentication problem, so the pull request itself may not be the cause)

Please tell me if there is anything I can do.

SukkaW (approved):
LGTM! Thanks for catching it!

PR #5064, opened 2022-09-17
refactor: drop `youtube` tag

yoshinorin:
## What does it do?

Breaking Changes

IMHO, an application should not depends on a specific service as possible. Actually, I never use embedded youtube on my blog. These tags should provide as a plugin and should not include any core module.

If we drop this, we have to announce and create a new plugin for current users.

## Screenshots

N/A

## Pull request tasks

- [x] Passed the CI test.


SukkaW:
Maybe we could implement those as a separate plugin, like `hexo-tag-embed`, so that anyone upgrading to Hexo 7 can still render the same result with their existing posts.

renbaoshuo (approved):

yoshinorin:
refs: https://github.com/hexojs/hexo/discussions/5113#discussion-4594136

PR #5092, opened 2022-11-01
refactor: migrate typescript

stevenjoezhang:
- [x] Pending https://github.com/hexojs/hexo-fs/pull/109
- [x] Pending https://github.com/hexojs/hexo/pull/5094
- [x] Pending https://github.com/hexojs/hexo/pull/5095
- [x] Pending https://github.com/hexojs/hexo-util/pull/294

## What does it do?

Issue resolved: https://github.com/hexojs/hexo/issues/4136

## Pull request tasks

- [ ] Add test cases for the changes.
- [ ] Passed the CI test.

PR #5094, opened 2022-11-02
refactor: prepare for migration to typescript

stevenjoezhang:
## What does it do?

Convert some `require`s to top level. They will be the static import after https://github.com/hexojs/hexo/pull/5092
Some lazyload `require`s e.g. `if (!highlight) highlight = require('hexo-util').highlight;` are left untouched.

And, found some possible bugs in #5092 
lib/plugins/console/generate.js 9ba40f4dd5894c20905a5380020833da8275015a
lib/hexo/scaffold.js 3323364f3a81b539f303f6dec4ae666bb3c34dcc

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/extend/index.js
+++ b/lib/extend/index.js
@@ -1,12 +1,16 @@
 'use strict';
 
-exports.Console = require('./console');
-exports.Deployer = require('./deployer');
-exports.Filter = require('./filter');
-exports.Generator = require('./generator');
-exports.Helper = require('./helper');
-exports.Injector = require('./injector');
-exports.Migrator = require('./migrator');
-exports.Processor = require('./processor');
-exports.Renderer = require('./renderer');
-exports.Tag = require('./tag');
+const Console = require('./console');
+const Deployer = require('./deployer');
+const Filter = require('./filter');
+const Generator = require('./generator');
+const Helper = require('./helper');
+const Injector = require('./injector');
+const Migrator = require('./migrator');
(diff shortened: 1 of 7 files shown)


lib/hexo/index.js
```diff
@@ -226,14 +235,14 @@ class Hexo extends EventEmitter {
...
-    require('../plugins/renderer')(this);
-    require('../plugins/tag')(this);
+    consolePlugin(this);
+    filterPlugin(this);
+    generatorPlugin(this);
+    helperPlugin(this);
+    injectorPlugin(this);
+    processorPlugin(this);
+    rendererPlugin(this);
+    tagPlugin(this);
```
SukkaW, line 245: This should leave untouched, to speed up initial booting.

lib/plugins/console/index.js
```diff
@@ -1,24 +1,34 @@
...
 
+const cleanConsole = require('./clean');
+const configConsole = require('./config');
+const deployConsole = require('./deploy');
+const generateConsole = require('./generate');
+const listConsole = require('./list');
+const migrateConsole = require('./migrate');
+const newConsole = require('./new');
+const publishConsole = require('./publish');
+const renderConsole = require('./render');
```
SukkaW, line 11: These should not be moved to the top, CLI commands should be loaded on demand.

SukkaW (approved):

Issue #5042, opened 2022-08-29
Tag Plugins post_link generate a Post not found error


DNACore:
## Check List

Please check followings before submitting a new issue.

- [x] I have already read [Docs page](https://hexo.io/docs/) & [Troubleshooting page](https://hexo.io/docs/troubleshooting)
- [x] I have already searched existing issues and they are not help  to me
- [x] I examined error or warning messages and it's difficult to solve
- [x] Using [the latest](https://github.com/hexojs/hexo/releases) version of Hexo (run `hexo version` to check)
- [x] Node.js is higher than [minimum required version](https://hexo.io/docs/#Minimum-required-Node-js-version)

## Expected behavior

`{% post_link somemarkdownfile %}` should generate a corrent link to the exsit markdown file.

## Actual behavior

it genrated a `post not found xxx` link.

## How to reproduce?

just write the post_link such as `{% post_link somemarkdownfile %}`

what's more:

I created a new hexo folder, and created a new post named `test.md`,
and wrote `{% post_link test %}` in hello-world.md.
run `hexo g`, it can genrate the correnct link.
and I move test.md to subfolder/test.md,
then run `hexo clean && hexo g`. it can't genrate the correct link, just said: `post not found`.
I moved the `test.md` just below the _post folder, then it can genrate the correct link again. 
**BUT for my original hexo folder, wherever i put the `somemarkdownfile.md`， it always said `post not found`.**

## Is the problem still there under "Safe mode"?

YES

## Environment & Settings

**Node.js & npm version(`node -v && npm -v`)**

```
v18.7.0
8.17.0
```

**Your site `_config.yml`** (Optional)

```yaml
```

**Hexo and Plugin version(`npm ls --depth 0`)**

```
hexo-site@0.0.0
├── hexo-cli@4.3.0
├── hexo-deployer-git@3.0.0
├── hexo-generator-archive@1.0.0
├── hexo-generator-category@1.0.0
├── hexo-generator-index@2.0.0
├── hexo-generator-sitemap@3.0.1
├── hexo-generator-tag@1.0.0
├── hexo-renderer-ejs@2.0.0
├── hexo-renderer-marked@5.0.0
├── hexo-renderer-stylus@2.1.0
├── hexo-server@3.0.0
└── hexo@6.2.0
```

**Your package.json `package.json`**

```json
{
  "name": "hexo-site",
  "version": "0.0.0",
  "private": true,
  "scripts": {
    "build": "hexo generate",
[...]

Issue #4778, opened 2021-09-16
post_link post not found error


hassiocn:
## Check List

Please check followings before submitting a new issue.

- [x] I have already read [Docs page](https://hexo.io/docs/) & [Troubleshooting page](https://hexo.io/docs/troubleshooting)
- [x] I have already searched existing issues and they are not help  to me
- [x] I examined error or warning messages and it's difficult to solve
- [x] Using [the latest](https://github.com/hexojs/hexo/releases) version of Hexo (run `hexo version` to check)
- [x] Node.js is higher than [minimum required version](https://hexo.io/docs/#Minimum-required-Node-js-version)

## Expected behavior
For instance: {% post_link how-to-bake-a-cake %}.

This will work as long as the filename of the post is how-to-bake-a-cake.md, even if the post is located at source/posts/2015-02-my-family-holiday and has permalink 2018/en/how-to-bake-a-cake

I have /source/_post/intro/1/intro_install.md
{% post_link intro-install 'intro install ' %} ,it shoud link to intro_install.md
## Actual behavior
after hexo g,hexo s, Post not found: intro-install intro install  on the page
## How to reproduce?

* Step1 mkdir /source/_post/intro/1/
* Step2 create file intro_install.md
* Step3 create file intro_index.md
* Step4 add {% post_link intro-install 'intro install ' %} to intro_index.md
* Step5 hexo g,hexo s
* Step6 explore the intro_index.html 

## Environment & Settings

**Node.js & npm version(`node -v && npm -v`)**

```
v16.6.2
7.20.3
```

**Your site `_config.yml`** (Optional)

```yaml
# Hexo Configuration
## Docs: https://hexo.io/docs/configuration.html
## Source: https://github.com/hexojs/hexo/

# Site
title: AI
subtitle: ''
description: ''
keywords: 'HomeAssitant HA '
author: 
language: zh-CN
timezone: 'Asia/Shanghai'

# URL
## Set your site url here. For example, if you use GitHub Page, set url as 'https://username.github.io/project'
url: http://hai-ai.gitee.io
permalink: :title.html
permalink_defaults:
pretty_urls:
  trailing_index: true # Set to false to remove trailing 'index.html' from permalinks
  trailing_html: true # Set to false to remove trailing '.html' from permalinks

# Directory
source_dir: source
public_dir: public
tag_dir: tags
archive_dir: archives
category_dir: categories
code_dir: downloads/code
i18n_dir: :lang
skip_render:

# Writing
[...]

PR #5114, opened 2022-11-24
feat(tags/post_link): search for both slug and title

stevenjoezhang:
## What does it do?

Issue resolved: #5042
Issue resolved: https://github.com/hexojs/hexo/issues/4778

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/tag/post_link.js
+++ b/lib/plugins/tag/post_link.js
@@ -8,7 +8,7 @@ const { postFindOneFactory } = require('./');
  * Post link tag
  *
  * Syntax:
- *   {% post_link slug [title] [escape] %}
+ *   {% post_link slug | title [title] [escape] %}
  */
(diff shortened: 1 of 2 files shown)


yoshinorin (commented):
Maybe LGTM. I'll debug this in this weekend.

yoshinorin (approved):
LGTM 👍 

---

BTW, the current implementation (before this PR) behavior is similar to many of ORM `first` method (like Rails's first, Laravel's first ...etc). Honestly, I don't like this. Because it can't find a post uniquely.

I know this behavior is generally in ORM. But, many hexo users may confusing.

stevenjoezhang:
Yes, the slug is unique, so there is no confusion. However, for the users, titles are more intuitive than the slug, so I believe this patch will improve usability when there're no duplicate titles.

PR #3624, opened 2019-07-13
Ignore categories / tags with zero posts

stevenjoezhang:
## What does it do?

Issue resolved: #3622

### Current behaviour
Categories / tags of draft are displayed / counted in non-draft generate.

In hexo-theme-next, we tried to filter out the categories / tags in the draft, but this caused performance issues.

More information:
theme-next/hexo-theme-next#42
theme-next/hexo-theme-next#87

## How to test

```sh
git clone -b master https://github.com/stevenjoezhang/hexo.git
cd hexo
npm install
npm test
```

## Pull request tasks

- [ ] Add test cases for the changes.
- [ ] Passed the CI test.

Issue #2579, opened 2017-05-23
[Question] Use Hexo for generate more than 10000 posts


mnlbox:
Hi guys,
I have a big list of posts that I converted to markdown files from my previous web site. My post count in `_posts` directory now is near 10000 posts.
I try `hexo server --debug` but it's not running after 15 minutes processing on a machine with 32 gig RAM and a SSD hard drive.
I also try `hexo clean && hexo generate` but it's also not finish after 25 minutes.
Now my question are:

1. Is it normal?
2. How we can speed up Hexo process?
3. Can we use Hexo for this scale site or It's for smaller sites?

NoahDragon:
1. Yes, it is a known performance issue on Hexo. But all static website generator has such issue. We are trying to improve it. https://github.com/hexojs/hexo/pull/550 
2. Here are several tips: 
- Run `hexo clean` only when the style or the theme is changed which may affect all the pages. The old posts are stored in database, if there is no global change the generating process will skip them.
- Use simple theme which does not contain many widgets.
- Don't use categories or tags.
- Remove unused Hexo plugins.
3. So far, we haven't encountered such amount of posts to process. But the performance is really varied by different themes, the number of categories/tags, and renderer engines.

mnlbox:
@NoahDragon Thanks for your reply.
I create simple theme based on bootstrap without any widgets. It's extra simple and just use some property from app _config (I put some global config in app _config rather than theme config) I don't know is it right or not?
I used many front-matter in my markdown page. (Category, Tags and two other custom attribute)
Each post has between 2 or 10 tags and just one category
My renderer engine is https://github.com/hexojs/hexo-renderer-marked

Can you make me some other suggestion based on this new information about my app?

rahil471:
I have been through the same problem. In the end, we decided to get rid of some hexo-plugins like. 
hexo-multiauthors, hexo-tag-generator, hexo-archives.  
And it's  not about the RAM, it has more to do with your CPU's. Hexo Generate is a CPU intensive task. 
@NoahDragon
If we could utilize multi-cores(if it doesn't) then maybe we can reduce the pain a bit.

mnlbox:
@rahid471 Yes I checked my system monitor and it seems Hexo only use 1 CPU core. I have 4 core and one of my core increased until 100% but others between 2% to 19%.

NoahDragon:
@mnlbox It's okay to put all configuration into the app _config.yml file, the theme configuration will fall back to the app when it is not set.
[...]

PR #5119, opened 2022-11-28
perf: reduce the number of traversals through posts

stevenjoezhang:
## What does it do?

Calling `post.tags[0].length` is slow, because Hexo will try to fetch all posts associated with one tag, which is unnecessary.
Tested with 8000 posts (copy https://github.com/SukkaLab/hexo-many-posts 16 times, with https://github.com/hexojs/hexo-starter), this patch will reduce the time of `hexo g` from ~17min to ~11min

TODO: `PostTag.find` is also slow. Need to replace it with a better data structure in the future.
See https://github.com/hexojs/hexo/issues/2579#issuecomment-1328065978
See also https://github.com/hexojs/hexo/pull/3624

## Pull request tasks

- [ ] Passed the CI test.

### DIFF
--- a/lib/hexo/index.js
+++ b/lib/hexo/index.js
@@ -200,15 +200,9 @@ class Hexo extends EventEmitter {
       return db.model('Page').find(query);
     });
 
-    locals.set('categories', () => {
-      // Ignore categories with zero posts
-      return db.model('Category').filter(category => category.length);
-    });
+    locals.set('categories', () => db.model('Category'));
 
-    locals.set('tags', () => {
-      // Ignore tags with zero posts
-      return db.model('Tag').filter(tag => tag.length);
-    });
+    locals.set('tags', () => db.model('Tag'));
 
     locals.set('data', () => {
       const obj = {};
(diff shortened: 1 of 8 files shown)


SukkaW (approved):

stevenjoezhang:
Although this patch can mitigate performance issues, Hexo still spends a lot of time dealing with post tags when the site has a lot of posts. 

As shown in the figure below, nearly half of the processing time is related to the tags.

截屏2022-11-28 23 11 19


It's difficult to make further optimizations for this issue with the current warehouse implementation. Perhaps we could consider supporting other database backends such as sqlite.

SukkaW:
> Although this patch can mitigate performance issues, Hexo still spends a lot of time dealing with post tags when the site has a lot of posts.
> 
> As shown in the figure below, nearly half of the processing time is related to the tags.
> 
> 截屏2022-11-28 23 11 19
> 
> It's difficult to make further optimizations for this issue with the current warehouse implementation. Perhaps we could consider supporting other database backends such as sqlite.

IMHO warehouse is still our best possible solution.

warehouse put all the data in the memory, which makes it faster for smaller websites (which most Hexo-powered sites are), while it might not scale well with larger sites.

PR #5145, opened 2023-01-12
perf(post): cache tags getter

SukkaW:
## What does it do?

The PR tries to resolve https://github.com/hexojs/hexo/pull/5119#issuecomment-1329284732

cc @stevenjoezhang 

## Screenshots

**Flamegraph before the PR**

image

**Flamegraph after the PR**

image

## Pull request tasks

- [x] Passed the CI test.


SukkaW:
@stevenjoezhang So it turns out the `post.tags` field is accessed more than once during rendering. Caching the getter could speed up the second access.

IMHO we could dig more into this, to see if we can prevent executing some functions over and over again.

renbaoshuo (approved):
👍 LGTM

Issue #2134, opened 2016-09-03
Won't generate with post_asset_folder on and permalink ending with ".html"


mogita:
## Environment Info

Node version(`node -v`)

```
v6.5.0
```

Your site `_config.yml`

```
# Hexo Configuration
## Docs: https://hexo.io/docs/configuration.html
## Source: https://github.com/hexojs/hexo/

# Site
title: 御宅型
subtitle: mogita 的技术博客
description: mogita 的技术博客
author: mogita
language: zh-CN
timezone: Asia/Shanghai

# URL
## If your site is put in a subdirectory, set url as 'http://yoursite.com/child' and root as '/child/'
url: http://otaku.mogita.com
root: /
permalink: :year/:month/:title.html
permalink_defaults:

# Directory
source_dir: source
public_dir: public
tag_dir: tags
archive_dir: archives
category_dir: categories
code_dir: downloads/code
i18n_dir: :lang
skip_render:

# Writing
new_post_name: :year-:month-:day-:title.md
default_layout: post
titlecase: false # Transform title into titlecase
external_link: true # Open external links in new tab
filename_case: 0
render_drafts: false
post_asset_folder: true
relative_link: false
future: true
highlight:
  enable: true
  line_number: true
  auto_detect: false
  tab_replace:

# Category & Tag
default_category: uncategorized
category_map:
tag_map:

# Date / Time format
## Hexo uses Moment.js to parse and display date
## You can customize the date format as defined in
## http://momentjs.com/docs/#/displaying/format/
date_format: YYYY-MM-DD
time_format: HH:mm:ss

# Pagination
## Set per_page to 0 to disable pagination
per_page: 10
pagination_dir: page

# Extensions
## Plugins: https://hexo.io/plugins/
## Themes: https://hexo.io/themes/
theme: clean-blog

#### Analytics
cnzz: true

feed:
  type: atom
  path: atom.xml
  limit: 20
  hub:
```

Your theme `_config.yml`

```
# Header
menu:
  关于: /about

# Title on top left of menu. Leave empty to use main blog title
menu_title: 

# URL of the Home page image
index_cover: /img/header.jpg

# Default post title
default_post_title: Untitled

# Comments. Choose one by filling up the information
comments:
  # Disqus comments
  disqus_shortname: otaku-mogita

# Google Analytics Tracking ID
google_analytics:

# Addthis ID
addthis:

# Social Accounts
twitter_url: https://twitter.com/mogita
twitter_handle:
facebook_url:
[...]

PR #5153, opened 2023-02-02
fix(post-asset): strip extensions better on permalink

KagamigawaMeguri:
## What does it do?
Fixes [#2134](https://github.com/hexojs/hexo/issues/2134)
Closes [#2881](https://github.com/hexojs/hexo/pull/2881)

A better way to strip extensions on permalink in post-asset.js
To some degree, it is possible to increase plugin compatibility.

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/models/post_asset.js
+++ b/lib/models/post_asset.js
@@ -19,8 +19,7 @@ module.exports = ctx => {
 
     // PostAsset.path is file path relative to `public_dir`
     // no need to urlescape, #1562
-    // strip /\.html?$/ extensions on permalink, #2134
-    // compatible with hexo-abbrlink when `post_asset_folder: true`
+    // strip extensions better on permalink, #2134
     return join(dirname(post.path), post.slug, this.slug);
   });
 
--- a/lib/models/post_asset.js
+++ b/lib/models/post_asset.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const { Schema } = require('warehouse').default;
-const { join } = require('path');
+const { join, dirname } = require('path');
 
 module.exports = ctx => {
   const PostAsset = new Schema({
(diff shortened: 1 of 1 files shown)


stevenjoezhang (changes requested):
Maybe you can consider modifying the hexo-abbrlink plugin instead?

KagamigawaMeguri:
> 

Others have consulted the hexo-abbrlink plugin developer, and he thinks it's hexo's problem.
Details in https://github.com/rozbo/hexo-abbrlink/issues/33

yoshinorin:
I'm not yet checking the behavior of `post_asset.js`. 

But I think maybe we can merge this. Because this PR passes all of the current tests. Also, it includes https://github.com/hexojs/hexo/pull/2881 for fixes #2134.

But we should change the PR title and source code comment. We should not change the source code for third-party plugins behavior.

KagamigawaMeguri:
> I'm not yet checking the behavior of `post_asset.js`.
> 
> But I think maybe we can merge this. Because this PR passes all of the current tests. Also, it includes #2881 for fixes #2134.
> 
> But we should change the PR title and source code comment. We should not change the source code for third-party plugins behavior.

Thank you for your patience and guidance! Please review again if my changes match your explanation.

yoshinorin (approved):
@KagamigawaMeguri 
Sorry for the late reply. 
LGTM, I'll merge this. Thank you :)

D-Sketon:
Is this a breaking change?
This PR doesn't pass the current tests.

uiolee:
It looks like this PR did not run check.

and I tested in my PC.

```bash
  1146 passing (16s)
  5 pending
  1 failing

  1) Hexo
       Models
         PostAsset
           path - virtual - when permalink contains .htm not in the end:

      AssertionError: expected '2023\09\24\bar\bar\foo.html' to deeply equal '2023\09\24\bar\.htm-foo\foo.html'
      + expected - actual

      -2023\09\24\bar\bar\foo.html
      +2023\09\24\bar\.htm-foo\foo.html

      at Context.<anonymous> (D:\git\hexo\test\scripts\models\post_asset.js:95:22)
```

yoshinorin:
@D-Sketon @uiolee 
Sorry, I made some misunderstandings. I think we should revert this PR.

PR #5178, opened 2023-03-27
Refactor types

Pcrab:
## What does it do?

Add types to Hexo class and extend apis.

Only types are changed except for `lib/extend/tag.ts`, where I added an intermediate variable. All these changes should have no effect on runtime behavior.

This may help plugin authors who can now add hexo as devdependency and add `declare var hexo: import("hexo")` to `global.d.ts` to get better type inference.

## Screenshots

After adds

image

I can get better type inference like this:

image

which used to be just any type.

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/extend/syntax_highlight.ts
+++ b/lib/extend/syntax_highlight.ts
@@ -1,10 +1,32 @@
-interface Options {
-  context?: any;
-  args?: any;
+import Hexo from '../hexo';
+
+export interface HighlightOptions {
+  // plulgins/filter/before_post_render/backtick_code_block
+  lang: string,
+  caption: string,
+  lines_length?: number,
+  firstLineNumber?: string | number
+
+
+  language_attr?: boolean;
+  firstLine?: number;
+  line_number?: boolean;
+  line_threshold?: number;
+  mark?: number[];
+  wrap?: boolean;
+
+}
+
+interface HighlightExecArgs {
+  context?: Hexo;
+  args?: [
+    content: string,
+    options: HighlightOptions,
+  ];
 }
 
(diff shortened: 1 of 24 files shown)


lib/extend/syntax_highlight.ts
```diff
@@ -1,10 +1,32 @@
...
+  wrap?: boolean;
+
+}
+
+interface HighlightExecArgs {
+  context?: Hexo;
+  args?: [
+    content: string,
+    options: HighlightOptions,
+  ];
```
SukkaW, line 25: This is kinda incomplete.

lib/plugins/console/index.ts
```diff
@@ -1,4 +1,6 @@
-export = function(ctx) {
+import Hexo from '../../hexo';
```
SukkaW, line 1:
```suggestion
import type Hexo from '../../hexo';
```

Use `import type` for type only imports.

lib/extend/syntax_highlight.ts
Pcrab: Fixed in 4035282.

lib/plugins/console/index.ts
Pcrab: Fixed in 3ad806d.

Maybe we should add a rule in `.eslintrc.json` to enforce this style?

```json
{
  "rules": {
    "@typescript-eslint/consistent-type-imports": "warn",
  }
}
```

stevenjoezhang (approved):

yoshinorin (approved):
Thank you :)

PR #5202, opened 2023-05-11
chore: change dependencies version

Pcrab:
### npm and yarn

Since hexo-front-matter release v4.1.0, it uses typescript and requires exactly two parameters:

```typescript
declare function parse(str: string, options: yaml.LoadOptions): any;
```

However, in hexo, it calls the `parse` function mostly with only one parameter:
```typescript
import { parse as yfmParse } from 'hexo-front-matter';
Object.assign(data, yfmParse(content));
```

Although in `package.json` its version is defined as `^4.0.0`, npm and yarn will automatically install its latest version v4.1.0, which makes the build fail.

### pnpm

If we use `pnpm`, it will use `moize` v6.1.0, not its latest version v6.1.6, together with `micro-memoiz` v4.1.2.

`micro-memoize` used to contain its `src` folder but was removed in v4.1.0.

In `moize` its type definition contains:

```typescript
import { MicroMemoize } from 'micro-memoize/src/types';
```

which no longer works with the latest `micro-memoiz`. This is fixed in v6.1.6.

## Screenshots

with npm:

image

with pnpm:

image

## Pull request tasks

- [x] Passed the CI test.


Pcrab:
The former one should be fixed in https://github.com/hexojs/hexo-front-matter/pull/67, but since `hexo-front-matter` hasn't released a new version to include this change yet, we should still stick to v4.0.0 here to prevent type error.

tomap (approved):

yoshinorin (approved):
@Pcrab 
Thank you :)
We will release a new version of `hexo-front-matter`.

PR #5209, opened 2023-05-18
chore: improvement hexo.extend

dimaslanjaka:
## What does it do?
- 244be2ff fix(TS2345): Argument of type 'string | Buffer' is not assignable to parameter of type 'string'.
- a5aef69c fix: conditional of data is actually an `object` 
  - chore: add `Pattern` result `boolean` type 
  - ~~**NEED PR** hexojs/hexo-util/pull/311~~
    - ~~**changes** hexojs/hexo-util@9ddb7e2~~
  - ~~**OR** temporarily install my custom `hexo-util`.~~
    - ~~`npm i hexo-util@https://github.com/dimaslanjaka/hexo/raw/1d4d41d/releases/hexo-util.tgz`~~
    - ~~See all improved tarballs https://github.com/dimaslanjaka/hexo/tree/monorepo-v7/releases~~
- ~~b33d7e72 chore: moved interfaces to `renderer-d.ts`~~ already merged on `hexojs/master`
   using ES Export style to easily import between source files 
  - chore(Renderer.register): add overload method types 
  - docs: add JSDoc
- 1ab1d588 feat(Filter.register): add overload method types 
  - fix(lint): eslint --fix
- f2e17aa3 chore(StoreFunction): add explicit this marked as internal hexo function 
  - fix(lint): eslint --fix
- 31ef3141 chore: convert interface `TagFunction` to overload type 
  - chore: convert interface `AsyncTagFunction` to overload type 
  - docs: add some JSDoc 
  - chore(coverage): add prefix `_` to unused variables 
  - feat: add interface `RegisterAsyncOptions` 
  - fix: missing types feat(Tag.register): add overload types and JSDoc
  - sample callback with single parameter https://github.com/hexojs/hexo-tag-embed/blob/0e9467d55a2dd746e97fafe5e1d255fc490e5ea7/lib/tags/vimeo.ts#L9-L20
- 4a61ca54 fix: cast `pattern` as StoreFunction 
  - fix: conditional when `fn` is not function
- 6aee1f3b chore: add missing type param `ctx`
- 1a11ad5e chore: change old type `object` to `Record<string, any>
- cc9518b2 feat: add data type `before_post_render`
- chore(hexo.loadPlugin): make `callback` as optional & add JSDoc
   sample without callback https://github.com/hexojs/hexo/blob/8b95bbc722e5c77a7e8125441ed64d2ea3524ac0/test/scripts/tags/blockquote.js#LL8C16-L8C96
- feat: add context interface `ExtendedTagProperty`
  - chore: cast tag function callback as custom context
  useful for determining which post or page source the tag processor is currently processing from. see sample https://github.com/dimaslanjaka/hexo-shortcodes/blob/8224f992eee8a221db15996823e6d80e84a13ca1/src/include_file.ts#L55-L56

## Demo

- hexo.extend.tag hotfixed for typescript. see https://github.com/dimaslanjaka/hexo-shortcodes

## Pull request tasks

- [ ] Passed the CI test.


lib/extend/renderer.ts
```diff
@@ -80,10 +41,50 @@ class Renderer {
...
   }
 
+  /**
+   * register renderer engine
+   * - [hexo-renderer-nunjucks example](https://github.com/hexojs/hexo-renderer-nunjucks/blob/c71a1979535c47c3949ff6bf3a85691641841e12/lib/renderer.js#L55-L56)
+   * - [typescript example](https://github.com/dimaslanjaka/hexo-renderers/blob/feed801e90920bea8a5e7000b275b912e4ef6c43/src/renderer-sass.ts#L37-L38)
+   * @param name input extension name. ex: ejs
+   * @param output output extension name. ex: html
+   * @param fn renderer function
+   */
```
SukkaW, line 51: Is it possible to reduce/dedupe the comment here?

lib/plugins/filter/before_post_render/index.ts
```diff
@@ -1,4 +1,4 @@
-export = ctx => {
+export = (ctx: import('../../../hexo')) => {
```
SukkaW, line 1:
```suggestion
import type Hexo from '../../../hexo';

export = (ctx: Hexo) => {
```

lib/plugins/filter/template_locals/i18n.ts
```diff
@@ -13,7 +13,13 @@ function i18nLocalsFilter(locals) {
...
     const data = pattern.match(locals.path);
 
-    if (data && data.lang && i18nLanguages.includes(data.lang)) {
+    if (
+      typeof data !== 'undefined'
+      && !Array.isArray(data)
+      && typeof data !== 'boolean'
+      && data.lang
+      && i18nLanguages.includes(data.lang)
+    ) {
```
SukkaW, line 22: This is somewhat over-complicated for a simple type guard. Is it possible to simplify it?

lib/extend/renderer.ts
dimaslanjaka: > Is it possible to reduce/dedupe the comment here?

yes, or should we using shortlink from http://bit.ly to reduce comment length ?

lib/plugins/filter/template_locals/i18n.ts
dimaslanjaka: > This is somewhat over-complicated for a simple type guard. Is it possible to simplify it?

because there has 3 return type (`boolean | RegExpMatchArray | Record<string, unknown>`)
see https://github.com/hexojs/hexo-util/pull/311/commits/e0719fb7386ce5c528caf6811aea53df259a48a8

see boolean return conflict
-    https://github.com/dimaslanjaka/hexo/blob/4ba63bc65d05d231fd1db1053013e2ae71661a70/lib/plugins/processor/common.ts#L29-L32
-    https://github.com/dimaslanjaka/hexo/blob/4ba63bc65d05d231fd1db1053013e2ae71661a70/lib/theme/processors/source.ts#L25-L32

ive tried:
- when only using negate conditional (`!data`), ts throw error `property lang does not exist on Array type`
- when only validate `!Array.isArray(data)`, ts throw error `property lang does not exist on boolean type`

dimaslanjaka: if there is any other suggestion, maybe it can be simplified without breaking (throw) during `npm run build`

dimaslanjaka: or. enable explicit any, down level iterator, and moved all export as ES Export style. might it can used square bracket (`data['page']`) to suppress error validation during `npm run build`

dimaslanjaka: sample supress syntax https://github.com/dimaslanjaka/static-blog-generator/blob/8cf940f1f7141085c42da874417c3d1a0cecf280/packages/sbg-utility/src/utils/scheduler.ts#L127

but these syntax style unsupported by hexo eslint config, it should be automatically fixed (moved to dots property style - `data.page`) on commit.

PR #5259, opened 2023-07-29
fix: pass the language option for codeblock tag

xihale:
## What does it do?

I will pass the language option when I use either of `highlight.js` or `prism.js` for codeblock tag!

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/tag/post_link.ts
+++ b/lib/plugins/tag/post_link.ts
@@ -41,7 +41,13 @@ export = (ctx: Hexo) => {
     const attrTitle = escapeHTML(post.title || post.slug);
     if (escape === 'true') title = escapeHTML(title);
 
-    const url = new URL(post.path, ctx.config.url).pathname + (hash ? `#${hash}` : '');
+    // guarantee the base url ends with a slash. (case of using a subdirectory in the url of the site)
+    let baseUrl = ctx.config.url;
+    if (!baseUrl.endsWith('/')) {
+      baseUrl += '/';
+    }
+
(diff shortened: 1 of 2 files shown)


SukkaW (changes requested):
Would you mind adding unit test cases toward your changes?

xihale:
> Would you mind adding unit test cases toward your changes?

but, I don't know what unit test I should add, can you help me?

xihale:
> Would you mind adding unit test cases toward your changes?

but, I don't know what unit test I should add, can you help me?

xihale:
I added a unit test to test my changes and find a inconspicuous bug then fixed it.

lib/plugins/tag/code.js
```diff
@@ -118,9 +118,11 @@ function parseArgs(args) {
 
 module.exports = ctx => function codeTag(args, content) {
 
+  const options = parseArgs(args);
+
   // If neither highlight.js nor prism.js is enabled, return escaped code directly
   if (!ctx.extend.highlight.query(ctx.config.syntax_highlighter)) {
-    return `<pre><code>${escapeHTML(content)}</code></pre>`;
+    return `<pre><code${options.lang !== undefined ? ` class="language-${options.lang}"` : ''}>${escapeHTML(content)}</code></pre>`;
```
StephenNeelDurai, line 125: **Identical blocks of code found** on this line (125 and 140). I think you can create this line as separate variable and use it both places.

StephenNeelDurai (approved):

PR #5262, opened 2023-08-05
feat(helper/tagcloud): level false to use same className for each tag

ChrAlpha:
## What does it do?

When I need to add the same `class` for each tag in tagcloud, such as `whitespace-nowrap` in order to prevent text wrapping and `m-4` to margin between tags, the `level` option forcing a suffix feels a bit annoying.

As a consideration of that, I tried to add a new rule - when `level: false`, no suffix is added to the class. When `level` is set to `null`/`undefined`/`0`/blank string, it won't trigger into this rule. So this change won't break prior behaviour.

## Pull request tasks

- [ ] Passed the CI test.

### DIFF
--- a/lib/plugins/helper/tagcloud.js
+++ b/lib/plugins/helper/tagcloud.js
@@ -21,7 +21,7 @@ function tagcloudHelper(tags, options) {
   const className = options.class;
   const showCount = options.show_count;
   const countClassName = options.count_class || 'count';
-  const level = options.level || 10;
+  const level = options.level === false ? false : options.level || 10;
   const { transform } = options;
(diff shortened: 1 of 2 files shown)


renbaoshuo (changes requested):
Could you please add some tests about your changes?

ChrAlpha:
> Could you please add some tests about your changes?

I ran test script locally and it dit throw several failures. However, those failures remained the same before and after my commit.

stevenjoezhang (changes requested):
For your initial goal, you can use other CSS selectors, such as `.parent-element a`. Modifying the tag cloud's code is unnecessary.

Issue #4353, opened 2020-06-10
The proposal of using monorepo


SukkaW:
**Why should we use monorepo?**

For example, Hexo is dropping Node.js 8 at this moment. That means every CI config under every repo has to be updated. It is painful.

If we are using monorepo, we could write only one ci config for all packages (run eslint & all unit test cases at once).

Also, with monorepo we could maintain a global `eslintrc` / `mocharc` for all packages.

**What should be included in monorepos**

Only for those repos that are not hexo plugins, which means:

- repos like `hexo-util`, `hexo-i18n`, `hexo-pagination`, `hexo-server`, `hexo-cli`, etc. should be merged into monorepos.
- repos like `hexo-renderer-marked`, `hexo-filter-nofollow`, `hexo-uglify`, etc. should remain separated.

**When should we start**

Right after Hexo 5.0.0 released.

**What monorepo manager should we use?**

lerna.

**Any play ground?**

https://github.com/hexojs/hexo-monorepo

@hexojs/core and owners of @hexojs should have access to the repo.

tomap:
I've seen Lerna used at work, and I was not convinced.
But I thought about it in numerous cases, especially all the plumbing you are mentionning (nodejs versions, eslint, tests, readme, ...)

SukkaW:
@tommy351 has some exprience on [migrating to monorepo](https://zespia.tw/blog/2020/05/10/yarn-2-and-monorepo/). Hopefully he could give us some suggestions.

tommy351:
I think it will take some time to setup the environment for testing and releasing and the build time on CI will be longer. But it would be great to be able to share configs and utils across the project.

SukkaW:
[Next.js](https://github.com/vercel/next.js) uses lerna.
[React](https://github.com/facebook/react) uses yarn workspace. 
[Babel](https://github.com/babel/babel/) used to use lerna, and it is now using yarn workspace, too.
Microsoft has [Rush](https://github.com/microsoft/rushstack).

So which one should we use?

@tomap @yoshinorin @stevenjoezhang @jiangtj

ertrzyiks:
I wouldn't recommend using lerna nowadays, it's [maintenance is seems to fade out](https://github.com/lerna/lerna/issues/2703#issuecomment-744601134) and we can get a lot with yarn 3 workspaces alone.

yoshinorin:
> I wouldn't recommend using lerna nowadays, it's maintenance is seems to fade out

+1

[...]

PR #5273, opened 2023-08-20
migrate to monorepo

uiolee:
## What does it do?

migrate to monorepo. #4353

- use pnpm , turbo
- use monoreop struct
- update ci

## packages

- hexo
- hexo-util
- eslint-config-hexo

## known issue

- hexo/benchmark rely on dist package. we have to run `build` first.
- eslint will check `path`/`packages`' existence. we have to run `build` first or turn off some rules.
- pnpm drop node 14.

## Pull request tasks

- [x] Passed the CI test.


yoshinorin:
IMHO we would like to release the `v7.0.0` before migrating to a monorepo first. Because, it's not good to develop for `v7.0.0` in parallel with the master branch for long terms. There are various issue, such as contributers not aware of the `v7.0.0` branch and they submitting pull-request to the `master` branch. And we have to keep resolve many of conflicts between `master` and `v7.0.0`.  etc...

uiolee:
I agree. We don't seem to have enough people to maintain both branches. I would recommend going back to the main branch.

yoshinorin:
Personally, I want to migrate hexo to mono in `v8.0.0`.

uiolee:
Do you mean after the release of v8? Or before?

I think it's acceptable to migrate at any time because it doesn't affect the code or the users.

yoshinorin:
>  it doesn't affect the code or the users.

Oh, sorry I don't know that. If we no need to rename package etc... we can publish it before `v8.0.0`.

PR #5282, opened 2023-08-29
ci: reduce the running of ci

uiolee:
I guess it's not always necessary to run ci. 

1. linter and tester only need to be executed when the code or environment changes
2. dependabot causes ci to run twice
    dependabot

## What does it do?

1. add `paths` filter. only run ci when the code changes.
2. add `branches` filter. only run ci in specified branches.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -1,6 +1,19 @@
 name: Linter
 
-on: [push, pull_request]
+on:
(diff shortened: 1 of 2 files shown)


SukkaW (approved):

yoshinorin (approved):

uiolee:
Duplicate of #5291

PR #5291, opened 2023-09-12
ci: reduce the running of ci

uiolee:
Same as #5282

## What does it do?

1. add `paths` filter. only run ci when the code changes.
2. add `branches` filter. only run ci in specified branches.


SukkaW (approved):

D-Sketon:
Did #5153 cause the test to fail?

yoshinorin (approved):

Issue #4830, opened 2021-11-30
The Backtick Code Block does not support additional options


yunnysunny:
## Check List

Please check followings before submitting a new issue.

- [x] I have already read [Docs page](https://hexo.io/docs/) & [Troubleshooting page](https://hexo.io/docs/troubleshooting)
- [x] I have already searched existing issues and they are not help  to me
- [x] I examined error or warning messages and it's difficult to solve
- [x] Using [the latest](https://github.com/hexojs/hexo/releases) version of Hexo (run `hexo version` to check)
- [x] Node.js is higher than [minimum required version](https://hexo.io/docs/#Minimum-required-Node-js-version)

## Expected behavior
```text
  ```javascript line_number:true highlight:true mark:1,2,3
  function listenInCluster(server, address, port, addressType,
  backlog, fd, exclusive, flags) {
  exclusive = !!exclusive;
  
  if (cluster === undefined) cluster = require('cluster');
  
  if (cluster.isPrimary || exclusive) {
  // Will create a new handle
  // _listen2 sets up the listened handle, it is still named like this
  // to avoid breaking code that wraps this method
  server._listen2(address, port, addressType, backlog, fd, flags);
  return;
  }
  
  .....
  }
  ```
```

Github does not support show the markdown code , so I split the issue into too commits.

yunnysunny:
## Actual behavior
The code above will be rendered as follows:
![image](https://user-images.githubusercontent.com/734818/144011517-98c2f02f-ea3c-441a-8e3f-fde9f2802be3.png)

## How to reproduce?



## Is the problem still there under "Safe mode"?



## Environment & Settings

**Node.js & npm version(`node -v && npm -v`)**



```
v14.17.4
6.14.14
```

**Your site `_config.yml`** (Optional)



```yaml
# Hexo Configuration
## Docs: https://hexo.io/docs/configuration.html
## Source: https://github.com/hexojs/hexo/

# Site
title: Hexo
subtitle: ''
description: ''
keywords:
author: John Doe
language: en
timezone: ''

# URL
## Set your site url here. For example, if you use GitHub Page, set url as 'https://username.github.io/project'
url: http://example.com
permalink: :year/:month/:day/:title/
permalink_defaults:
pretty_urls:
  trailing_index: true # Set to false to remove trailing 'index.html' from permalinks
  trailing_html: true # Set to false to remove trailing '.html' from permalinks

# Directory
source_dir: source
public_dir: public
tag_dir: tags
[...]

PR #5299, opened 2023-09-21
Support additional options for Backtick Code Block

TimonPeng:
## What does it do?

Support additional options for Backtick Code Block.

## Screenshots

Before:

Before

After:

After

Source markdown:

Markdown

## Pull request tasks

- [ ] Passed the CI test.

Issue #4830 

Syntax of Backtick Code Block is:

```
[language] [title] [url] [link text] [additional options]
```

I was going to simply use split spaces to extract it, but the problem is... If do it with this way, then won't be able to use spaces in the title and link text, which is barely ok, but not what I want.

1. the first one [language] is sure there's no spaces
2. the second one [title] might have spaces
3. the third one [url] must be a link
4. the fourth one [link text] might have spaces
5. the fifth one [additional options] must have a colon and probably have spaces

It also would be chaos to matching with regex too, although I did write:

```javascript
/(\w+)\s+(.+?)\s+(?=http)([\S]+)\s+(.+?)\s+((?:\w+:\S+\s*)+)/

.exec('java title https://google.com google mark:1,4-7,10 wrap:true');

[
  'java title https://google.com google mark:1,4-7,10 wrap:true',
  'java',
  'title',
  'https://google.com',
  'google',
  'mark:1,4-7,10 wrap:true',
  index: 0,
  input: 'java title https://google.com google mark:1,4-7,10 wrap:true',
  groups: undefined
]

.exec('python Backtick Code Syntax https://flask.palletsprojects.com/en/2.3.x/ Flask Docs mark:1,5-7 wrap:true')

[
  'python Backtick Code Syntax https://flask.palletsprojects.com/en/2.3.x/ Flask Docs mark:1,5-7 wrap:true',
  'python',
  'Backtick Code Syntax',
  'https://flask.palletsprojects.com/en/2.3.x/',
  'Flask Docs',
  'mark:1,5-7 wrap:true',
  index: 0,
  input: 'python Backtick Code Syntax https://flask.palletsprojects.com/en/2.3.x/ Flask Docs mark:1,5-7 wrap:true',
  groups: undefined
]
```

In the end I decided to use the same code logic as the Code Block Plugin(lib/plugins/tag/code.js), which is simpler and easier to synchronise updates and maintain.

### DIFF
--- a/lib/plugins/filter/before_post_render/backtick_code_block.js
+++ b/lib/plugins/filter/before_post_render/backtick_code_block.js
@@ -48,11 +48,11 @@ module.exports = ctx => {
           case 'line_threshold':
             if (!isNaN(value)) line_threshold = +value;
             break;
+          case 'language_attr':
+            language_attr = value === 'true';
+            break;
           case 'first_line':
             if (!isNaN(value)) firstLine = +value;
-            break;
-          case 'wrap':
-            wrap = value === 'true';
             break;
           case 'mark': {
             for (const cur of value.split(',')) {
@@ -75,10 +75,9 @@ module.exports = ctx => {
             }
             break;
           }
-          case 'language_attr': {
-            language_attr = value === 'true';
+          case 'wrap':
+            wrap = value === 'true';
             break;
-          }
           default: {
             _else.push(args[i]);
           }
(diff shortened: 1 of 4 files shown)


D-Sketon:
can you add some tests?

TimonPeng:
> can you add some tests?

Added, it's looks good to you?

Issue #5260, opened 2023-08-01
duplicates of `site.data` in db.json leads to large size


EmptyDreams:
## Check List

Please check followings before submitting a new issue.

- [x] I have already read [Docs page](https://hexo.io/docs/) & [Troubleshooting page](https://hexo.io/docs/troubleshooting)
- [x] I have already searched existing issues and they are not help  to me
- [x] I examined error or warning messages and it's difficult to solve
- [x] Using [the latest](https://github.com/hexojs/hexo/releases) version of Hexo (run `hexo version` to check)
- [x] Node.js is higher than [minimum required version](https://hexo.io/docs/#Minimum-required-Node-js-version)

## Expected behavior

The db.json file should be a reasonable size. (Should it be under 50MB at the very least?)

## Actual behavior

But in fact, even though I deleted most of the posts from my blog and kept only one post (the md file was 5KB in size), the resulting db.json was still over 70MB.

I'm not quite clear on why this happened, I didn't have this problem originally either, it seemed to come out of nowhere but I can't remember what I was modifying at the time.

Through my observation, there are a lot of duplicate `_id & data` fields in db.json, such as the `source/_data/avatar/cpen.webp` file, which is recorded more than 100 times in db.json, which takes up a lot of space.

This problem caused my `hexo se` and `hexo g` to wait quite a long time after the `INFO Validating config` step, and I had to run `hexo cl` or manually delete the db.json before each command, which clearly defeats the purpose of the db.json file.

## Is the problem still there under "Safe mode"?

The problem persists even if I add the `--safe` directive.

## Environment & Settings

**Node.js & npm version(`node -v && npm -v`)**

node version: 18.12.1

npm version: 9.8.1

**Your site `_config.yml`** (Optional)

**Hexo and Plugin version(`npm ls --depth 0`)**

<details><summary>Details</summary>
<p>

```
+-- @neilsustc/markdown-it-katex@1.0.0
+-- cheerio@1.0.0-rc.12
+-- gulp-clean@0.4.0
+-- gulp-cssnano@2.1.3
+-- gulp-html-minifier-terser@7.1.0
+-- gulp-htmlclean@2.7.22
+-- gulp-terser@2.1.0
+-- gulp@4.0.2
+-- hexo-abbrlink@2.2.1
+-- hexo-asset-image@1.0.0
+-- hexo-butterfly-envelope@1.0.15
+-- hexo-deployer-git@3.0.0
+-- hexo-filter-nofollow@2.0.2
[...]

PR #5325, opened 2023-10-22
revert: Access data files from source folder (#1969)

stevenjoezhang:
## What does it do?

reverts #1969
fixes https://github.com/hexojs/hexo/issues/5260

Motivation for the modification: I suspect that this feature is rarely used (I noticed it for the first time today). In fact, users can access the 'site' variable through the template engine, without the need for tags within the article.

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/hexo/index.ts
+++ b/lib/hexo/index.ts
@@ -555,7 +555,7 @@ class Hexo extends EventEmitter {
     this.emit('generateBefore');
 
     // Run before_generate filters
-    return this.execFilter('before_generate', this.locals.get('data'), { context: this })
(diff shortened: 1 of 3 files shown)


uiolee (approved):

D-Sketon:
breaking change？
maybe adding a parameter to control this behavior is a better choice?

stevenjoezhang:
Yes, it is a breaking change, we could include it in the v7.0.0 release. However, I don't think it's necessary to add a new option. This usage is not mentioned in Hexo's documentation; it's a completely undocumented interface. We don't need to consider its stability and compatibility.

PR #5344, opened 2023-11-05
refactor: refactor types

D-Sketon:
## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/extend/renderer.ts
+++ b/lib/extend/renderer.ts
@@ -14,7 +14,7 @@ export interface StoreFunctionData {
   text?: string;
   engine?: string;
   toString?: any;
-  onRenderEnd?: any;
+  onRenderEnd?: (...args: any[]) => any;
 }
 
 export interface StoreSyncFunction {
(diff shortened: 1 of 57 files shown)


D-Sketon:
should we change test files to ts ?

D-Sketon:
found some type errors when changing type files to ts

uiolee:
> should we change test files to ts ?

I guess that's acceptable. That way 100% ts can be achieved. But I believe this should be done in another PR

图片

uiolee (dismissed):

lib/extend/renderer.ts
```diff
@@ -57,7 +59,7 @@ class Renderer {
     this.storeSync = {};
   }
 
-  list(sync: boolean): Store | SyncStore {
+  list(sync?: boolean): Store | SyncStore {
```
SukkaW, line 62:
```suggestion
  list(sync = false): Store | SyncStore {
```

Let's use default boolean value here.

SukkaW (dismissed):

SukkaW (approved):

PR #5385, opened 2023-12-22
fix(tag): use url_for

stevenjoezhang:
## What does it do?

Issue resolved: https://github.com/hexojs/hexo/issues/5383

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/models/post_asset.ts
+++ b/lib/models/post_asset.ts
@@ -1,5 +1,5 @@
 import warehouse from 'warehouse';
-import { join } from 'path';
+import { join, posix } from 'path';
 import type Hexo from '../hexo';
 
 export = (ctx: Hexo) => {
@@ -19,7 +19,8 @@ export = (ctx: Hexo) => {
     // PostAsset.path is file path relative to `public_dir`
     // no need to urlescape, #1562
     // strip /\.html?$/ extensions on permalink, #2134
-    return join(post.path.replace(/\.html?$/, ''), this.slug);
+    // Replace backslashes on Windows using posix.join
+    return posix.join(post.path.replace(/\.html?$/, ''), this.slug);
   });
 
   PostAsset.virtual('source').get(function() {
--- a/lib/models/post_asset.ts
+++ b/lib/models/post_asset.ts
@@ -19,7 +19,7 @@ export = (ctx: Hexo) => {
     // PostAsset.path is file path relative to `public_dir`
     // no need to urlescape, #1562
     // strip /\.html?$/ extensions on permalink, #2134
-    // Replace backslashes on Windows using posix.join
+    // Use path.posix.join to avoid path.join introducing unwanted backslashes on Windows.
     return posix.join(post.path.replace(/\.html?$/, ''), this.slug);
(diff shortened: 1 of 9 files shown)


lib/plugins/tag/post_link.ts
```diff
@@ -41,7 +41,7 @@ export = (ctx: Hexo) => {
     const attrTitle = escapeHTML(post.title || post.slug);
     if (escape === 'true') title = escapeHTML(title);
 
-    const url = new URL(post.path, ctx.config.url).pathname + (hash ? `#${hash}` : '');
+    const url = url_for.call(ctx, post.path) + (hash ? `#${hash}` : '');
```
stevenjoezhang, line 44: Potential double encode here

hx-yinpengfei:
Help, when will the package be updated?

uiolee:
fail to pass test on Windows because of backslash (`%5c`)

log

try:
https://github.com/hexojs/hexo/blob/f5e0f92eda9491116c585142c565efcf7bf7a3e3/lib/theme/index.ts#L53

stevenjoezhang:
I'm sorry, I made some mistakes in operation while rebasing and force pushing, and the unfinished version was pushed to master. I will use this pull request to fix the unit tests that are currently failing.

uiolee:
It doesn't look like it. activity indicates that there are no changes in the master branch

stevenjoezhang:
I think we can introduce a new convention: In the database models used by Hexo (including Asset, Page, Post, etc.), the `source` attribute is used to specify the file system path under the source directory, which includes backslashes (`\`) on Windows; whereas the `path` attribute is used to specify their paths on the web, where backslashes should be replaced by `/`.

stevenjoezhang:
Database model `Asset` also contains an `_id` attribute (instead of using the UUID automatically assigned by warehouse), which by default is the file's relative path. I have found that when using it, backslashes are replaced, which seems unnecessary. I will try to continue with the refactoring.

stevenjoezhang:
Another issue is that when calculating relative paths, sometimes `substring` is used and sometimes `relative` is used. This inconsistency also needs to be addressed and fixed in the future.

lib/models/post_asset.ts
```diff
@@ -19,7 +19,8 @@ export = (ctx: Hexo) => {
     // PostAsset.path is file path relative to `public_dir`
     // no need to urlescape, #1562
     // strip /\.html?$/ extensions on permalink, #2134
-    return join(post.path.replace(/\.html?$/, ''), this.slug);
+    // Replace backslashes on Windows using posix.join
```
uiolee, line 22: posix.join will not replace backslash.

```node
path.posix.join('asd\\zxc','qwe')
```

stevenjoezhang: This comment should be changed to:
```
Use `path.posix.join` to avoid `path.join` introducing unwanted backslashes on Windows.
```
The input parameters here actually do not include backslashes, so the code is fine.

uiolee (approved):

PR #5388, opened 2023-12-28
revert(categories,tags): revert behavior of locals.tags and locals.categories

uiolee:
https://github.com/hexojs/hexo/issues/5380#issuecomment-1871282281

## What does it do?

revert some changes of #5119

fix #5380
fix #5367

## Pull request tasks

- [x] Add test cases for the changes.
- [x] Passed the CI test.

PR #5398, opened 2024-01-19
refactor: refactor types

D-Sketon:
## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -131,6 +131,15 @@ export interface TagSchema {
   _id?: string | number;
   name?: string;
 }
+
+export interface AssetSchema {
+  _id?: string;
+  path: string;
+  modified: boolean;
+  renderable: boolean;
+  source: string;
+}
+
 export interface LocalsType {
   page: PostSchema | PageSchema;
   path: string;
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -38,7 +38,7 @@ export interface RenderData {
   engine?: string;
   content?: string;
   disableNunjucks?: boolean;
-  markdown?: object;
+  markdown?: any;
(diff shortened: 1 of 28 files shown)


lib/types.ts
```diff
@@ -121,9 +233,9 @@ export type NormalPageGenerator = {
...
 
 export interface SiteLocals {
-  posts: any; // _Query
-  pages: any; // _Query
-  categories: any; // _Model
-  tags: any; // _Model
+  posts: Query<PostSchema>; // _Query
+  pages: Query<PageSchema>; // _Query
+  categories: Model<CategorySchema>; // _Model
+  tags: Model<TagSchema>; // _Model
```
D-Sketon, line 239: after #5388, Model should be Query again

yoshinorin (commented):
~~Shouldn’t we update the warehouse package's version from `^5.0.0` to `^5.0.1`?~~

~~Because this PR depends on `warehouse@5.0.1`. Of course we are using caret for specify version, it's means no worry with installation. But, IMHO we should explicitly specify package versions to clarify dependencies.~~

-----------------------

P.S:

Sorry, I didn't notice that...

https://github.com/hexojs/hexo/pull/5398/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519

SukkaW (dismissed):

uiolee (approved):

D-Sketon:
fix #5403

SukkaW (commented):
`any` would work for now. But IMHO we could use interface merging + theme overloading in the future.

SukkaW (approved):

PR #5410, opened 2024-01-28
docs(README): Update Sponsors images

uiolee:
## What does it do?

Remove invalid picture

Add the image of `Sponsors` from `opencollective`

### before

before

### after

after

## Pull request tasks

- [ ] Passed the CI test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ Also, we welcome PR or issue to [official-plugins](https://github.com/hexojs).
 
 ## Sponsors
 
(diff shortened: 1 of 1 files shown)


yoshinorin:
I can't determine whether to remove RSS3 from the sponsors because I don't know if they are still providing support or not.

uiolee:
From `opencollective` it looks like `rss3` is no longer active. 

I'm not sure if there are other support channels

yoshinorin (approved):
> From opencollective it looks like rss3 is no longer active.

Indeed. I have confirmed their transactions. We can merge this PR.

https://opencollective.com/rss3_/transactions

PR #5417, opened 2024-02-07
refactor: migrate typescript

D-Sketon:
## Pull request tasks

- [x] Passed the CI test.


lib/hexo/render.ts
```diff
@@ -56,7 +56,7 @@ class Render {
     return this.getRenderer(ext, true);
   }
 
-  render(data: StoreFunctionData, options?: { highlight?: boolean; }, callback?: NodeJSLikeCallback<any>): Promise<any> {
+  render(data: StoreFunctionData, options?: any | NodeJSLikeCallback<any>, callback?: NodeJSLikeCallback<any>): Promise<any> {
```
uiolee, line 59: How about using overloads, is it better?

tsconfig.json
```diff
@@ -7,7 +7,8 @@
     "declaration": true,
     "esModuleInterop": true,
     "types": [
-      "node"
+      "node",
+      "mocha"
```
uiolee, line 11: It looks like it will load `@types/mocha` everywhere including `src`, I don't know if this is correct and acceptable

uiolee (dismissed):
looks great to me

.mocharc.yml
```diff
@@ -3,4 +3,3 @@ reporter: spec
 ui: bdd
 full-trace: true
 exit: true
-parallel: true
```
SukkaW, line 6: Any reason why `parallel` was disabled?

D-Sketon: I've tried keeping parallel, but then tests containing async on my computer throw exceptions, and I don't know why
image
image

uiolee: parallel makes it easier to make tests timeout in my experience

SukkaW (approved):

PR #5457, opened 2024-04-08
refactor: backslashes on Windows

stevenjoezhang:
## What does it do?

I have noticed that currently on Windows, paths in Hexo are converted frequently. For example, when adding files to the `ctx.model('Cache')` database, the path separators `\` are replaced with `/`, and later when Assets use these paths, the `join` function is called, which on Windows, will replace `/` back with `\`. This pull request aims to standardize the separator in the cache's `_id` to match the current operating system's support and to avoid using hardcoded separators (using `path.sep` instead). It will not change existing functionalities but is intended to improve the uniformity and maintainability of the code.

~~WIP: pending https://github.com/hexojs/hexo/pull/5385~~

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/box/index.ts
+++ b/lib/box/index.ts
@@ -113,7 +113,7 @@ class Box extends EventEmitter {
     const src = join(this.base, path);
 
     return Cache.compareFile(
-      escapeBackslash(src.substring(ctx.base_dir.length)),
+      src.substring(ctx.base_dir.length),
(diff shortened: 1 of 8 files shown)


SukkaW (approved):

stevenjoezhang:
See also https://github.com/hexojs/hexo-util/pull/230

PR #5470, opened 2024-04-18
feat: add option to use slug as title of post

### LINKED ISSUE #5372: option to use filenames or a title tag in files as the title for a post.
### Check List

- [X] I have already read [Docs page](https://hexo.io/docs/).
- [X] I have already searched existing issues.

### Feature Request

I'd like to be able to use the filename of a post as the title or use a tag inside the file to indicate the title it should have, instead of **(no title)** 

To me this seems like a pretty basic feature, but I haven't been able to find anything like it.

### Others

No. I do not want to always manage a hexo instance locally just to add new posts. I just want to push an md file and have the title be picked up automatically, like described above.

If a page title could be updated this way, that would be nice but I don't consider it a requirement.

uiolee:
## What does it do?

use `slug` as `title` of post when `title` is not specified.

closed #5372

## Pull request tasks

- [x] Passed the CI test.
- [ ] add to [configuration](https://hexo.io/docs/configuration.html#Writing)

### DIFF
--- a/lib/plugins/processor/post.ts
+++ b/lib/plugins/processor/post.ts
@@ -70,8 +70,8 @@ function processPost(ctx: Hexo, file: _File) {
   const { path } = file.params;
   const doc = Post.findOne({source: file.path});
   const { config } = ctx;
-  const { timezone: timezoneCfg } = config;
-  const updated_option = config.updated_option;
+  const { timezone: timezoneCfg, updated_option, use_filename_as_post_title } = config;
+
   let categories, tags;
 
   if (file.type === 'skip' && doc) {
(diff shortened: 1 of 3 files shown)


stevenjoezhang:
See also https://github.com/hexojs/hexo/pull/4344
Adding a new option is a good solution, especially if the theme does not support empty titles

lib/plugins/processor/post.ts
```diff
@@ -109,6 +109,12 @@ function processPost(ctx: Hexo, file: _File) {
       if (!preservedKeys[key]) data[key] = info[key];
     }
 
+    // use `slug` as `title` of post when `title` is not specified.
+    // https://github.com/hexojs/hexo/issues/5372
+    if (use_filename_as_post_title && !Object.keys(data).includes('title')) {
```
SukkaW, line 114:
```suggestion
    if (use_filename_as_post_title && !('title' in data)) {
```

SukkaW (approved):

PR #5476, opened 2024-04-21
WIP: Allow `publish` handle files with filenames separated by spaces

yapinxxx:
Allow the publish command to handle files with filenames separated by spaces in the _draft folder.

## What does it do?

The update enables filenames to include spaces in the `_draft` folder. 
However, the source code modifies the original filename(`data.slug`) and then incorrectly compares it with the filenames in the `_draft` folder, leading to erroneous results.

### DIFF
--- a/lib/hexo/post.ts
+++ b/lib/hexo/post.ts
@@ -348,8 +348,7 @@ class Post {
     const { config } = ctx;
     const draftDir = join(ctx.source_dir, '_drafts');
     const slug = slugize(data.slug.toString(), { transform: config.filename_case });
-    data.slug = slug;
-    const regex = new RegExp(`^${escapeRegExp(slug)}(?:[^\\/\\\\]+)`);
+    const regex = new RegExp(`^${escapeRegExp(data.slug.toString())}(?:[^\\/\\\\]+)`);
     let src = '';
     const result: Result = {};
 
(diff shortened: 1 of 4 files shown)


SukkaW (changes requested):
Would you like to create a test case against your changes?

yapinxxx:
> Would you like to create a test case against your changes?

If you could assist me in creating a test case, I would greatly appreciate it.

---

Before submitting my change, I experimented with modifications in the node_modules directory to test my idea. 

Once I confirmed the concept could work, I directly edited the TypeScript source code. 

However, I faced challenges building the project and comparing changes before and after, so I submitted the pull request without a comparison. 

Additionally, creating a test case proved difficult due to my limited expertise in front-end development. 
I struggled with building the project and providing a comparison, making it hard to create the test case on my own.

yapinxxx:
> Would you like to create a test case against your changes?

Hello @SukkaW ,
Could you please provide me with some guidance on creating tests for my changes?
I've already tried using `npm test`, but I'm not certain if that's what you're referring to.

SukkaW:
> > Would you like to create a test case against your changes?
> 
> Hello @SukkaW , Could you please provide me with some guidance on creating tests for my changes? I've already tried using `npm test`, but I'm not certain if that's what you're referring to.

`npm test` runs tests against the existing test cases to make sure Hexo's existing behaviors do not change (no regression is introduced in your changes).

You are introducing a new behavior (correct handling of the special filename) that doesn't exist in the current version of Hexo.

You will need to add a test case to make sure that Hexo will always handle the special filename properly. So in the future when we run `npm test` again, we can make sure Hexo will always handle the special filename correctly.

stevenjoezhang:
Superseded by #5805  Thanks for your contribution!

PR #5482, opened 2024-04-28
WIP: perf(router): avoid calling `set` when the generated data unmodified

yoshinorin:
## What does it do?

TODO

## Screenshots

N/A

## Pull request tasks

- [ ] Passed the CI test.

### DIFF
--- a/lib/hexo/index.ts
+++ b/lib/hexo/index.ts
@@ -626,7 +626,10 @@ class Hexo extends EventEmitter {
       const { data, layout } = generatorResult;
 
       if (!layout) {
-        route.set(path, data);
+        // @ts-ignore
+        if (data.modified == null || data.modified) {
+          route.set(path, data);
+        }
         return path;
       }
 


yoshinorin:
The test cases for the watcher are failing.

test/scripts/hexo/hexo.ts
```diff
@@ -218,26 +218,26 @@ describe('Hexo', () => {
...
     await writeFile(target, newBody); // Update the file
     await Promise.delay(300);
-    await checkStream(route.get('test.txt'), newBody); // Check the new route
+    await checkStream(route.get(fileName), newBody); // Check the new route
     hexo.unwatch(); // Stop watching
     await unlink(target); // Delete the file
   }
 
-  it('watch() - source', async () => await testWatch(hexo.source_dir));
+  it('watch() - source', async () => await testWatch(hexo.source_dir, 'source.test.watch.txt'));
```
yoshinorin, line 238: I changed it to use unique files for each test for watch. Because using the same files (`test.txt`) as other tests caused failures. Perhaps caused by caching or something similar, but I'm not sure.

yoshinorin:
I've noticed that when `db.json` exists before `hexo s` execution, CSS files and others aren't registered to the route. This is likely related to the tests failing when the file `test.txt` is used. I'll investigate.

yoshinorin:
> I've noticed that when db.json exists before hexo s execution, CSS files and others aren't registered to the route in this PR. This is likely related to the tests failing when the file test.txt is used. I'll investigate.

It seems that the `_routerRefresh` method is called not only for updating the `router` but also for initializing the `router`. Therefore, it seems necessary to make changes such as separating the method for initialization and updating.

lib/hexo/index.ts
```diff
@@ -626,7 +626,11 @@ class Hexo extends EventEmitter {
       const { data, layout } = generatorResult;
 
       if (!layout) {
-        route.set(path, data);
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
```
SukkaW, line 630: Always prefer `// @ts-expect-error`, do not use `// @ts-ignore`

SukkaW (changes requested):

PR #5487, opened 2024-05-05
feat(helper/toc): specify maximum number of items to output

KentarouTakeda:
## What does it do?

Added the ability to set an upper limit on the number of items in the table of contents output by the `toc()` helper.

If no value is specified, the default is `Infinity`. This is no different from previous behavior. If we specify a number greater than or equal to 1, the elements will be truncated in the following order.

* Delete items from descendants to ancestors (from `<h6>` to `<h1>`) to fit within the specified number.
* Subheadings, which are lower-level headings on the page, will be removed in groups by level.
* The behavior differs only for the highest heading that appears on the page, which is deleted sequentially from the back.

For example, if we have a heading like this:

> \## 1
> \### 1-1
> \### 1-2
> \## 2
> \### 2-1

If `5` is specified for `max_items`, all items will be output. On the other hand, if we specify `4` or `3`,

> **\## 1**
> ~~\### 1-1~~ *truncated*
> ~~\### 1-2~~ *truncated*
> **\## 2**
> ~~\### 2-1~~ *truncated*

All `<h3>` are removed, resulting in two items being output. This is the same result as specifying `2` for `max_items`.
This is because this outline requires five elements to display up to `<h3>`.

Only the highest heading that appears on the page behaves differently. For example, if we specify `1` for `max_items`,

> **\## 1**
> ~~\### 1-1~~ *truncated*
> ~~\### 1-2~~ *truncated*
> ~~\## 2~~ *truncated*
> ~~\### 2-1~~ *truncated*

Instead of all `<h2>` being deleted at once as before, they will be deleted sequentially from the bottom. This is to avoid unintentionally not outputting anything.

In addition to the code, I welcome any feedback regarding the specifications described here. I believe this specification is the best, but if you have any other good ideas, I would definitely consider them.

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/helper/toc.ts
+++ b/lib/plugins/helper/toc.ts
@@ -110,7 +110,7 @@ function getAndTruncateTocObj(str: string, options: {min_depth: number, max_dept
   if (data.length === 0) {
     return data;
   }
-  if (max_items < 1) {
+  if (max_items < 1 || max_items === Infinity) {
     return data;
   }
 
--- a/lib/plugins/helper/toc.ts
+++ b/lib/plugins/helper/toc.ts
@@ -114,8 +114,9 @@ function getAndTruncateTocObj(str: string, options: {min_depth: number, max_dept
     return data;
   }
 
-  const min = Math.min(...data.map(item => item.level));
-  const max = Math.max(...data.map(item => item.level));
+  const levels = data.map(item => item.level);
+  const min = Math.min(...levels);
+  const max = Math.max(...levels);
 
   for (let currentLevel = max; data.length > max_items && currentLevel > min; currentLevel--) {
     data = data.filter(item => item.level < currentLevel);
--- a/lib/plugins/helper/toc.ts
+++ b/lib/plugins/helper/toc.ts
@@ -3,6 +3,7 @@ import { tocObj, escapeHTML, encodeURL } from 'hexo-util';
 interface Options {
   min_depth?: number;
   max_depth?: number;
+  max_items?: number;
   class?: string;
   class_item?: string;
   class_link?: string;
@@ -17,6 +18,7 @@ function tocHelper(str: string, options: Options = {}) {
   options = Object.assign({
     min_depth: 1,
     max_depth: 6,
+    max_items: Infinity,
     class: 'toc',
     class_item: '',
     class_link: '',
@@ -27,7 +29,7 @@ function tocHelper(str: string, options: Options = {}) {
     list_number: true
   }, options);
 
-  const data = tocObj(str, { min_depth: options.min_depth, max_depth: options.max_depth });
+  const data = getAndTruncateTocObj(str, { min_depth: options.min_depth, max_depth: options.max_depth }, options.max_items);
 
   if (!data.length) return '';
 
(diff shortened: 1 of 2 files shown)


KentarouTakeda:
It seems that the test is failing at some points unrelated to this pull request. Other pull requests have also failed, so I'll leave it as is.

lib/plugins/helper/toc.ts
```diff
@@ -27,7 +29,7 @@ function tocHelper(str: string, options: Options = {}) {
     list_number: true
   }, options);
 
-  const data = tocObj(str, { min_depth: options.min_depth, max_depth: options.max_depth });
+  const data = getAndTruncateTocObj(str, { min_depth: options.min_depth, max_depth: options.max_depth }, options.max_items);
```
uiolee, line 32: If the user does not need `max_items`, then there is no need to perform this operation everytime.

lib/plugins/helper/toc.ts
```diff
@@ -102,4 +104,26 @@ function tocHelper(str: string, options: Options = {}) {
...
+
+  if (data.length === 0) {
+    return data;
+  }
+  if (max_items < 1) {
+    return data;
+  }
+
+  const min = Math.min(...data.map(item => item.level));
+  const max = Math.max(...data.map(item => item.level));
```
uiolee, line 118: `...data.map(item => item.level)` They look the same, 
I guess it doesn't need to be executed twice

uiolee (changes requested):

lib/plugins/helper/toc.ts
KentarouTakeda: fixed: c17db64

I thought it was necessary to avoid `tocObj()` appearing multiple times in different places, so I wrote the skip judgment at the beginning of `getAndTruncateTocObj()` instead of where you gave the example.

lib/plugins/helper/toc.ts
KentarouTakeda: fixed: f8d04a6

lib/plugins/helper/toc.ts
```diff
@@ -102,4 +104,27 @@ function tocHelper(str: string, options: Options = {}) {
...
+    return data;
+  }
+
+  const levels = data.map(item => item.level);
+  const min = Math.min(...levels);
+  const max = Math.max(...levels);
+
+  for (let currentLevel = max; data.length > max_items && currentLevel > min; currentLevel--) {
+    data = data.filter(item => item.level < currentLevel);
+  }
```
SukkaW, line 123: This might be inefficient (iterating data over and over again). But I don't have any better idea for now.

lib/plugins/helper/toc.ts
```diff
@@ -102,4 +104,27 @@ function tocHelper(str: string, options: Options = {}) {
...
+
+  data = data.slice(0, max_items);
```
SukkaW, line 125: This is not necessary.

yoshinorin: It seems necessary because if we delete this code, the tests for [`max_items - result of h1 was truncated`](https://github.com/hexojs/hexo/pull/5487/files#diff-2625d80ed02e147c8693d6cb4c52d0477bd1e299518df8097b574ef14e6987f9R666) will fail.

yoshinorin (approved):
@uiolee @SukkaW 
Do you have any suggestions? If possible, I would like to release this in version `7.3.0`. #5506

SukkaW (approved):
The implementation is inefficient. But since the benchmark doesn't show any regression, I am good with that.

D-Sketon:
Site needs to be updated

KentarouTakeda:
Of course!
Please wait just a moment.

Issue #5215, opened 2023-05-26
Use nativebird instead of bluebird


so1ve:
## Check List

Please check followings before submitting a new feature request.

- [x] I have already read [Docs page](https://hexo.io/docs/)
- [x] I have already searched existing issues

## Feature Request

Hexo v7 still dependents on bluebird, which seems to be an overkill. Most of its features have been implemented in native promise now, so we can switch to [nativebird](https://github.com/doodlewind/nativebird) for better performance.

## Others

If this proposal is accepted I'm willing to implement

stevenjoezhang:
See https://github.com/hexojs/hexo/issues/3328

so1ve:
Yeah I have seen that before, but I'm not sure whether hexo uses some non-builtin promise methods. Feel free to close.

SukkaW:
> I've seen a lot of Bluebird-specific APIs used in the Hexo source code 🤔 WDYT?

Bluebird and Hexo were born when ECMAScript Promise didn't exist/API was not stable.

And since some Hexo plugins may still rely on callback-based APIs, using Bluebird's `.asCallback` method can help maintain backward compatibility.

so1ve:
Makes sense!

PR #5511, opened 2024-06-30
cleanup(wip): remove bluebird

talentlessguy:
## What does it do?

Removes Bluebird and replaces it with native Promise API. Promises have been in Node since [0.12](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#browser_compatibility) so there's no need to use it anymore.

Needs a PR to https://github.com/hexojs/warehouse as well

## Pull request tasks

- [ ] Passed the CI test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ coverage/
 .tmp*
 .vscode
 dist/
+bun.lockb
--- a/lib/box/file.ts
+++ b/lib/box/file.ts
@@ -1,4 +1,3 @@
-import type Promise from 'bluebird';
 import { readFile, readFileSync, stat, statSync, type ReadFileOptions } from 'hexo-fs';
 import type fs from 'fs';
 
--- a/lib/box/index.ts
+++ b/lib/box/index.ts
@@ -1,5 +1,4 @@
 import { join, sep } from 'path';
-import BlueBirdPromise from 'bluebird';
 import File from './file';
 import { Pattern, createSha1Hash } from 'hexo-util';
 import { createReadStream, readdir, stat, watch } from 'hexo-fs';
@@ -99,7 +98,7 @@ class Box extends EventEmitter {
     });
   }
 
-  _readDir(base: string, prefix = ''): BlueBirdPromise<any> {
+  _readDir(base: string, prefix = ''): Promise<any> {
     const { context: ctx } = this;
     const results = [];
     return readDirWalker(ctx, base, results, this.ignore, prefix)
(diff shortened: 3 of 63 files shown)


SukkaW:
https://github.com/hexojs/hexo/issues/5215#issuecomment-1564116657

talentlessguy:
Ah crap, didn't see this. I'll switch to nativebird instead.

SukkaW:
> Ah crap, didn't see this. I'll switch to nativebird instead. Or close the PR?

You can give `nativebird` a shot, and we can compare the performance. If there is no performance regression, we can continue!

talentlessguy:
@SukkaW oh awesome, I'll revert my changes then and use nativebird

talentlessguy:
Temporarily blocked by https://github.com/doodlewind/nativebird/pull/13

And a few other methods which implementations of I'll add to nativebird later

benjamingr:
Just refactor the bb specific stuff instead of wrapping it? You don't need Promise.asCallback when you have util.callbackify built in, Promise.promisify with util.promisify etc

benjamingr:
@SukkaW IMO just rip the bandaid, bluebird isn't getting major updates. We worked hard to make sure there are alternatives for every API we had in bluebird (I'm a maintainer) in Node.js (I'm.. also a maintainer).

Most stuff should just be refactored to native (e.g. Promise.method to async functions), other stuff to platform alternatives e.g. AbortSignal/AbortController for cancellation, util.promisify for converting callback functions, streams for map+concurrency etc.

The amount of stuff actually missing is very small (stuff like typed catch) and can be relatively easily worked around.


----

Also yes, I realize bluebird has no dependencies, we didn't want risk and to slow things down in the library - the library also has internal build tooling to reduce its size and improve its speed by not having a lot of runtime checks :]

SukkaW:
> IMO just rip the bandaid, bluebird isn't getting major updates. We worked hard to make sure there are alternatives for every API we had in bluebird (I'm a maintainer) in Node.js (I'm.. also a maintainer).

What I am really curious about is the bluebird performance. We've tried to replace bluebird 4 years ago: https://github.com/hexojs/hexo/pull/3939, https://github.com/hexojs/hexo/pull/4019. Our internal benchmark shows that there was about 175% performance regression at the moment.

To this day, bluebird still has comparable performance with the native promise while consuming 50% of the memory:

image

The benchmark suite can be found here: https://github.com/SukkaW/promise-performance-tests (originated from v8's promise benchmark suite).

talentlessguy:
I guess this can be closed since Bluebird won't be removed due to perf reasons

PR #5520, opened 2024-07-18
Update the centering layout of the previous section in the README file

lmliheng:
## Before
ada8bb0245c610cd4265baa34edc83c

## After
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,19 +7,19 @@
     <a href="https://www.npmjs.com/package/hexo">
       <img alt="NPM version" src="https://badge.fury.io/js/hexo.svg" />
     </a>
-    <a href="https://github.com/hexojs/hexo/actions?query=workflow%3ATester">
+    <a href="">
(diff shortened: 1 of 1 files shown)


uiolee:
Thank you for your efforts.

Is there any reason to use complex html instead of simple markdown? 
Personally, I don't think i can accept it.

lmliheng:
> 谢谢你的努力。
> 
> 有什么理由使用复杂的 html 而不是简单的 Markdown 吗？就个人而言，我认为我不能接受。

Okay, but it's definitely more beautiful, and it's a one-time deal

yoshinorin:
Thank you for submitting the PR. 

However, we do not see a sufficient reason to accept this PR. Aesthetic judgments are subjective and generally do not constitute a solid basis for acceptance. In this case, it seems to fall within the personal preference. Additionally, no one can ensure that this will be a one-time change.

PR #5555, opened 2024-10-05
feat: bind hexo context to helper function callback

dimaslanjaka:
## What does it do?

- Apply hexo context for `this` callback, useful for plugin written with typescript. [Sample](https://github.com/dimaslanjaka/hexo-renderers/blob/147340f6d03a8d3103e9589ddf86778ed7f9019b/src/helper/related-posts.ts#L106-L113)
- Fix some missmatch type (fail compile fixed)

## Screenshots

image
image

based on https://hexo.io/api/helper#How-do-I-use-another-registered-helper-in-my-custom-helper
image

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/filter/template_locals/i18n.ts
+++ b/lib/plugins/filter/template_locals/i18n.ts
@@ -13,9 +13,9 @@ function i18nLocalsFilter(this: Hexo, locals: LocalsType): void {
 
   if (!lang) {
     const pattern = new Pattern(`${i18nDir}/*path`);
-    const data = pattern.match(locals.path);
+    const data = pattern.match(locals.path) as Record<string, any>;
 
-    if (data && data.lang && i18nLanguages.includes(data.lang)) {
+    if (data && typeof data.lang === 'string' && i18nLanguages.includes(data.lang)) {
       lang = data.lang;
       page.canonical_path = data.path;
     } else {
--- a/lib/plugins/filter/template_locals/i18n.ts
+++ b/lib/plugins/filter/template_locals/i18n.ts
@@ -13,9 +13,9 @@ function i18nLocalsFilter(this: Hexo, locals: LocalsType): void {
 
   if (!lang) {
     const pattern = new Pattern(`${i18nDir}/*path`);
-    const data = pattern.match(locals.path) as Record<string, any>;
+    const data = pattern.match(locals.path) as any;
 
-    if (data && typeof data.lang === 'string' && i18nLanguages.includes(data.lang)) {
+    if (data && 'lang' in data && i18nLanguages.includes(data.lang)) {
       lang = data.lang;
       page.canonical_path = data.path;
     } else {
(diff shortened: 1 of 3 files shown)


dimaslanjaka:
@uiolee review pls, any miss/wrong ?

lib/plugins/filter/template_locals/i18n.ts
```diff
@@ -13,9 +13,9 @@ function i18nLocalsFilter(this: Hexo, locals: LocalsType): void {
 
   if (!lang) {
     const pattern = new Pattern(`${i18nDir}/*path`);
-    const data = pattern.match(locals.path);
+    const data = pattern.match(locals.path) as Record<string, any>;
 
-    if (data && data.lang && i18nLanguages.includes(data.lang)) {
+    if (data && typeof data.lang === 'string' && i18nLanguages.includes(data.lang)) {
```
SukkaW, line 18: Use `"lang" in data` to properly guard the type, not use `as Record`

SukkaW (changes requested):

lib/plugins/filter/template_locals/i18n.ts
dimaslanjaka: when i remove `as Record<string, any>;` from `data` variable got this error

image

dimaslanjaka: when i cast it to `any`, no error showing and successful build.

image

SukkaW: Do `typeof data === 'object' && 'lang' in data`. You can always narrow the type down. Do not cheat using `as` or `any`.

dimaslanjaka: > Do `typeof data === 'object' && 'lang' in data` then. You can always narrow the type down. Do not cheat using `as` or `any`.

`typeof data === 'object'` it makes property of object `unknown`

image

dimaslanjaka: > Do `typeof data === 'object' && 'lang' in data` then. You can always narrow the type down. Do not cheat using `as` or `any`.

`RegExpMatchArray` also result `object` when using `typeof`, also for Array, and null when you check using `typeof` it return `object`. This is typescript

dimaslanjaka: You cannot generalizing type variable checking by `object`.
For example `Buffer` when you check it using `typeof` the result also `object`. You should using `Buffer.isBuffer` to check the variable is buffer or not.
And array, you should using `Array.isArray` to check that is type of array or not.
when you generalized all to `typeof variable === object` it will always true.
That why, typescript makes data to error. because of this, and the `data` variable has 3 different types including Array, object and boolean.

dimaslanjaka: i dont know again to keep this working without enforcing cast to `any`, this caused by tsconfig.json `strict` rule.

image

fail to run `npm run build` when not enforcing cast to any

renbaoshuo (approved):
Looks good to me.

PR #5568, opened 2024-10-29
fix: wrong type for StoreFunction of helper

Mufanc:
## What does it do?

The `require` actually leads to the imported type becoming `any`, which will cause some return value of helpers that actually return `boolean` or complex types to be incorrectly threated as `string`, yet cannot be detected by eslint, for example here:

https://github.com/hexojs/hexo/blob/492debff7c3eb697594a516ab36bed96c94a8cc1/lib/plugins/helper/index.ts#L31-L41

and here:

https://github.com/hexojs/hexo/blob/492debff7c3eb697594a516ab36bed96c94a8cc1/lib/plugins/helper/index.ts#L69-L71

And this might lead to some undefined behaviors, so I think the return type of `StoreFunction` in the helper should be changed to `any`.

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/extend/helper.ts
+++ b/lib/extend/helper.ts
@@ -1,5 +1,5 @@
 interface StoreFunction {
-  (...args: any[]): string;
+  (...args: any[]): any;
 }
 
 interface Store {


stevenjoezhang:
Thanks for your contribution! The fix is included in https://github.com/hexojs/hexo/pull/5555

PR #5573, opened 2024-11-15
changed the judgement of whether it has been injected

mRNA16:
## What does it do?
When I introduced the Hexo injector in my blog post: 'Here we can see that between  and  there are many div blocks that we injected using the injector, and at the same time, between  and  there are JavaScript script blocks and canvas elements that we injected. I believe by now you can understand the functionality of the injector.' However, I found that the custom effects implemented through the injector were not working, so I checked the source code and found an issue with the logic that determines whether the injector has inserted content. When content like hexo injector body exists in my article, the injector mistakenly thinks it has already been inserted.
I tried to ensure that the comments and tags are tightly connected by using the condition (flag==='head_begin' && /<head.*?>\s*[\S\s]*/.test(input)), but this causes the test cases in the test to fail. So, I changed the condition to if (input.includes( and  in the article, the injector stopped working."

## Pull request tasks

- [ ] Passed the CI test.


mRNA16:
Oh，the comments mentioned in my blog text couldn't show in the website too!I mentioned the comments that mark the injector region here:' between     ( )    and     ( )        there are JavaScript script blocks and canvas elements that we injected. I believe by now you can understand the functionality of the injector.'

mRNA16:
I have made modifications to node_modules/hexo/dist/extend/injector.js locally. Both methods I proposed can solve the above issue. The changes I suggested should fix the bug. Please review my PR，thx！
屏幕截图 2024-12-24 223417
屏幕截图 2024-12-24 223713
屏幕截图 2024-12-24 223611

stevenjoezhang (approved):

PR #5578, opened 2024-11-30
feat: add url config validation

XuXianTao:
## What does it do?
Add `url` config validation.

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/hexo/validate_config.ts
+++ b/lib/hexo/validate_config.ts
@@ -12,6 +12,8 @@ export = (ctx: Hexo): void => {
   try {
     // eslint-disable-next-line no-new
     new URL(config.url);
+    // eslint-disable-next-line no-new
+    new URL('source/_post/xxx', config.url);
   } catch {
     throw new TypeError('Invalid config detected: "url" should be a valid URL!');
   }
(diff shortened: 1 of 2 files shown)


lib/hexo/validate_config.ts
```diff
@@ -12,6 +12,8 @@ export = (ctx: Hexo): void => {
   try {
     // eslint-disable-next-line no-new
     new URL(config.url);
+    // eslint-disable-next-line no-new
+    new URL('source/_post/xxx', config.url);
```
SukkaW, line 16: Maybe change `source/_post/xxx` to `config.root`, `config.permalink`, or something similar?

SukkaW (commented):
I

lib/hexo/validate_config.ts
XuXianTao: It's a good idea, but I think it will fail because of wrong `config.root` or `config.permalink`, not like the error description `Invalid config detected: "url" should be a valid URL!`.

I think `config.root`, `config.permalink` should be check in other part. And the `config.root` is already checked after this.

So, I think a simple string variable may be better.

ShaytonXu: > How about
> 
> ```js
> assert(new URL(config.url).protocol.startsWith('http'))
> ```
> 
> `http://localhost:4000` passes the test, but `localhost:4000` will not

Got it, done~

stevenjoezhang (approved):

SukkaW (approved):

PR #5591, opened 2024-12-22
feat(load_config): enforce stricter extension checks

stevenjoezhang:
## What does it do?

I've noticed that Hexo supports searching for configuration files with different suffixes, such as `_config.txt` and `_config.json`. However, this check is very loose, meaning that `_config.backup` or `_config.yml.bak` are also considered valid configuration files. This feature can cause confusion and even security issues for users. Therefore, I refactor this feature in the pull request.

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/hexo/load_config.ts
+++ b/lib/hexo/load_config.ts
@@ -1,4 +1,4 @@
-import { sep, resolve, join, parse } from 'path';
+import { sep, resolve, join, parse, basename, extname } from 'path';
 import tildify from 'tildify';
 import Theme from '../theme';
 import Source from './source';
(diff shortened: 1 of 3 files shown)


SukkaW (changes requested):
I still prefer to choose `config.yml` or `config.json` automatically without specifying it. We might wanna make `findConfigPath` more strict though.

stevenjoezhang:
@SukkaW I've updated the code, and now the JSON suffix is also supported.

SukkaW (changes requested):

yoshinorin (approved):
LGTM :heart:

PR #5592, opened 2024-12-22
chore: drop Node.js 16

stevenjoezhang:
## What does it do?

Versions of Node.js prior than 16 are already in End-Of-Life status. We should upgrade the supported Node.js versions in `package.json`. The required Node.js versions for other Hexo plugins also need to be upgraded accordingly.

## Pull request tasks

- [ ] Passed the CI test.


yoshinorin (approved):
👍

PR #5596, opened 2024-12-27
perf(postGenerator): delete unnecessary `toArray`

yoshinorin:
## What does it do?

`locals.posts` is iterable, there is no need to use [toArray](https://github.com/hexojs/warehouse/blob/b550c426b238a1aa24aa48a92c5c4606b8befb2a/src/model.ts#L440-L448) again here. `toArray` scans the data once, so this change is expected to slightly improve processing speed and reduce memory usage.

## Screenshots

N/A

## Pull request tasks

- [ ] Passed the CI test.

### DIFF
--- a/lib/plugins/generator/post.ts
+++ b/lib/plugins/generator/post.ts
@@ -2,7 +2,7 @@ import type { PostGenerator, PostSchema, SiteLocals } from '../../types';
 import type Document from 'warehouse/dist/document';
 
 function postGenerator(locals: SiteLocals): PostGenerator[] {
-  const posts = locals.posts.sort('-date').toArray();
+  const posts = locals.posts.sort('-date');
   const { length } = posts;
 
   return posts.map((post: Document<PostSchema>, i: number) => {


yoshinorin:
> TypeError: Cannot read properties of undefined (reading '_id')
> at Context.<anonymous> (/home/runner/work/hexo/hexo/test/scripts/generators/post.ts:76:24)

🤔

D-Sketon:
https://github.com/hexojs/hexo/blob/6637528a5dd78d231832bca3faeb1b405146c0e7/lib/plugins/generator/post.ts#L19

Before PR posts are `Document<PostSchema>[]`, after PR posts are `Query<PostSchema`, `posts[i+1]` is undefined

D-Sketon:
Notice that `locals.posts.sort('-date')` returns `Query`, so it actually calls https://github.com/hexojs/warehouse/blob/b550c426b238a1aa24aa48a92c5c4606b8befb2a/src/query.ts#L51-L53

SukkaW:
> https://github.com/hexojs/hexo/blob/6637528a5dd78d231832bca3faeb1b405146c0e7/lib/plugins/generator/post.ts#L19
> 
> Before PR `posts` are `Document<PostSchema>[]`, after PR `posts` are `Query<PostSchema>`, `posts[i+1]` is `undefined`

We do have `.eq` on `Query`, so we can probably do `posts.eq(i + 1)`.

yoshinorin:
> Before PR posts are Document<PostSchema>[], after PR posts are Query<PostSchema>, posts[i+1] is undefined

Thank you ❤️ 

[Query sets Document<T>[] in the constructor](https://github.com/hexojs/warehouse/blob/b550c426b238a1aa24aa48a92c5c4606b8befb2a/src/query.ts#L20). Since `toArray` is just returning that, so my understanding is that removing `toArray` in this PR does not affect performance.

I will close this PR.

PR #5611, opened 2025-01-09
feat: add permalink variable `timestamp`

### LINKED ISSUE #5586: hexo permalink timestamp
### Check List

- [X] I have already read [Docs page](https://hexo.io/docs/).
- [X] I have already searched existing issues.

### Feature Request

hexo支持 时间戳 作为 permalink 变量吗
```
permalink: post/:timestamp.html
```
---
```
---
data: 2024-12-10 12:10:46
---
```

时间戳：1733803846
效果：https://demo.com/post/1733803846.html

---

看到之前有人问过相关的：https://github.com/hexojs/hexo/issues/1595
如果现在不支持的话，以后会更新进去吗？

### Others

_No response_

uiolee:
## What does it do?

fix #5586

## Pull request tasks

- [x] Passed the CI test.
- [x] https://github.com/hexojs/site/pull/2326

### DIFF
--- a/lib/plugins/filter/post_permalink.ts
+++ b/lib/plugins/filter/post_permalink.ts
@@ -35,6 +35,7 @@ function postPermalinkFilter(this: Hexo, data: PostSchema): string {
     second: date.format('ss'),
     i_month: date.format('M'),
     i_day: date.format('D'),
+    timestamp: date.format('X'),
     hash,
     category: config.default_category
   };
--- a/test/scripts/filters/post_permalink.ts
+++ b/test/scripts/filters/post_permalink.ts
@@ -96,6 +96,29 @@ describe('post_permalink', () => {
     Post.removeById(post._id);
   });
 
+  it('timestamp', async () => {
+    hexo.config.permalink = ':timestamp/:slug';
+    const timestamp = '1736401514';
+    const dates = [
+      moment('2025-01-09 05:45:14Z'),
+      moment('2025-01-08 22:45:14-07')
+    ];
+    const posts = await Post.insert(
+      dates.map((date, idx) => {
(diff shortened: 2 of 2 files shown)


uiolee:
Something wrong when working on machines with different time zones

SukkaW (approved):

SukkaW (commented):
Should we add `:update_timestamp` as well IMHO?

uiolee:
> Should we add `:update_timestamp` as well IMHO?

I'm not sure if there is such a use case. But generally speaking, `permalink` should not change often

lib/plugins/filter/post_permalink.ts
```diff
@@ -35,6 +35,7 @@ function postPermalinkFilter(this: Hexo, data: PostSchema): string {
     second: date.format('ss'),
     i_month: date.format('M'),
     i_day: date.format('D'),
+    timestamp: date.format('X'),
     hash,
     category: config.default_category
```
SukkaW, line 40: Just a thought, we could use an approach like the following to only calculate the permalink value when needed, and cache it afterward:

```
function defineLazyProperty(object, propertyName, valueGetter) {
	const define = value => Object.defineProperty(object, propertyName, {value, enumerable: true, writable: true});

	Object.defineProperty(object, propertyName, {
		configurable: true,
		enumerable: true,
		get() {
			const result = valueGetter();
			define(result);
			return result;
		},
		set(value) {
			define(value);
		}
	});

	return object;
}


const permalink = {};

defineLazyProperty(permalink, 'second', () => date.format('ss'));
// ...
defineLazyProperty(permalink, 'timestamp', () => date.format('X'));
```

Or we can use `Map` instead of `object` here.

We can do this in another PR.

SukkaW (approved):

PR #5620, opened 2025-01-19
perf(escapeAllSwigTags): reducing GC overhead

D-Sketon:
## What does it do?

A large amount of string concatenation can result in numerous consString objects, which cannot be cleared during Scavenge. This leads to old generation expansion.

12x many-posts
before PR
{7542781E-F3A2-469E-8F24-D23D2BD2F2E7}
after PR
{405183C1-BE25-4F12-AB74-3F6AFD5EE625}

12x many-posts concurrency=20
before PR
{078D8B04-6C65-4DC3-824B-8310CFD81762}
after PR
{7CCECB0D-DD2C-4BA3-807D-6C52B52EDB0C}

4x many-posts concurrency=20
before PR
{80FEC4BA-F199-46A7-9B7E-B81BF2ADD67B}

after PR
{28230638-86F6-4081-A1ED-5F6DC02A9506}

1x many-posts concurrency=20
before PR
{CBF42156-63A6-47BC-A8B2-6FA2C0BB5CCC}
after PR
{13F8198A-B0EF-412A-93CE-F21BDA67EDE0}

## Pull request tasks

- [x] Passed the CI test.


SukkaW (changes requested):
I believe that the root cause of the issue is that string concatenation doesn't copy strings; instead, it points to the existing strings using pointers. This is good for performance, so we should retain this behavior.

On the other hand, the concatenated string holds the reference of the old string, preventing the old string from being released. 

Instead of manually creating a `StringBuilder`, you can try replacing `return output` with `return output.replace('#'.repeat(output.length + 1), '')` to see if it reduces GC pressure.

See also: https://romgrk.com/posts/optimizing-javascript#8-use-strings-carefully

D-Sketon:
I might have figured out why there's a memory difference.

Although `consString` uses references, each `consString` itself carries a size of 32 bytes.

Assuming each post has 5,000 characters and there are 4,000 posts, `consString` would occupy an additional 5,000 * 4,000 * 32B = 640MB.

On the other hand, using `StringBuilder` would result in each post taking an additional 32 bytes (I’m not entirely sure), totaling an additional 4,000 * 32B = 128KB.

SukkaW (approved):

renbaoshuo (approved):
Looks good to me.

PR #5624, opened 2025-01-25
perf(listArchives): add cache for posts

D-Sketon:
## What does it do?

add cache for posts when relative_link is true(we cannot use cache in template)
before
{5BA484CC-9C9D-493D-ADE9-05545841F6A0}
after
{470B797A-99C5-41A5-996E-B67548D22020}

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/extend/injector.ts
+++ b/lib/extend/injector.ts
@@ -44,7 +44,7 @@ class Injector {
   }
 
   getSize(entry: Entry): number {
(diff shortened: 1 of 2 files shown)


SukkaW (approved):
I am wondering if it is possible to add caches directly to the `warehouse` though, and invalidate automatically when the data is updated.

SukkaW (approved):

PR #5625, opened 2025-01-27
feat: Support additional options for Backtick Code Block

### LINKED ISSUE #4830: The Backtick Code Block does not support additional options

D-Sketon:
## What does it do?

resolve #4830

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/filter/before_post_render/backtick_code_block.ts
+++ b/lib/plugins/filter/before_post_render/backtick_code_block.ts
@@ -23,9 +23,7 @@ function parseArgs(args: string) {
   const mark: number[] = [];
   let firstLine = 1;
   for (let i = 0; i < len; i++) {
-    const kv = matches[i].split(':');
-    const key = kv[0];
-    const value = kv[1];
+    const [key, value] = matches[i].split(':');
 
     switch (key) {
       case 'highlight':
@@ -51,9 +49,7 @@ function parseArgs(args: string) {
             let b = +cur.slice(hyphen + 1);
             if (Number.isNaN(a) || Number.isNaN(b)) continue;
             if (b < a) { // switch a & b
-              const temp = a;
-              a = b;
-              b = temp;
+              [a, b] = [b, a];
             }
 
             for (; a <= b; a++) {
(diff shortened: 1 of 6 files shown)


uiolee (dismissed):

lib/plugins/filter/before_post_render/backtick_code_block.ts
```diff
@@ -1,17 +1,87 @@
...
+    matches.push(match[1]);
+  }
+
+  const len = matches.length;
+  const mark: number[] = [];
+  let firstLine = 1;
+  for (let i = 0; i < len; i++) {
+    const kv = matches[i].split(':');
+    const key = kv[0];
+    const value = kv[1];
```
uiolee, line 28: it could be `const [ket, value] = matches[i].split(':');`

lib/plugins/filter/before_post_render/backtick_code_block.ts
```diff
@@ -1,17 +1,87 @@
...
+      case 'mark': {
+        for (const cur of value.split(',')) {
+          const hyphen = cur.indexOf('-');
+          if (hyphen !== -1) {
+            let a = +cur.slice(0, hyphen);
+            let b = +cur.slice(hyphen + 1);
+            if (Number.isNaN(a) || Number.isNaN(b)) continue;
+            if (b < a) { // switch a & b
+              const temp = a;
+              a = b;
```
uiolee, line 55: `[a, b] = [b, a]`

PR #5629, opened 2025-02-09
Fix clean console

dimaslanjaka:
## What does it do?

- test script failed error, because original binded function should have parameter, but the usage not have it. here the screenshoot.

image

- On windows 10 git scm, the git hook file should have shebang `#!/bin/sh`, or will got error like below screenshoot

image

## Pull request tasks

- [x] Passed the CI test.


dimaslanjaka:
@SukkaW need review

.husky/pre-commit
```diff
@@ -1 +1,3 @@
+#!/bin/sh
+
```
uiolee, line 2: https://github.com/typicode/husky/releases/tag/v9.0.1#:~:text=shebang

dimaslanjaka: but got error on my windows, see the screenshoot

dimaslanjaka: git scm windows version still need shebang. and shebang still allowed by husky. here the doc

image

uiolee: I could see your screenshot of error. However, I could not reproduce on my window.

lib/plugins/console/clean.ts
```diff
@@ -2,7 +2,8 @@ import Promise from 'bluebird';
 import { exists, unlink, rmdir } from 'hexo-fs';
 import type Hexo from '../../hexo';
 
-function cleanConsole(this: Hexo): Promise<[void, void, any]> {
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
uiolee, line 5:
```suggestion
```

Even if I delete it, eslint have not report any errors.

uiolee:
I can't reproduce the two questions you mentioned, hexo ci also shows normal. I'll suspect the problem comes from your operating environment

Issue #5635, opened 2025-02-20
Markdown text is escaped when two tags share the same prefix.


zombie12138:
### Check List

- [x] I have already read [Docs page](https://hexo.io/docs/) & [Troubleshooting page](https://hexo.io/docs/troubleshooting).
- [x] I have already searched existing issues and they are not help to me.
- [x] I examined error or warning messages and it's difficult to solve.
- [x] I am using the [latest](https://github.com/hexojs/hexo/releases) version of Hexo. (run `hexo version` to check)
- [x] My Node.js is matched [the required version](https://hexo.io/docs/#Required-Node-js-version).

### Expected behavior

The Post:

```md
---
title: Hello World
date: 2021-10-09 21:50:25
refplus: true
---
<!-- more -->
Test {% ref self %}.
## Ref
{% references %}
[self] {% post_link hello-world hello-world %}
{% endreferences %}
```

All Markdown content between the `ref` and `endreferences` tags is rendered properly. And `## Ref` is rendered as an `h2` element.

### Actual behavior

All Markdown content between the ref and endreferences tags is escaped, resulting in unrendered text.

<img width="673" alt="Image" src="https://github.com/user-attachments/assets/0b7f92f3-8a6f-4d5d-ae31-c43ff6e9e204" />

### How to reproduce?

```bash
sudo apt install npm
sudo npm install hexo@7.3.0 -g
hexo init testenv
cd testenv
npm install hexo-reference-plus
cd source/_posts/
tee >hello-world.md  <<EOF
---
title: Hello World
date: 2021-10-09 21:50:25
refplus: true
---
<!-- more -->
Test {% ref self %}.
## Ref
{% references %}
[self] {% post_link hello-world hello-world %}
{% endreferences %}
EOF
```

### Is the problem still there under `Safe mode`?

Yes

### Your Node.js & npm version

```text
v18.19.1
9.2.0
```

### Your Hexo and Plugin version

```text
hexo-site@0.0.0 /home/admin/test/testenv
├── hexo-generator-archive@2.0.0
├── hexo-generator-category@2.0.0
├── hexo-generator-index@4.0.0
├── hexo-generator-tag@2.0.0
├── hexo-reference-plus@1.1.0
├── hexo-renderer-ejs@2.0.0
├── hexo-renderer-marked@7.0.0
├── hexo-renderer-stylus@3.0.1
├── hexo-server@3.0.0
├── hexo-theme-landscape@1.0.0
└── hexo@7.3.0
```

[...]

PR #5636, opened 2025-02-23
fix(escapeAllSwigTags): Prevent swig tag prefix collision during esca…

zombie12138:
## What does it do?

fix bug in #5635

When a non-block Swig tag (Tag A) serves as a prefix for a block-level tag (Tag B), this hierarchical relationship triggers unintended Markdown escaping, leading to unrendered content.

This pull request resolves this case by implementing a block tag registry table. This methodology achieves maximum precision, but it introduces increased code coupling as an implementation trade-off.

## Pull request tasks

- [ ] Passed the CI test.


lib/extend/tag.ts
```diff
@@ -254,6 +267,7 @@ class Tag {
     const { env } = this;
 
     if (env.hasExtension(name)) env.removeExtension(name);
+    delete this.block_swig_tag_map[name];
```
SukkaW, line 270: Why not `this.block_swig_tag_map[name] = false`?

SukkaW (changes requested):

PR #5650, opened 2025-04-05
perf: skip tag render when there is no swigTags

D-Sketon:
Can we skip `tag.render` if there are no swig Tags?


lib/hexo/post.ts
```diff
@@ -517,7 +514,8 @@ class Post {
     }).then(() => {
       data.content = cacheObj.escapeCodeBlocks(data.content);
       // Escape all Nunjucks/Swig tags
-      if (disableNunjucks === false) {
+      const hasTag = /(\{\{.+?\}\})|(\{#.+?#\})|(\{%.+?%\})/s.test(data.content);
```
SukkaW, line 517: Hoist the regexp.

We forgot to do this last time, should not miss this again.

lib/hexo/post.ts
```diff
@@ -517,7 +514,8 @@ class Post {
...
+      if (disableNunjucks === false && hasTag) {
```
SukkaW, line 518: If we already disable nunjucks, we don't have to test anymore.

SukkaW (changes requested):

lib/hexo/post.ts
```diff
@@ -517,8 +515,12 @@ class Post {
     }).then(() => {
       data.content = cacheObj.escapeCodeBlocks(data.content);
       // Escape all Nunjucks/Swig tags
+      let hasTag = true;
```
yoshinorin, line 518: IMHO:

I prefer to make it clear what kind of tag it is in the variable name like `hasSwigTag`.

lib/hexo/post.ts
```diff
@@ -536,9 +538,9 @@ class Post {
           data.content = cacheObj.restoreAllSwigTags(content);
 
           // Return content after replace the placeholders
-          if (disableNunjucks) return data.content;
+          if (disableNunjucks || !hasTag) return data.content;
```
yoshinorin, line 541: The conditional operator here is a bit hard to follow, but since I can't think of a better solution, I think this is fine as it is.

yoshinorin (commented):
LGTM 👍

SukkaW (approved):

Issue #3282, opened 2018-10-11
permalink date is incorrect if the time zone of _config.yml and the time zone setting of the machine are different


MatchaChoco010:
## Environment Info

Node version(`node -v`):v10.9.0

site to reproduce the ploblem:

[https://github.com/MatchaChoco010/hexo-timezone-permalink](https://github.com/MatchaChoco010/hexo-timezone-permalink)

```
$ hexo version
hexo: 3.7.1
hexo-cli: 1.1.0
os: Linux 4.4.0-17134-Microsoft linux x64
http_parser: 2.8.0
node: 10.9.0
v8: 6.8.275.24-node.14
uv: 1.22.0
zlib: 1.2.11
ares: 1.14.0
modules: 64
nghttp2: 1.32.0
napi: 3
openssl: 1.1.0i
icu: 62.1
unicode: 11.0
cldr: 33.1
tz: 2018e
```

## For BUG

_config.yml is this:

``` yaml _config.yml
...
timezone: America/New_York
...
permalink: :year/:month/:day/:title/
...
```

and `source/_posts/test.md` is this:

``` yaml front_matter
---
title: test
date: 2018-10-11 23:00:00
tags:
---
```

The permalink of this article should be `/2018/10/11/test/`.
But `hexo generate` in Japan, which +0900, `/2018/10/12/test/index.html` is generated.

![screenshot0](https://user-images.githubusercontent.com/20170107/46803714-aebc4880-cd9b-11e8-9df2-9efa3696f3f8.png)

The article date is correct (2018/10/11).

![screenshot1](https://user-images.githubusercontent.com/20170107/46803726-b54ac000-cd9b-11e8-92ec-e367f463789d.png)

Change the machine timezone to -0500 and `hexo generate`, `/2018/10/11/test.index.html` is generated.

It seems that the permalink is not generated correctly if the time zone of _config.yml and the time zone setting of the machine are incorrect.

YoshinoriN:
@MatchaChoco010 
Thanks a report. I reproduced it.
But, I can't judge fix this behavior or not. I think this problem impact is big.

I ask to other hexo maintainer this problem.

YoshinoriN:
@tommy351 @NoahDragon @JLHwung 
How do you think about this.

itszero:
[...]

PR #5654, opened 2025-04-20
refactor(moment): remove unused timezone setting

stevenjoezhang:
## What does it do?

1. According to https://github.com/hexojs/hexo/issues/3282#issuecomment-462077102, variables in `ctx.config` in `lib/models/page.ts` and `lib/models/post.ts` are Hexo default config, instead of variables from `_config.yml`.
Removing the code only affects its own unit tests, no other functionalities are affected.
I propose to remove it because it's not used and may cause confusion when developing timezone-related features.
2. Besides, I renamed `timezone` function in `common.ts` to `adjustDateForTimezone`, to make the name more precise.

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/models/types/moment.ts
+++ b/lib/models/types/moment.ts
@@ -20,11 +20,7 @@ class SchemaTypeMoment extends warehouse.SchemaType<moment.Moment> {
     value = super.cast(value, data);
     if (value == null) return value;
 
-    const { options } = this;
     value = toMoment(value);
-
-    if (options.language) value = value.locale(toMomentLocale(options.language));
-    if (options.timezone) value = value.tz(options.timezone);
 
     return value;
   }
--- a/lib/models/types/moment.ts
+++ b/lib/models/types/moment.ts
@@ -1,5 +1,5 @@
 import warehouse from 'warehouse';
-import { moment, toMomentLocale } from '../../plugins/helper/date';
+import { moment } from '../../plugins/helper/date';
 
 // It'll pollute the moment module.
 // declare module 'moment' {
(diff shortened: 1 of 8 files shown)


lib/models/types/moment.ts
yoshinorin, line 1: We can remove `toMomentLocale` from the `import` at the top of this file.

```diff
- import { moment, toMomentLocale } from '../../plugins/helper/date';
+ import { moment } from '../../plugins/helper/date';
```

lib/models/types/moment.ts
```diff
@@ -20,12 +20,8 @@ class SchemaTypeMoment extends warehouse.SchemaType<moment.Moment> {
     value = super.cast(value, data);
     if (value == null) return value;
 
-    const { options } = this;
     value = toMoment(value);
 
-    if (options.language) value = value.locale(toMomentLocale(options.language));
-    if (options.timezone) value = value.tz(options.timezone);
-
     return value;
```
yoshinorin, line 25: Personally, I prefer `return toMoment(value);`

yoshinorin (commented):
Regarding the title of the PR, shouldn't it be `refactor` instead of `feat`? 🤔

lib/models/types/moment.ts
stevenjoezhang: Updated

lib/models/types/moment.ts
stevenjoezhang: Updated

yoshinorin (approved):
LGTM 👌

PR #5599, opened 2024-12-30
chore(deps): update eslint

stevenjoezhang:
## What does it do?

See https://github.com/hexojs/eslint-config-hexo/pull/64

## Pull request tasks

- [ ] Add test cases for the changes.
- [ ] Passed the CI test.

PR #5655, opened 2025-04-22
chore(lint): apply eslint to test files

yoshinorin:
## What does it do?

I'm sorry for not noticing during my review of #5599 that the [`test` directory was excluded](https://github.com/hexojs/hexo/pull/5599/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R13) from linting. This PR makes the following changes:

1. Add the `test` directory to eslint scope.
2. Manage the rules for both lib and test directories in a single `eslint.config.js` file at the root.
    a. During the review of https://github.com/hexojs/warehouse/pull/286, I noticed that placing an `eslint.config.js` in the `test` directory has no effect.

## Screenshots

N/A

## Pull request tasks

- [x] Passed the CI test.


eslint.config.js
```diff
@@ -15,5 +17,26 @@ module.exports = [
...
       ]
     }
+  },
+  // Configurations applied only to test files
+  {
+    files: [
+      'test/**/*.ts'
+    ],
+    languageOptions: {
+      ...testConfig.languageOptions
```
yoshinorin, line 27: This approach is not preferable since it requires knowledge of what objects `testConfig` contains. However, I cannot think of a more appropriate method.

eslint.config.js
```diff
@@ -15,5 +17,26 @@ module.exports = [
...
+    },
+    rules: {
+      ...testConfig.rules,
+      '@typescript-eslint/ban-ts-comment': 0,
+      '@typescript-eslint/no-unused-expressions': 0,
+      '@typescript-eslint/no-unused-vars': [
```
yoshinorin, line 33: Since this is for tests, we might consider ignoring all no-unused-vars.

eslint.config.js
```diff
@@ -15,5 +17,26 @@ module.exports = [
...
+  // Configurations applied only to test files
+  {
+    files: [
+      'test/**/*.ts'
+    ],
+    languageOptions: {
+      ...testConfig.languageOptions
+    },
+    rules: {
+      ...testConfig.rules,
```
yoshinorin, line 30: Also here

test/scripts/hexo/post.ts
```diff
@@ -1534,7 +1534,7 @@ describe('Post', () => {
         engine: 'markdown'
       });
       should.fail();
-    } catch (err) {}
+    } catch (_) {}
```
stevenjoezhang, line 1537: The block can be removed 

```suggestion
    } catch {}
```

PR #5685, opened 2025-08-12
feat: add configurable image auto-copy for VSCode markdown editing

JKWTCN:
Add use_vscode_edit_md config to enable automatic copying of images from source/_posts/image to public/image, preventing "Cannot GET /image/" errors when pasting images in VSCode. Disabled by default for

## What does it do?
Adds a `use_vscode_edit_md` configuration option to enable automatic copying of images from `source/_posts/image` to `public/image` when processing markdown files. This addresses the "Cannot GET /image/" errors that occur when pasting images in VSCode.

- The feature is disabled by default to maintain backward compatibility
- When enabled, it ensures images referenced in markdown posts are properly copied to the public directory
- Solves the common workflow issue where images pasted via VSCode's markdown editor are saved in the post's folder but not deployed

## Screenshots
image
image

## Pull request tasks

- [ x] Add test cases for the changes.
- [ x] Passed the CI test.

### DIFF
--- a/lib/hexo/default_config.ts
+++ b/lib/hexo/default_config.ts
@@ -84,6 +84,9 @@ export = {
   // ignore files from processing
   ignore: [] as string[],
 
+  // VS Code integration
+  use_vscode_edit_md: false,
+
   // Category & Tag
   meta_generator: true
 };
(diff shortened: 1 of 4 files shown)


yoshinorin (commented):
Thank you very much for taking the time to submit this PR. However, IMHO, I don’t think we should merge this one. There are two reasons:

1. The Hexo config is already quite messy, and I’d prefer not to add more settings if possible.
2. This PR depends on VSCode. The Hexo core should avoid such dependencies as much as possible.

The second point is especially important. In this case, it could probably be implemented either as a VSCode plugin or a Hexo plugin.

PR #5697, opened 2025-10-01
fix(regression): missing tags and categories when running `hexo s` with cache existed

D-Sketon:
## What does it do?

I forgot to load the binary relation index after loadDatabase during the watch phase. This caused categories and tags to be empty when running hexo s for preview with an existing db.json.

fix https://github.com/orgs/hexojs/discussions/5692

## Pull request tasks

- [x] Passed the CI test.


jerryc127:
原来这是 bug 啊。。

yoshinorin (approved):
LGTM

PR #5698, opened 2025-10-02
perf: reduce ObjectAssign overhead

D-Sketon:
## What does it do?

Interestingly, just modifying this single line can significantly reduce the overhead of ObjectAssign.

## Screenshots
8x hexo-many-posts + hexo-theme-reimu

before
{5CD5CAD9-97B6-4AE7-9DC7-CBD69F5FE496}

after
{DB9C014C-0B5B-4420-90B7-A790DB641960}

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/helper/partial.ts
+++ b/lib/plugins/helper/partial.ts
@@ -15,7 +15,7 @@ export = (ctx: Hexo) => function partial(this: LocalsType, name: string, locals?
   const currentView = this.filename.substring(viewDir.length);
   const path = join(dirname(currentView), name);
   const view = ctx.theme.getView(path) || ctx.theme.getView(name);
-  const viewLocals = { layout: false };
+  const viewLocals: Record<string, any> = {};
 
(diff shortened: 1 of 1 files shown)


SukkaW (approved):
Basically, `Object.assign` is just reading and assigning properties and values one by one, that's why the following use case of `Object.assign` also works:

```ts
Object.assign(htmlElement.style, { fontSize: '12px', color: 'navy' });
// This is basically the same as
htmlElement.style.fontSize = '12px';
htmlElement.style.color = 'navy';

Object.assign(function() {}, { extraPropOnFunction });
// This is basically the same as
(function() {}).extraPropOnFunction = extraPropOnFunction;
```

Thus, `Object.assign` on a plain object will also face the same `Monomorphism` JIT issue of V8. If we predefine the shape of the object, V8 will have to de-optimize later. But if we don't define the shape, V8 will determine the shape itself.

Issue #5681, opened 2025-07-31
asset_img not working with post_asset_folder: true


Nutingnon:
### Check List

- [x] I have already read [Docs page](https://hexo.io/docs/) & [Troubleshooting page](https://hexo.io/docs/troubleshooting).
- [x] I have already searched existing issues and they are not help to me.
- [x] I examined error or warning messages and it's difficult to solve.
- [x] I am using the [latest](https://github.com/hexojs/hexo/releases) version of Hexo. (run `hexo version` to check)
- [x] My Node.js is matched [the required version](https://hexo.io/docs/#Required-Node-js-version).

### Expected behavior

I can see the insert picture in my post when I have following content:

```markdown
This is the post no.1 

{% asset_img data_flow.jpg %}

The picture is expected to show above
```

the content should looks like

```
This is the post no.1 

<data_flow.jpg>

The picture is expected to show above

```

### Actual behavior

but the content actually looks like

```
This is the post no.1 

The picture is expected to show above

```

### How to reproduce?

Step. 1, setting `post_asset_folder: true`  in _config.yml
Step 2. create post by running hexo new post_1
Step 3. put the image to `./source/_posts/post_1/data_flow.jpg`
Step 4. write content in `./source/_posts/post_1.md`, with inserting img command as `{% asset_img data_flow.jpg %}`
Step 5. running `hexo server` in the terminal
Step 6. check the content, no picture appearing.

### Is the problem still there under `Safe mode`?

not under safe mode

### Your Node.js & npm version

```text
v22.17.1
11.5.1
```

### Your Hexo and Plugin version

```text
hexo-site@0.0.0 /home/yixin/blog
├── hexo-asset-image@1.0.0
├── hexo-generator-archive@2.0.0
├── hexo-generator-category@2.0.0
├── hexo-generator-index@4.0.0
├── hexo-generator-tag@2.0.0
├── hexo-renderer-ejs@2.0.0
├── hexo-renderer-marked@7.0.1
├── hexo-renderer-stylus@3.0.1
├── hexo-server@3.0.0
├── hexo-theme-landscape@1.1.0
└── hexo@7.3.0
```

### Your `package.json`

```json
{
  "name": "hexo-site",
  "version": "0.0.0",
  "private": true,
  "scripts": {
    "build": "hexo generate",
    "clean": "hexo clean",
    "deploy": "hexo deploy",
    "server": "hexo server"
  },
  "hexo": {
    "version": "7.3.0"
  },
[...]

Issue #5671, opened 2025-07-22
"hexo g" doesn't copy dir in source/_posts/ to public/post/


youyeyejie:
### Check List

- [x] I have already read [Docs page](https://hexo.io/docs/) & [Troubleshooting page](https://hexo.io/docs/troubleshooting).
- [x] I have already searched existing issues and they are not help to me.
- [x] I examined error or warning messages and it's difficult to solve.
- [x] I am using the [latest](https://github.com/hexojs/hexo/releases) version of Hexo. (run `hexo version` to check)
- [x] My Node.js is matched [the required version](https://hexo.io/docs/#Required-Node-js-version).

### Expected behavior

我装了hexo-asset-img插件，运行`hexo g`时，位于source/_post/下的文件夹下的内容不会被复制到public/post/ 下，即
I installed the hexo-asset-img plugin, and when I run hexo g, the content under the folder under source/_post/ will not be copied to public/post/, i.
```txt
source/
  |-_posts/
        |-test.md
        |-test/
             |-1.png
```
在“hexo g”之后，我希望它会是
after `hexo g` ,I hope it will be
```txt
public
  |-post/
        |-test/
             |-index.html
             |-1.png
```

### Actual behavior

在“hexo g”之后，实际上是
after `hexo g` , acctually it's
```txt
public
  |-post/
        |-test/
             |-index.html
```

只有运行 “hexo cl && hexo g" 才能实现预期效果
The desired effect can only be achieved by running `hexo cl & & hexo g`

### How to reproduce?

1. npm install hexo-asset-img --save
2. hexo new "test"
3. touch source/_posts/test/1.png
4. hexo g

### Is the problem still there under `Safe mode`?

no

### Your Node.js & npm version

```text
v22.17.0
11.4.2
```

### Your Hexo and Plugin version

```text
hexo-site@0.0.0 E:\Blog
├── font-awesome@4.7.0
├── hexo-asset-img@1.2.0
├── hexo-deployer-git@4.0.0
├── hexo-filter-github-emojis@3.0.5
├── hexo-filter-mermaid-diagrams@1.0.5
├── hexo-generator-archive@2.0.0
├── hexo-generator-category@2.0.0
├── hexo-generator-index@4.0.0
├── hexo-generator-search@2.4.3
├── hexo-generator-sitemap@3.0.1
├── hexo-generator-tag@2.0.0
├── hexo-renderer-ejs@2.0.0
├── hexo-renderer-pandoc@0.5.0
[...]

Issue #5659, opened 2025-04-29
Failed to Generate using Symbolic Link Directories as Assets


Omnisch:
### Check List

- [x] I have already read [Docs page](https://hexo.io/docs/) & [Troubleshooting page](https://hexo.io/docs/troubleshooting).
- [x] I have already searched existing issues and they are not help to me.
- [x] I examined error or warning messages and it's difficult to solve.
- [x] I am using the [latest](https://github.com/hexojs/hexo/releases) version of Hexo. (run `hexo version` to check)
- [x] My Node.js is matched [the required version](https://hexo.io/docs/#Required-Node-js-version).

### Expected behavior

To generate correctly with a **symbolic link directory** as an asset in the post asset folder.

### Actual behavior

Generated, but after `INFO  Files loaded in *.**s`, there is an error `Error: EISDIR: illegal operation on a directory, read`. And the symbolic link directory did not appear in `/public`.

### How to reproduce?

1. Set `post_asset_folder` to `true`.
2. Create a symbolic link or a junction using Windows cmd, like `mklink /D "path\to\source\_post\my_post\asset_link" "path\to\asset\in\somewhere\else"`.
3. Run `hexo clean && hexo g`.

### Is the problem still there under `Safe mode`?

Yes

### Your Node.js & npm version

```text
node v23.4.0
npm 10.9.2
```

### Your Hexo and Plugin version

```text
hexo-site@0.0.0
├── hexo-deployer-git@4.0.0
├── hexo-footnotes@1.0.2
├── hexo-generator-archive@2.0.0
├── hexo-generator-category@2.0.0
├── hexo-generator-index@4.0.0
├── hexo-generator-tag@2.0.0
├── hexo-hide-posts@0.4.3
├── hexo-renderer-ejs@2.0.0
├── hexo-renderer-markdown-it-plus@1.0.6
├── hexo-renderer-stylus@3.0.1
├── hexo-server@3.0.0
├── hexo-theme-landscape@1.0.0
├── hexo-theme-next@8.21.1
└── hexo@7.3.0
```

### Your `package.json`

```json
{
  "name": "hexo-site",
  "version": "0.0.0",
  "private": true,
  "scripts": {
    "build": "hexo generate",
    "clean": "hexo clean",
    "deploy": "hexo deploy",
    "server": "hexo server"
  },
  "hexo": {
    "version": "7.3.0"
  },
  "dependencies": {
    "hexo": "^7.3.0",
[...]

PR #5473, opened 2024-04-18
perf(processor/post): improve processing speed when `config.post_asset_folder` is enabled

yoshinorin:
- refs: 
    - https://github.com/hexojs/hexo/issues/2579#issuecomment-2053889785
    - https://github.com/hexojs/hexo/issues/5456

I think other issues also related this PR

## When slow?

`config.post_asset_folder` is enabled.

## Why?

When `config.post_asset_folder` is enabled, Hexo calls the `Post.toArray()` function for each file. The time complexity of `Post.toArray()` is `O(n)`. Additionally, Hexo uses the `Post.find` method to search for the post, and the average time complexity of `find` is also `O(n)`.

So, if there are 100 articles and 100 files, the time complexity would be `O(100 * 100 * 100) = O(n^3)`.

## How to solve?

Avoid calling `Post.toArray()`. Instead, we will search based on whether the directory where the files are stored (`assetDir`) matches. Also, if the file found with `PostAsset`, use `Post.findById`. The time complexity of `Post.findById` is `O(1)`.

## Benchmark Environment

Below is bench env.

<details>

### Machine 

```
# OS
Microsoft Windows [Version 10.0.22631.3447]

# Cpu
AMD Ryzen 7 PRO 4750G with Radeon Graphics

# Memory
Capacity     Name             Tag
17179869184  Physical Memory  Physical Memory 1
17179869184  Physical Memory  Physical Memory 3
```

### Hexo and Node.js

```
$ hexo -v
hexo: 7.2.0
hexo-cli: 4.3.1
os: win32 10.0.22631
node: 20.11.1
...
v8: 11.3.244.8-node.17
```

### `package.json`

```
"dependencies": {
  "0x": "^5.7.0",
  "hexo": "file://C:\\Users\\<userName>\\development\\hexo\\hexo",
  "hexo-filter-nofollow": "2.0.2",
  "hexo-generator-archive": "git+https://github.com/hexojs/hexo-generator-archive.git#master",
  "hexo-generator-category": "git+https://github.com/hexojs/hexo-generator-category.git#master",
  "hexo-generator-index": "git+https://github.com/hexojs/hexo-generator-index.git#master",
  "hexo-generator-sitemap": "git+https://github.com/yoshinorin/_hexo-generator-sitemap.git#master",
  "hexo-generator-tag": "git+https://github.com/hexojs/hexo-generator-tag.git#master",
  "hexo-pagination": "git+https://github.com/yoshinorin/hexo-pagination.git#my-site",
[...]

PR #5704, opened 2025-10-11
fix(processor/post): updated post assets not being copied in hot processing when `post_asset_folder` is enabled

yoshinorin:
The following issues will likely be fixed:

- https://github.com/hexojs/hexo/issues/5671
- https://github.com/hexojs/hexo/issues/5681
- https://github.com/hexojs/hexo/issues/5659

## What does it do?

This PR fixes an issue where updated post assets were not being copied in hot processing. This issue introduced in #5473.

## Cause

The `processAsset` function had three issues.

#### 1. Path separator mismatch:

As @D-Sketon pointed out in PR [#5473 comment](https://github.com/hexojs/hexo/pull/5473#issuecomment-3369127019), the original code used OS-specific `sep` (backslash `\` on Windows) when extracting the asset directory path. This caused the path matching to fail on Windows.

#### 2. Path format inconsistency: 

The extracted asset directory path was in relative format with forward slashes (e.g., `source/_posts/bar`), but `post.asset_dir` is an OS-specific
 absolute path (e.g., `C:\Users\...\source\_posts\bar\` on Windows). This format mismatch prevented retrieving the correct post.

#### 3. Incorrect `Post.findOne()` usage

`Post.findOne()` with both query function and object syntax returned incorrect results. It either returned the wrong post or `null`, preventing
assets from being associated with the correct post. I have not investigated why `Post.findOne()` doesn't work correctly here.

## Solution

- Use `posix.sep` (`/`) consistently when extracting the asset directory from the normalized `id`
- Convert the relative path to an OS-specific absolute path for comparison with `post.asset_dir`
- Replace `Post.findOne()` with `Post.filter()` to ensure correct post lookup

## Screenshots

N/A

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/test/scripts/extend/tag.ts
+++ b/test/scripts/extend/tag.ts
@@ -255,7 +255,7 @@ describe('Tag', () => {
       'date: 2023-09-03 16:59:46',
       'tags: foo',
       '---'
-    ];
+    ].join('\n');
 
     const file2 = newFile({
       path: 'test2.html',
@@ -269,7 +269,7 @@ describe('Tag', () => {
       'date: 2023-09-03 16:59:49',
       'tags: foo',
       '---'
-    ];
+    ].join('\n');
 
     const file3 = newFile({
       path: 'test3.html',


D-Sketon (approved):
works well on my windows

PR #5720, opened 2025-11-04
fix(moment): use timezone from Hexo config instead of maching settings

### LINKED ISSUE #4548: Date & Timezone issue related with Front-Matter
## Check List

Please check the following before submitting a new issue.

- [x] I have already searched existing issues and they are not helpful to me
- [x] Node.js is higher than [minimum required version](https://hexo.io/docs/#Minimum-required-Node-js-version)

The bug has been reported in

- #4239 
- #3397
- #3282 
- #2063 

-----

For example, I have a post setup as following:

```yaml
# _config.yml
timezone: Asia/Shanghai
```

```yaml
title: Hello World
date: 2020-09-29 23:00:00
```

After the front-matter being processed by `js-yaml`, the "Date-like string" will be converted into `Date` object:

```js
const jsYaml = require("js-yaml");

jsYaml.load('date: 2020-09-29 23:00:00');
// > "Tue, 29 Sep 2020 23:00:00 GMT" 
```

Notice that the input was converted into the environment local timezone? It is not what users desired. They will only fill in the front-matter with their local time, not the UTC time.

So during the processing of front-matter, `Date#getTimezoneOffset` has been used:

```js
// https://github.com/hexojs/hexo-front-matter/blob/ccbdff36d151a56932418cdc6d0329d866032a1b/lib/front_matter.js#L55-L62

// Convert timezone
Object.keys(data).forEach(key => {
  const item = data[key];

  if (item instanceof Date) {
    data[key] = new Date(item.getTime() + (item.getTimezoneOffset() * 60 * 1000));
  }
});
```

It will result in:

```
"Tue, 29 Sep 2020 15:00:00 GMT"
```

Which is the desired behavior.

Then, during the processing pf posts, `moment.timezone` was used:

https://github.com/hexojs/hexo/blob/557487a2f8ab0065bf94d5a9466c54034f1db17f/lib/plugins/processor/post.js#L80-L82
https://github.com/hexojs/hexo/blob/557487a2f8ab0065bf94d5a9466c54034f1db17f/lib/plugins/processor/common.js#L51-L60

If the `config.timezone` is configured correctly, there will be no differences between before and after `timezone()`:

```
Tue Sep 29 2020 20:00:00 GMT+0800 (China Standard Time)
// "Tue, 29 Sep 2020 15:00:00 GMT"
// There's no differences
```

A demo can be found here: https://runkit.com/sukkaw/5f732261a7aac8001a42bcc4

```
Date#getTimezoneOffset: -480
Date parsed by js-yaml: Tue, 29 Sep 2020 23:00:00 GMT
Date after calculating the timezone offset: Tue, 29 Sep 2020 15:00:00 GMT
Date after moment-timezone: Tue, 29 Sep 2020 15:00:00 GMT
Date after moment-timezone (without pre offset): Tue, 29 Sep 2020 15:00:00 GMT // Will be discussed later
```

So, what about a different timezone environment? For example, the CI environment.

Then here is another demo: https://repl.it/repls/LividBewitchedControlflowgraph#index.js

```
Date#getTimezoneOffset: 0 // See? It is not -480 anymore, it is now 0
Date parsed by js-yaml: Tue, 29 Sep 2020 23:00:00 GMT
Date after calculating the timezone offset: Tue, 29 Sep 2020 23:00:00 GMT
Date after moment-timezone: Tue, 29 Sep 2020 15:00:00 GMT
Date after moment-timezone (without pre offset): Tue, 29 Sep 2020 15:00:00 GMT // Will be discussed later
```

As you can see, `Date after calculating the timezone offset` is now wrong, while `Date after moment-timezone` is still correct. That's because `timezone()` takes `getTimezoneOffset` into consideration.

----

So, what about removing "pre-offset" operation from `hexo-front-matter`? So here goes `Date after moment-timezone (without pre offset)`. It is generated without using `Date#getTimezoneOffset`, and the result is still desired. It also means it is not affected by the environment (no `Date#getTimezoneOffset being used).

By eliminating the environment (only dependents on users' `config.timezone` configuration) the timezone issue should be solved.

stevenjoezhang:
## What does it do?

See https://github.com/hexojs/hexo-front-matter/pull/146

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/hexo/post.ts
+++ b/lib/hexo/post.ts
@@ -509,7 +509,9 @@ class Post {
       return readFile(src);
     }).then(content => {
       // Create post
-      Object.assign(data, yfmParse(content));
+      Object.assign(data, yfmParse(content, {
+        defaultTimeZone: config.timezone
+      }));
       data.content = data._content;
       data._content = undefined;
 
--- a/lib/hexo/validate_config.ts
+++ b/lib/hexo/validate_config.ts
@@ -27,8 +27,11 @@ export = (ctx: Hexo): void => {
   }
 
   if (!config.timezone) {
-    log.warn('No timezone setting detected! Using the default timezone UTC.');
-    config.timezone = 'UTC';
+    log.warn('No timezone setting detected! Using LocalTimeZone as the default timezone.');
+    log.warn('This behavior will be changed to UTC in the next major version. Please set timezone explicitly (e.g. LocalTimeZone or America/New_York) in _config.yml to avoid this warning.');
+    config.timezone = moment.tz.guess();
+  } else if (config.timezone.toLowerCase() === 'localtimezone') {
+    config.timezone = moment.tz.guess();
   } else {
     const configTimezone = moment.tz.zone(config.timezone);
     if (!configTimezone) {
--- a/lib/hexo/validate_config.ts
+++ b/lib/hexo/validate_config.ts
@@ -27,14 +27,15 @@ export = (ctx: Hexo): void => {
   }
 
   if (!config.timezone) {
-    log.warn('No timezone setting detected! Using LocalTimeZone as the default timezone.');
-    log.warn('This behavior will be changed to UTC in the next major version. Please set timezone explicitly (e.g. LocalTimeZone or America/New_York) in _config.yml to avoid this warning.');
+    log.warn('No timezone setting detected! Using the default timezone UTC.');
+    config.timezone = 'UTC';
   } else {
     const configTimezone = moment.tz.zone(config.timezone);
     if (!configTimezone) {
       log.warn(
-        `Invalid timezone setting detected! "${config.timezone}" is not a valid timezone.`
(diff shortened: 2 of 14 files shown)


yoshinorin (commented):
Since this PR is a potentially breaking change, I think it would be better to include it in the 9.x release rather than a minor version. What are your thoughts?

yoshinorin:
> I think merging it into 8.x or waiting until 9.0 would both be acceptable approaches. We should remind users to properly configure their default timezone, or, let Hexo write the timezone into the configuration file automatically during the first initialization.

Most users probably have the same timezone set for both their PC and `config.yml`, so this likely won't be an issue for many users. 

That said, I don't think typical users would expect this kind of change to be included in a minor version update. The possibility of URLs changing is particularly concerning (I apologize, I overlooked this aspect). 

Therefore, I personally think it would be preferable to release this in a major version.

tomap (approved):

yoshinorin (approved):
LGTM 👍

PR #5785, opened 2026-06-13
fix: Swig tag breaks Markdown processing on the same line

KentarouTakeda:
> **Note:** This PR recreates #5752. I accidentally closed that PR — a force-push during a rebase auto-closed it — and GitHub no longer allows it to be reopened (*"the repository may be missing relevant data"*), so I'm recreating it here with the same changes. Original review thread: #5752. cc @VibhorGautam — thanks again for the suggestion, which is now included as a fix below.

### Problem

When a Swig/Nunjucks tag appears at the beginning of a line, Markdown syntax on the rest of that line is not processed.

**Input:**
~~~markdown
{% post_link my-post 'My Post' %} and [link](https://example.com)
~~~

**Expected:** `link` becomes a clickable link.
**Actual:** `[link](https://example.com)` is rendered as plain text.

### Cause

Hexo replaces Swig tags with HTML comment placeholders (`<!--swig￼N-->`) before Markdown rendering. When this placeholder starts a line, it triggers an [HTML block (type 2)](https://spec.commonmark.org/0.31.2/#html-blocks) per the CommonMark spec — the entire line is treated as raw HTML and no Markdown processing occurs (see [Example 177](https://spec.commonmark.org/0.31.2/#example-177)).

### Fix

A single placeholder format cannot satisfy both requirements:

- **Block-level tags** (e.g. `{% blockquote %}`) need the HTML comment placeholder to trigger HTML block parsing, which prevents the Markdown renderer from wrapping the output in `<p>` tags.
- **Inline tags with trailing content** (e.g. `{% post_link %} [link](url)`) must *not* trigger HTML block parsing, or the rest of the line loses Markdown processing.

This PR adds an inline placeholder format (`￼swigN￼`) that uses U+FFFC (Object Replacement Character) as delimiters instead of HTML comments. U+FFFC is already used in the existing block placeholder format and is safe for this purpose — it has no special meaning in HTML or Markdown, does not appear in normal text, and survives Markdown rendering unchanged.

The state machine in `escapeAllSwigTags` looks ahead after each closing `%}` / `}}`: if non-whitespace content follows on the same line, the inline format is used; otherwise the existing block format (`<!--swig￼N-->`) is used.

### Additional fix: trailing Markdown after a block tag's *closing*

Following @VibhorGautam's review suggestion on #5752, I added a regression test for a full block tag whose **closing** is followed by Markdown on the same line:

~~~markdown
{% blockquote %}quote{% endblockquote %} [link](https://example.com)
~~~

The test surfaced a real bug: the same-line Markdown was **not** processed, because the full-block-tag branch in `escapeAllSwigTags` was the only escape path that didn't pass the inline-placeholder flag — the same root cause as above, on the closing side. Fixed by applying the same `hasTrailingContentOnLine` look-ahead there. All `post.ts` tests pass (121 passing / 1 pending).

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/hexo/post.ts
+++ b/lib/hexo/post.ts
@@ -16,6 +16,7 @@ const rHexoPostRenderEscape = /<hexoPostRenderCodeBlock>([\s\S]+?)<\/hexoPostRen
 const rSwigTag = /(\{\{.+?\}\})|(\{#.+?#\})|(\{%.+?%\})/s;
 
 const rSwigPlaceHolder = /(?:<|&lt;)!--swig\uFFFC(\d+)--(?:>|&gt;)/g;
+const rInlineSwigPlaceHolder = /(?:\uFFFC|&#(?:xFFFC|xfffc|65532);)swig(\d+)(?:\uFFFC|&#(?:xFFFC|xfffc|65532);)/g;
 const rCodeBlockPlaceHolder = /(?:<|&lt;)!--code\uFFFC(\d+)--(?:>|&gt;)/g;
 const rCommentHolder = /(?:<|&lt;)!--comment\uFFFC(\d+)--(?:>|&gt;)/g;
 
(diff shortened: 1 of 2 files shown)


stevenjoezhang:
Thank you for the contribution and for the thorough regression tests.

We are currently refactoring Hexo’s post-render parsing to use a shared, context-aware lexer (#5800). This work overlaps substantially with the code changed by this PR, so merging it now would introduce conflicts and duplicate logic.

Once the refactor is complete, we plan to rebase and adapt this PR onto the new implementation, then merge the behavior and tests. No action is needed from you for now. Thank you for your patience!

PR #5633, opened 2025-02-13
refactor(model): add Code model for include_code tag

D-Sketon:
## What does it do?

try to fix #5479 and #5486

ref https://github.com/hexojs/hexo/issues/5479#issuecomment-2080453292

Since the current model doesn't satisfy `include_code`, I experimentally added the Code Model

## Pull request tasks

- [x] Add test cases for the changes.
- [x] Passed the CI test.

Issue #5486, opened 2024-05-03
`include_code` tag is broken in Hexo 7.2.0


njzjz:
### Check List

- [X] I have already read [Docs page](https://hexo.io/docs/) & [Troubleshooting page](https://hexo.io/docs/troubleshooting).
- [X] I have already searched existing issues and they are not help to me.
- [X] I examined error or warning messages and it's difficult to solve.
- [X] I am using the [latest](https://github.com/hexojs/hexo/releases) version of Hexo. (run `hexo version` to check)
- [X] My Node.js is matched [the required version](https://hexo.io/docs/#Required-Node-js-version).

### Expected behavior

`include_code` tag can be correctly rendered in Hexo 7.1.1. See https://0bf5f2be.hexo-tag-publications.pages.dev.

![image](https://github.com/hexojs/hexo/assets/9496702/9cf27c12-cba0-43c7-b2c3-5cb186457bfb)

### Actual behavior

After bumping Hexo to v7.2.0 (https://github.com/njzjz/hexo-tag-publications/pull/329), nothing is rendered: https://3e49fa9c.hexo-tag-publications.pages.dev/

![image](https://github.com/hexojs/hexo/assets/9496702/39fde997-f91d-4497-bf45-d9449795f066)

### How to reproduce?

```
{% include_code pub.bib lang:bib pub.bib %}
```

### Is the problem still there under `Safe mode`?

No

### Your Node.js & npm version

```text
node v18.20.2
yarn v1.22.4
```

### Your Hexo and Plugin version

```text
7.2.0
```

### Your `package.json`

```json
{
  "name": "hexo-tag-publications-site",
  "version": "0.0.0",
  "private": true,
  "scripts": {
    "build": "hexo generate",
    "clean": "hexo clean",
    "deploy": "hexo deploy",
    "server": "hexo server"
  },
  "hexo": {
    "version": "6.2.0"
  },
  "dependencies": {
    "@njzjz/hexo-sitemap": "0.0.3",
    "hexo": "7.2.0",
    "hexo-generator-archive": "2.0.0",
    "hexo-generator-category": "2.0.0",
    "hexo-generator-index": "3.0.0",
    "hexo-generator-tag": "2.0.0",
    "hexo-renderer-ejs": "2.0.0",
    "hexo-renderer-marked": "6.3.0",
    "hexo-renderer-stylus": "3.0.1",
    "hexo-server": "3.0.0",
    "hexo-tag-publications": "file:./../",
[...]

PR #5807, opened 2026-08-08
refactor(model): add Code model for include_code tag

### LINKED ISSUE #5479: .j2 file in code_dir causing rendering error
### Check List

- [X] I have already read [Docs page](https://hexo.io/docs/) & [Troubleshooting page](https://hexo.io/docs/troubleshooting).
- [X] I have already searched existing issues and they are not help to me.
- [X] I examined error or warning messages and it's difficult to solve.
- [X] I am using the [latest](https://github.com/hexojs/hexo/releases) version of Hexo. (run `hexo version` to check)
- [X] My Node.js is matched [the required version](https://hexo.io/docs/#Required-Node-js-version).

### Expected behavior

`.j2` files in `code_dir` folder should not breaks `hexo generate`

### Actual behavior

`hexo generate` will fail. (see the reproducing steps)

### How to reproduce?

```
hexo init hexo-test
cd hexo-test

mkdir -p source/downloads/code
touch source/downloads/code/j2-test.j2

echo '{% include_code j2-test.j2 j2-test.j2 %}' >> source/_posts/hello-world.md

hexo clean && hexo generate --debug
```

Gives the following error:

```
06:30:54.928 FATAL Something's wrong. Maybe you can find the solution here: https://hexo.io/docs/troubleshooting.html
Template render error: (unknown path)
  TypeError: Cannot read properties of undefined (reading 'split')
    at Object._prettifyError (/Users/jinzhouz/tmp/hexo-test/node_modules/nunjucks/src/lib.js:32:11)
    at /Users/jinzhouz/tmp/hexo-test/node_modules/nunjucks/src/environment.js:464:19
    at eval (eval at _compile (/Users/jinzhouz/tmp/hexo-test/node_modules/nunjucks/src/environment.js:527:18), <anonymous>:11:11)
    at tryCatcher (/Users/jinzhouz/tmp/hexo-test/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/Users/jinzhouz/tmp/hexo-test/node_modules/bluebird/js/release/promise.js:547:31)
    at Promise._settlePromise (/Users/jinzhouz/tmp/hexo-test/node_modules/bluebird/js/release/promise.js:604:18)
    at Promise._settlePromiseCtx (/Users/jinzhouz/tmp/hexo-test/node_modules/bluebird/js/release/promise.js:641:10)
    at _drainQueueStep (/Users/jinzhouz/tmp/hexo-test/node_modules/bluebird/js/release/async.js:97:12)
    at _drainQueue (/Users/jinzhouz/tmp/hexo-test/node_modules/bluebird/js/release/async.js:86:9)
    at Async._drainQueues (/Users/jinzhouz/tmp/hexo-test/node_modules/bluebird/js/release/async.js:102:5)
    at Async.drainQueues (/Users/jinzhouz/tmp/hexo-test/node_modules/bluebird/js/release/async.js:15:14)
    at process.processImmediate (node:internal/timers:478:21)
```

no

```text
v21.4.0
10.2.4
```

```text
hexo-site@0.0.0 /Users/jinzhouz/tmp/hexo-test
├── hexo-generator-archive@2.0.0
├── hexo-generator-category@2.0.0
├── hexo-generator-index@3.0.0
├── hexo-generator-tag@2.0.0
├── hexo-renderer-ejs@2.0.0
├── hexo-renderer-marked@6.3.0
├── hexo-renderer-stylus@3.0.1
├── hexo-server@3.0.0
├── hexo-theme-landscape@1.0.0
└── hexo@7.2.0
```

### Your `package.json`

```json
{
  "name": "hexo-site",
  "version": "0.0.0",
  "private": true,
  "scripts": {
    "build": "hexo generate",
    "clean": "hexo clean",
    "deploy": "hexo deploy",
    "server": "hexo server"
  },
  "hexo": {
    "version": "7.2.0"
  },
  "dependencies": {
    "hexo": "^7.0.0",
    "hexo-generator-archive": "^2.0.0",
    "hexo-generator-category": "^2.0.0",
    "hexo-generator-index": "^3.0.0",
    "hexo-generator-tag": "^2.0.0",
    "hexo-renderer-ejs": "^2.0.0",
    "hexo-renderer-marked": "^6.0.0",
    "hexo-renderer-stylus": "^3.0.0",
    "hexo-server": "^3.0.0",
    "hexo-theme-landscape": "^1.0.0"
  }
}
```

_No response_

### Others

_No response_

D-Sketon:
see https://github.com/hexojs/hexo/pull/5633

try to fix #5479 and #5486

ref https://github.com/hexojs/hexo/issues/5479#issuecomment-2080453292

Since the current model doesn't satisfy `include_code`, I experimentally added the Code Model


lib/plugins/processor/code.ts
```diff
@@ -0,0 +1,45 @@
+import { Pattern } from 'hexo-util';
+import { relative } from 'path';
+import { isExcludedFile } from './common';
+import type Hexo from '../../hexo';
+import type { _File } from '../../box';
+
+export = (ctx: Hexo) => {
+  let codeDir = ctx.config.code_dir;
```
stevenjoezhang: Could we resolve `code_dir` inside the pattern callback, or otherwise defer it until after the configuration has been loaded?

Internal processors are registered before `_config.yml` is loaded, so `ctx.config.code_dir` still has its default value when this closure is created. For example, with `code_dir: snippets`, files under `source/snippets/` are not added to the `Code` model, while `include_code` looks them up there and consequently produces no output.

The asset processor captures `code_dir` at the same stage, so it should be updated consistently. It would also be helpful to add an integration test that configures a custom `code_dir` and then calls `hexo.init()` and `hexo.load()`.

lib/plugins/processor/code.ts
```diff
@@ -0,0 +1,45 @@
...
+
+      if (file.type === 'delete') {
+        if (doc) {
+          return doc.remove();
+        }
+
+        return;
+      }
+
+      if (file.type === 'skip' && doc) {
```
stevenjoezhang, line 30: Could we mark the existing document as unmodified here instead of returning immediately?

After a file is initially created, its `modified` field is set to `true`. On subsequent unchanged builds, this branch leaves it as `true`, so the code generator treats the route as modified and rewrites it on every build. The asset processor avoids this by persisting `modified: file.type !== 'skip'`.

Please update and save the document with `modified: false` in this branch, and extend the existing `type: skip` test to assert the value.

PR #5809, opened 2026-08-08
fix: migrate YAML parsing from js-yaml to yaml

### LINKED ISSUE #5801: `source/_data` YAML can define executable JS via default `!!js/function` (`js-yaml-js-types`.all)
### Check List

- [x] I have already read [Docs page](https://hexo.io/docs/) & [Troubleshooting page](https://hexo.io/docs/troubleshooting).
- [x] I have already searched existing issues and they are not help to me.
- [x] I examined error or warning messages and it's difficult to solve.
- [x] I am using the [latest](https://github.com/hexojs/hexo/releases) version of Hexo. (run `hexo version` to check)
- [x] My Node.js is matched [the required version](https://hexo.io/docs/#Required-Node-js-version).

### Expected behavior

YAML files under `source/_data/` should be parsed as plain data. Default parsing should not construct JavaScript functions from YAML tags.

If `!!js/regexp` / `!!js/undefined` are still needed, only those should be enabled — not `!!js/function` by default.

### Actual behavior

Hexo’s YAML renderer enables `js-yaml-js-types`.all by default:

https://github.com/hexojs/hexo/blob/master/lib/plugins/renderer/yaml.ts

```js
schema = yaml.DEFAULT_SCHEMA.extend(require('js-yaml-js-types').all);
```

That includes `!!js/function`. A `source/_data/*.yml` file can define a real Function (for example on `toString`). When a theme renders `{{ site.data.menu }}`, Nunjucks stringifies the object, calls that function, and arbitrary Node.js code runs in the Hexo process.

`js-yaml` v4 made these tags opt-in; re-enabling `.all` by default is unexpected for site data files.

### How to reproduce?

1. Create a minimal Hexo site with `hexo@8.1.2` and a theme layout containing:

```njk
{{ site.data.menu }}
```

2. Create `source/_data/menu.yml`:

```yaml
toString: !!js/function 'function (){ process.getBuiltinModule("fs").writeFileSync("/tmp/hexo_js_function","1"); return ""; }'
Home: /
```

3. Load the site and render that template (same path themes use when printing `site.data`):

```js
const Hexo = require('hexo');
const nunjucks = require(require.resolve('nunjucks', { paths: [require.resolve('hexo')] }));

(async () => {
  const hexo = new Hexo(process.cwd(), { silent: true });
  await hexo.init();
  await hexo.load();
  const menu = hexo.locals.get('data').menu;
  console.log(typeof menu.toString); // function
  nunjucks.renderString('{{ site.data.menu }}', { site: { data: { menu } } });
})();
```

4. Observe that `/tmp/hexo_js_function` is created with contents `1`.

Yes. This comes from Hexo’s built-in YAML renderer (`lib/plugins/renderer/yaml.ts`), not from a third-party plugin/script. Disabling plugins does not remove this schema extension.

```text
node v22.22.0
npm 9.2.0
```

```text
hexo@8.1.2
```

### Your `package.json`

```json
{
  "name": "hexo-yaml-jsfunction-repro",
  "private": true,
  "hexo": {},
  "dependencies": {
    "hexo": "8.1.2"
  }
}
```

```yaml

```

### Others

**Suggested fix**

- Do not use `js-yaml-js-types`.all by default.
- Keep the default safe schema, or extend only `regexp` / `undefined` if still required.
- If `!!js/function` must remain for compatibility, make it an explicit opt-in (default off).

stevenjoezhang:
## What does it do?

Fixes #5801.

This migrates Hexo's direct YAML parsing and stringification from `js-yaml` to `yaml`, and removes the direct dependencies on `js-yaml`, `js-yaml-js-types`, and `@types/js-yaml`.

`js-yaml-js-types` extends the schema with `!!js/function`, allowing YAML input to construct executable functions. The new shared parser rejects unresolved tags such as `!!js/function`, while preserving the existing behavior for:

- YAML timestamps parsed as `Date`
- YAML merge keys (`<<`)
- `!!js/regexp`
- Duplicate keys in JSON multi-config files

YAML parsing and stringification in the renderer, multi-config handling, config command, and scaffold front matter have been migrated. Related tests now use `yaml` as well.

`js-yaml` remains a transitive dependency of `hexo-front-matter@4.2.1`. Upgrading to `hexo-front-matter@5` is kept separate because it is tied to the timezone migration. The vulnerable `js-yaml-js-types` dependency is removed entirely.

Local validation:

- `npm run build`
- `npm run eslint`
- YAML renderer tests: 6 passing
- Full test suite: 1281 passing, 5 pending; the 7 failures under Node.js 26 are also present on the base branch and are unrelated to this change

## Screenshots

N/A — internal parser and dependency changes only.

## Pull request tasks

- [ ] Passed the CI test.

### DIFF
--- a/lib/hexo/multi_config_path.ts
+++ b/lib/hexo/multi_config_path.ts
@@ -1,8 +1,9 @@
 import { isAbsolute, resolve, join, extname } from 'path';
 import { existsSync, readFileSync, writeFileSync } from 'hexo-fs';
-import yml from 'js-yaml';
+import { stringify } from 'yaml';
 import { deepMerge } from 'hexo-util';
 import type Hexo from './index';
+import parseYaml from './yaml';
 
 export = (ctx: Hexo) => function multiConfigPath(base: string, configPaths?: string, outputDir?: string): string {
   const { log } = ctx;
@@ -47,10 +48,10 @@ export = (ctx: Hexo) => function multiConfigPath(base: string, configPaths?: str
     const ext = extname(paths[i]).toLowerCase();
 
     if (ext === '.yml') {
-      combinedConfig = deepMerge(combinedConfig, yml.load(file));
+      combinedConfig = deepMerge(combinedConfig, parseYaml(file));
       count++;
     } else if (ext === '.json') {
-      combinedConfig = deepMerge(combinedConfig, yml.load(file, {json: true}));
+      combinedConfig = deepMerge(combinedConfig, parseYaml(file, { uniqueKeys: false }));
       count++;
     } else {
       log.w(`Config file ${paths[i]} not supported type.`);
@@ -69,7 +70,7 @@ export = (ctx: Hexo) => function multiConfigPath(base: string, configPaths?: str
 
   log.d(`Writing _multiconfig.yml to ${outputPath}`);
 
-  writeFileSync(outputPath, yml.dump(combinedConfig));
+  writeFileSync(outputPath, stringify(combinedConfig));
 
   // write file and return path
   return outputPath;
--- a/lib/hexo/post.ts
+++ b/lib/hexo/post.ts
@@ -3,12 +3,12 @@ import moment from 'moment';
 import Promise from 'bluebird';
 import { join, extname, basename } from 'path';
 import { magenta } from 'picocolors';
-import { load } from 'js-yaml';
 import { slugize, escapeRegExp, deepMerge} from 'hexo-util';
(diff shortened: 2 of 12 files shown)


yoshinorin (commented):
@stevenjoezhang 
Question:

I haven't actually tried this myself, but it looks like the issue raised in the original report could be fixed while staying on `js-yaml`, simply by dropping `.all`. Is the motivation for switching the library to `yaml` that `hexo-front-matter` moves to yaml in v5?

PR #5823, opened 2026-08-27
feat(theme): validate config with JSON Schema

### LINKED ISSUE #4580: Verify the configuration file
## Check List

## Feature Request

> ## Expected behavior
> Verify the configuration file of the theme, when our configuration changes, or the user fills in the wrong content, wecan remind the user to modify.
> 
> I try to use ajv check
> 
> * https://github.com/ajv-validator/ajv
> 
> ```shell
> yarn add hexo-theme-cake
> # modify `theme` to `cake`
> # exec check. see impl: https://github.com/jiangtj/hexo-theme-cake/blob/master/lib/console/check.js
> hexo check
> ```
> 
> But the problem is that we need to maintain a huge json-schema. Even the compressed next theme (cake), config file line about 140. ~The schema has 276 lines. [schema.json](https://github.com/jiangtj/hexo-theme-cake/blob/master/lib/console/schema.json).~ The schema has 160 lines. [schema.yml](https://github.com/jiangtj/hexo-theme-cake/blob/master/lib/console/schema.yml).
> 
> Do you think? @stevenjoezhang
> 
> ## Example
> When I fill in a wrong value
> 
> image

_Originally posted by @jiangtj  in https://github.com/next-theme/hexo-theme-next/issues/20_

stevenjoezhang:
## What does it do?

Closes #4580.

This adds opt-in theme configuration validation to Hexo. A theme can provide a conventional `config.schema.json` file in its root; Hexo discovers it automatically without requiring a `package.json` declaration.

Before generators run, Hexo validates the fully merged theme configuration with JSON Schema 2020-12 and standard format support. Themes without the schema file keep the current behavior.

Validation is deliberately non-mutating: it does not coerce types, inject defaults, or remove unknown properties. It reports all detected paths in one error while avoiding configuration values in the message. Compiled validators are cached by schema path and content, so schema edits are picked up during watch mode.

Example failure:

```text
Theme config validation failed using "themes/next/config.schema.json":
  - /darkmode must be boolean
```

### Verification

- Added unit coverage for opt-in behavior, valid and invalid configurations, merged theme overrides, malformed and asynchronous schemas, and schema reloads.
- `npm run eslint`
- `npm run build`
- 49 focused filter and Hexo lifecycle tests
- End-to-end generation against NexT 8.29.0: 696 routes generated with the real configuration; an invalid `darkmode` override was rejected before public output was written.

## Screenshots

N/A — CLI-only behavior; representative output is included above.

## Pull request tasks

- [x] Passed the CI test.

### DIFF
--- a/lib/plugins/filter/before_generate/index.ts
+++ b/lib/plugins/filter/before_generate/index.ts
@@ -3,5 +3,6 @@ import type Hexo from '../../../hexo';
 export = (ctx: Hexo) => {
   const { filter } = ctx.extend;
 
+  filter.register('before_generate', require('./validate_theme_config'), -100);
   filter.register('before_generate', require('./render_post'));
 };
--- a/lib/plugins/filter/before_generate/validate_theme_config.ts
+++ b/lib/plugins/filter/before_generate/validate_theme_config.ts
@@ -1,0 +1,100 @@
+import { join } from 'path';
+import Ajv2020, { type ErrorObject, type ValidateFunction } from 'ajv/dist/2020';
+import addFormats from 'ajv-formats';
+import { exists, readFile } from 'hexo-fs';
+import type Hexo from '../../../hexo';
+
+interface CachedValidator {
+  source: string;
+  validate: ValidateFunction;
+}
+
+const validatorCache = new Map<string, CachedValidator>();
+
+function errorMessage(error: unknown): string {
+  return error instanceof Error ? error.message : String(error);
+}
+
+async function loadValidator(schemaPath: string): Promise<ValidateFunction> {
+  let source: string,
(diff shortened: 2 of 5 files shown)


SukkaW (changes requested):
IMHO I am not a fan of JSON schema, very poor DX. I'd rather choose zod or valibot.

So here is the deal: instead of forcing the theme authors to use JSON schema, we should provide a new theme config validation callback filter API (`hexo.extend.filter.register('theme_config_validation', (parsedThemeConfig) => {})`, or we should call it `before_theme_config`?) that theme authors can register via the `scripts`, and themes can choose what validation they want to use.

PR #5827, opened 2026-09-09
fix(renderer): disable executable !!js/function by default in YAML renderer

### LINKED ISSUE #5801: `source/_data` YAML can define executable JS via default `!!js/function` (`js-yaml-js-types`.all)

jabrailkhalil:
## Summary

The YAML renderer previously extended `yaml.DEFAULT_SCHEMA` with `js-yaml-js-types.all`, which enables the executable `!!js/function` tag. Any `source/_data/*.yml` file (or config/theme data) could then define a real JavaScript function that gets invoked when the data is rendered, allowing arbitrary Node.js code to run in the Hexo process.

With this change only the safe JS types are enabled: `!!js/regexp` and `!!js/undefined`. The executable `!!js/function` tag is no longer accepted and now throws a YAML parsing error instead of constructing a function. The `#4917` workaround (catch around schema creation) is preserved.

## Tests

Added three tests to `test/scripts/renderers/yaml.ts`:

- `!!js/regexp` still parses into a `RegExp` with the right source/flags.
- `!!js/undefined` still parses into `undefined`.
- `!!js/function` is rejected (throws).

## Verification

- `npm test -- --grep yaml` - 7/7 passing (including the 3 new tests)
- `npm run eslint` - clean
- `npm run build` (tsc -b) - clean
- Full `npm test`: 1294 passing, 5 pending, 1 failing

The single failing test is pre-existing and unrelated to this change: `test/scripts/console/generate.ts` -> "generate - future posts -> reprocesses future posts after publish date passes" fails on this machine (Windows 10, Node v22.21.1) with "expected [ _Document{...(21)} ] to have a length of +0 but got 1". It also fails on a pristine checkout of `master` without this change, including when run in isolation via `--grep`. The sinon fake-timer clock (fake `Date` around year 2098) does not appear to take effect in this environment.

Fixes #5801


jabrailkhalil:
The CI workflows for the current head `bf09f2df77` are waiting for contributor workflow approval (`action_required`): [Benchmark](https://github.com/hexojs/hexo/actions/runs/34404758015), [Linter](https://github.com/hexojs/hexo/actions/runs/34404758119), [Tester](https://github.com/hexojs/hexo/actions/runs/34404758101). Could a maintainer approve the pending runs when convenient? These runs have not executed their checks yet; I will address any failures once they run.

